PR #11, opened 2018-07-26
Re-modelling SRIOV device plugin

zshi-redhat:
Remove cnishim as CNI doesn't talk with Device Plugin directly.
  * api proto between cnishim and device plugin
  * cnishim cmd code

Remove checkpoint file restore as Device Plugin doesn't track
consumed Devices
  * checkpoint folder
  * restorePodMapping function in sriov device plugin cmd

Remove allocatedDevices and managedDevices in sriovManager

Remove patches for kubernetes DP API and sriov-cni

Remove cnishim build in build.sh

Modify cnishim-crd.yaml to sriov-crd.yaml and update pod-tc[x].yaml


cmd/sriovdp/sriov-device-plugin.go
```diff
@@ -80,37 +75,11 @@ func newSriovManager() *sriovManager {
 		glog.Errorf("Error. Could not create K8s Client using supplied config. %v", err)
 		return nil
 	}
-	sm := &sriovManager{
+	return &sriovManager{
 		k8ClientSet:      clientset,
```
ahalim-intel: K8sClientSet no longer needed if there's no communication to the API server. You can remove this as well

PR #17, opened 2018-09-13
Gracefully handle non-dirs under /sys/class/net/

booxter:
Some inodes there, e.g. bonding_masters, may not be directories at all but
regular files. Without the fix, the plugin just panics with segfault when such
scenario occurs.


zshi-redhat:
Thanks Ihar for the PR!
Tested on fedora 28 with bond interfaces created, no panic.
And it seems os.IsNotExist[https://github.com/intel/sriov-network-device-plugin/blob/master/cmd/sriovdp/sriov-device-plugin.go#L85](url) doesn't capture the error when sriovDirPath is a file.

LGTM.

PR #23, opened 2018-09-19
Switch to kubelet plugin watcher mode

zshi-redhat:
KubeletPluginWatcher is enabled by default in kube-1.12.0;
This commit switches registration of Device Plugin to watcher
mode instead of explicitly calling Register in Device Plugin.
This also means Device Plugin will only work with
KubeletPluginWatcher featureGate enabled.


cmd/sriovdp/sriov-device-plugin.go
```diff
@@ -373,11 +388,17 @@ func main() {
...
+	// Catch termination signals and kubelet restart
 	select {
 	case sig := <-sigCh:
 		glog.Infof("Received signal \"%v\", shutting down.", sig)
 		sm.Stop()
 		return
+	case event := <-watcher.Events:
+		if event.Op&fsnotify.Create == fsnotify.Create {
+			glog.Infof("Received kubelet restart notification \"%v\". shutting down.", event)
+			sm.Stop()
```
ahalim-intel: Instead of stopping the server it needs to re-register itself with Kubelet Device manager again.
However, the Kubelet Device manager recently added feature where individual plugin no longer needs to watch for Kubelet restart. Instead Kubelet re-initiate plugin registration on restart.  But older approach is still supported.

Please see here for more info: https://docs.google.com/document/d/1dtHpGY-gPe9sY7zzMGnm8Ywo09zJfNH-E1KEALFV39s 

And the PR(Merged) : https://github.com/kubernetes/kubernetes/pull/58755

cmd/sriovdp/sriov-device-plugin.go
```diff
@@ -373,11 +388,17 @@ func main() {
...
 
-	// Catch termination signals
+	// Catch termination signals and kubelet restart
 	select {
 	case sig := <-sigCh:
 		glog.Infof("Received signal \"%v\", shutting down.", sig)
 		sm.Stop()
 		return
+	case event := <-watcher.Events:
+		if event.Op&fsnotify.Create == fsnotify.Create {
```
ahalim-intel: Should it be we watching for 'Create' OR 'Remove' event?

cmd/sriovdp/sriov-device-plugin.go
zshi-redhat: Thanks Abdul for the review and info!

My original thinking is that: when kubelet gets restarted, the pod running sriov device plugin will be restarted if sriov device plugin explicitly exit;  by restarting the pod, sriov device plugin binary will be executed again which automatically does the re-registration to kubelet. In this case, what sriov device plugin should do is to capture this event and exit gracefully,  pod restart will trigger the re-run of sriov device plugin binary.

cmd/sriovdp/sriov-device-plugin.go
zshi-redhat: Yes, I think basically we need to capture all the 'negative' events and exit.

cmd/sriovdp/sriov-device-plugin.go
```diff
@@ -29,6 +29,7 @@ import (
 	"time"
 
 	"github.com/golang/glog"
+	"github.com/fsnotify/fsnotify"
```
s1061123: I'm wondering that why glide.yaml is not changed although you add fsnotify in vendor directory...

Did you check glide.yaml in your workspace?

zshi-redhat: Yep, I updated the dependency package manually so glide.yaml is not changed, in my env, glide.yaml doesn't work as expected and it might be a problem we shall fix, but need more testing.

cmd/sriovdp/sriov-device-plugin.go
ahalim-intel: > Thanks Abdul for the review and info!
> 
> My original thinking is that: when kubelet gets restarted, the pod running sriov device plugin will be restarted if sriov device plugin explicitly exit; by restarting the pod, sriov device plugin binary will be executed again which automatically does the re-registration to kubelet. In this case, what sriov device plugin should do is to capture this event and exit gracefully, pod restart will trigger the re-run of sriov device plugin binary.

Kubelet restart does not affect existing Pod running. Normally Pod wont' restart with Kubelet, they are expected to be functioning even without Kubelet's presence. What we want to do here is in such events, the device plugin should detect Kubelet restarts then re-register with it when its back on running. Device plugin can be still running in its Pod, just there is not request coming in from Kubelet. So, for kubelet to be aware a device plugin presence it needs to re-register. In probe-based mechanism this won't be needed but device plugin needs to implement additional service that Kubelet will communicate and invoke registration process when it finds a valid socket file.

zshi-redhat:
Change the commit title to "Introduce kubelet plugin watcher model " and updated another version;
tested with kube-1.12.0; device plugin can be re-registered when kubelet gets restarted.

zshi-redhat:
Deprecating this PR ( it is outdated and was created against master branch ) and creating a new [one](https://github.com/intel/sriov-network-device-plugin/pull/99) to add plugin watch support in release-v1 branch.

PR #30, opened 2018-10-26
allowing proxy-settings to be passed in for Docker

ahalim-intel:
This changes will allow proxy settings to be passed in for docker
as well as Env variable in Dockerfile.

simply invoke 'make image HTTP_PROXY=http://whateverhost:port'.

Change-Id: I4264b2a96d86a6feedee1b21a9f1bac8e74c8fc5

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,8 @@ FROM centos:centos7
 
 ADD . /usr/src/sriov-network-device-plugin
 
+ENV HTTP_PROXY $http_proxy
+ENV HTTPS_PROXY $https_proxy
 ENV INSTALL_PKGS "git golang make"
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -2,6 +2,8 @@ FROM centos:centos7
 
 ADD . /usr/src/sriov-network-device-plugin
 
+ENV HTTP_PROXY $http_proxy
```
zshi-redhat: These ENV variables will persist in the build/final images, I'm not sure if it'll cause any problem when the image is running with HTTP(S)_PROXY set as None, any thought on that?

zshi-redhat: Sounds good, Thanks

zshi-redhat:
LGTM

PR #32, opened 2018-11-08
add link status detection

zshi-redhat:
Detects Link status for Linux network devices and
updates associated VFs health state accordingly every
5 seconds.

Add GetVFList util and simplify discoverNetworks.

Add rootDevices in sriovManager to track discovered
and SR-IOV enabled PF devices.

### DIFF
--- a/cmd/sriovdp/sriov-device-plugin.go
+++ b/cmd/sriovdp/sriov-device-plugin.go
@@ -190,6 +190,7 @@ func (sm *sriovManager) discoverNetworks() error {
 	return nil
 }
 
+// IsNetlinkStatusUp returns 'false' if 'operstate' is not "up" for a Linux netowrk device
 func IsNetlinkStatusUp(dev string) bool {
 	opsFile := filepath.Join(netDirectory, dev, "operstate")
 	bytes, err := ioutil.ReadFile(opsFile)
@@ -199,6 +200,7 @@ func IsNetlinkStatusUp(dev string) bool {
 	return true
 }
 
+// Probe returns 'true' if device health changes 'false' otherwise
 func (sm *sriovManager) Probe() bool {
 	// Network device should check link status for each physical port and update health status for
 	// all associated VFs if there is any
@@ -313,9 +315,8 @@ func Register(kubeletEndpoint, pluginEndpoint, resourceName string) error {
 
 // Implements DevicePlugin service functions
 func (sm *sriovManager) ListAndWatch(empty *pluginapi.Empty, stream pluginapi.DevicePlugin_ListAndWatchServer) error {
-	changed := true
 	for {
-		if sm.Probe() || changed {
+		if sm.Probe() {
 			resp := new(pluginapi.ListAndWatchResponse)
 			for _, dev := range sm.devices {
 				resp.Devices = append(resp.Devices, &pluginapi.Device{ID: dev.ID, Health: dev.Health})
@@ -327,7 +328,6 @@ func (sm *sriovManager) ListAndWatch(empty *pluginapi.Empty, stream pluginapi.De
 				return err
 			}
 		}
-		changed = false
 		time.Sleep(5 * time.Second)
 	}
 	return nil
--- a/cmd/sriovdp/sriov-device-plugin.go
+++ b/cmd/sriovdp/sriov-device-plugin.go
@@ -25,6 +25,7 @@ import (
 	"path"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


s1061123:
How about to use [netlink's LinkSubscribe](https://godoc.org/github.com/vishvananda/netlink#LinkSubscribe) for monitoring link status?
With this API, we don't have to polling interface and get events from kernel through netlink.

I understand that could be trade-off and there is no absolute 'best' solution, so it is just an idea/comment.

cmd/sriovdp/sriov-device-plugin.go
```diff
@@ -149,41 +181,49 @@ func (sm *sriovManager) discoverNetworks() error {
...
-				if err != nil {
-					glog.Errorf("Error. Cannot read symbolic link between virtual function and PCI - Device: %s, VF: %v. Err: %v", dev, vf, err)
-					return err
-				}
-
-				pciAddr := pciInfo[len("../"):]
-				glog.Infof("PCI Address for device %s, VF %v is %s", dev, vf, pciAddr)
-
-				devName := pciAddr
-				sm.devices[devName] = pluginapi.Device{ID: devName, Health: pluginapi.Healthy}
```
ahalim-intel: I see you have refactored this section and extracted the GetVFList() method. But I don't see the initial device list is being populated anywhere. We only get a list of device only after a Probe() is invoked. The ListAndWatch() should send the the initial device list(which is now empty) first and will send updated list only when any device's health is changed.

cmd/sriovdp/sriov-device-plugin.go
```diff
@@ -149,41 +181,49 @@ func (sm *sriovManager) discoverNetworks() error {
...
 		}
 	}
+	glog.Infof("Discovered SR-IOV PF devices: %v", sm.rootDevices)
 	return nil
 }
 
-func (sm *sriovManager) GetDeviceState(DeviceName string) string {
-	// TODO: Discover device health
-	return pluginapi.Healthy
+func IsNetlinkStatusUp(dev string) bool {
```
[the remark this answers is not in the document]
zshi-redhat: +1, will add comment.

cmd/sriovdp/sriov-device-plugin.go
```diff
@@ -149,41 +181,49 @@ func (sm *sriovManager) discoverNetworks() error {
...
+	opsFile := filepath.Join(netDirectory, dev, "operstate")
+	bytes, err := ioutil.ReadFile(opsFile)
+	if err != nil || strings.TrimSpace(string(bytes)) != "up" {
+		return false
+	}
+	return true
+}
+
+func (sm *sriovManager) Probe() bool {
```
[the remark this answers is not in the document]
zshi-redhat: +1, will add comment.

cmd/sriovdp/sriov-device-plugin.go
zshi-redhat: The ListAndWatch() updates device information by calling Probe() first; when sm.devices[] list is empty, Probe() assigns default value to it and set the 'changed' flag to true which triggers the ListAndWatch to send ListAndWatchResponse back. The initial assignment of sm.devices[] is at https://github.com/intel/sriov-network-device-plugin/pull/32/files#diff-8dbc0a94295122ac9cb2460632637cbdR220

ahalim-intel: Ok, thanks for explanation.

ahalim-intel:
> How about to use [netlink's LinkSubscribe](https://godoc.org/github.com/vishvananda/netlink#LinkSubscribe) for monitoring link status?
> With this API, we don't have to polling interface and get events from kernel through netlink.
> 
> I understand that could be trade-off and there is no absolute 'best' solution, so it is just an idea/comment.

This might be a good solution going forward. But it will need a good bit of efforts to implementation. Maybe something we can look into in future as an improvement.

zshi-redhat:
> > How about to use [netlink's LinkSubscribe](https://godoc.org/github.com/vishvananda/netlink#LinkSubscribe) for monitoring link status?
> > With this API, we don't have to polling interface and get events from kernel through netlink.
> > I understand that could be trade-off and there is no absolute 'best' solution, so it is just an idea/comment.
> 
> This might be a good solution going forward. But it will need a good bit of efforts to implementation. Maybe something we can look into in future as an improvement.

+1 on the notification feature for link state change and agree with Abdul that we could look into it in future, Implement it on master branch first, then backport to release.

s1061123:
> > > How about to use [netlink's LinkSubscribe](https://godoc.org/github.com/vishvananda/netlink#LinkSubscribe) for monitoring link status?
> > > With this API, we don't have to polling interface and get events from kernel through netlink.
> > > I understand that could be trade-off and there is no absolute 'best' solution, so it is just an idea/comment.
> > 
> > 
> > This might be a good solution going forward. But it will need a good bit of efforts to implementation. Maybe something we can look into in future as an improvement.
> 
> +1 on the notification feature for link state change and agree with Abdul that we could look into it in future, Implement it on master branch first, then backport to release.

I've implement prototype with netlink LinkSubscribe(). It is not tested because I don't have any SR-IOV device ready testbed. Just FYI.
https://github.com/s1061123/sriov-network-device-plugin/tree/dev/prototype-netlink-status

ahalim-intel:
@s1061123 @zshi-redhat 
I have added a comment here: 
https://github.com/s1061123/sriov-network-device-plugin/commit/55ae86fd7fcf627ece2631ab40d6fe333f291b25#commitcomment-31301361

PR #34, opened 2018-11-09
removed /dev volume from daemonset spec

ahalim-intel:
the /dev voume is not needed and cause some unexpected issues.
Also removed the hostNetwork flag as well.

Change-Id: If5447a768ebb0361e2d36b3d9effc5b95965cdd8

### DIFF
--- a/images/sriovdp-daemonset.yaml
+++ b/images/sriovdp-daemonset.yaml
@@ -21,8 +21,6 @@ spec:
         tier: node
         app: sriovdp
     spec:
-      hostNetwork: true
-      hostPID: true
       nodeSelector:
         beta.kubernetes.io/arch: amd64
       tolerations:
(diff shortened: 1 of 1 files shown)


images/sriovdp-daemonset.yaml
```diff
@@ -21,8 +21,6 @@ spec:
         tier: node
         app: sriovdp
     spec:
-      hostNetwork: true
-      hostPID: true
```
booxter: This breaks dp startup:

```
I1112 19:41:58.694322      21 main.go:65] Starting all servers...
I1112 19:41:58.694684      21 server.go:165] starting sriov_net_vfio device plugin endpoint at: sriov_net_vfio.sock
E1112 19:41:58.694805      21 server.go:169] error starting sriov_net_vfio device plugin endpoint: listen unix /var/lib/kubelet/device-plugins/sriov_net_vfio.sock: bind: address already in use
E1112 19:41:58.695528      21 main.go:67] error starting resource servers listen unix /var/lib/kubelet/device-plugins/sriov_net_vfio.sock: bind: address already in use
```

ahalim-intel: I don't think this error related to the above changes. Can you please double check that there is no instances of DP already running with the same config file in your system?

PR #36, opened 2018-11-12
Use host PID namespace

booxter:
This fixes plugin startup failure:

error starting resource servers listen unix
/var/lib/kubelet/device-plugins/sriov.sock: bind: address already in use

### DIFF
--- a/images/sriovdp-daemonset.yaml
+++ b/images/sriovdp-daemonset.yaml
@@ -21,6 +21,7 @@ spec:
         tier: node
         app: sriovdp
     spec:
+      hostPID: true
       nodeSelector:
         beta.kubernetes.io/arch: amd64
       tolerations:


images/sriovdp-daemonset.yaml
```diff
@@ -21,6 +21,7 @@ spec:
         tier: node
         app: sriovdp
     spec:
+      hostPID: true
```
zshi-redhat: I didn't see why this is solving the problem of bind address already in use, could you please explain it a little bit more?

ahalim-intel: Please see https://github.com/intel/sriov-network-device-plugin/pull/34#discussion_r232981011

booxter: I now have a suspicion that this has something to do with the way we deploy our clusters. I experience permission issues with SR-IOV CNI plugin too in the same cluster: https://github.com/intel/sriov-cni/pull/24 while you cannot reproduce it on yours. There must be something in your cluster that grants additional privileges to the daemonsets that I don't have configured.

booxter:
When cluster deployed with kubeadm, the proposed change is NOT needed. I suspect it's some difference between kubeadm and local-cluster-up.sh. I don't think we should merge this patch if it works as-is for kubeadm.

PR #37, opened 2018-11-15
fix endless waiting

zshi-redhat:
Previously sm.Probe() is evaluated first in the
ListAndWatch for loop so there is no problem by
adding continue in the select case of 10 seconds;
but with recent change, sm.Probe() is moved under
select which causes an endless 10 seconds waiting
and Probe() is never invoked until the termSignal
is received. Fix it by removing continue.

### DIFF
--- a/cmd/sriovdp/sriov-device-plugin.go
+++ b/cmd/sriovdp/sriov-device-plugin.go
@@ -400,7 +400,6 @@ func (sm *sriovManager) ListAndWatch(empty *pluginapi.Empty, stream pluginapi.De
 	for {
 		select {
 		case <-time.After(10 * time.Second):
-			continue
 		case <-sm.termSignal:
 			glog.Infof("Terminate signal received, exiting ListAndWatch.")
 			return nil
--- a/cmd/sriovdp/sriov-device-plugin.go
+++ b/cmd/sriovdp/sriov-device-plugin.go
@@ -398,12 +398,6 @@ func (sm *sriovManager) ListAndWatch(empty *pluginapi.Empty, stream pluginapi.De
 	// Probes device state every 10 seconds and updates if changed.
 	// Terminates when termSignal received.
 	for {
-		select {
-		case <-time.After(10 * time.Second):
-		case <-sm.termSignal:
-			glog.Infof("Terminate signal received, exiting ListAndWatch.")
-			return nil
-		}
 		if sm.Probe() {
 			resp := new(pluginapi.ListAndWatchResponse)
 			for _, dev := range sm.devices {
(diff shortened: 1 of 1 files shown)


cmd/sriovdp/sriov-device-plugin.go
```diff
@@ -400,7 +400,6 @@ func (sm *sriovManager) ListAndWatch(empty *pluginapi.Empty, stream pluginapi.De
 	for {
 		select {
 		case <-time.After(10 * time.Second):
-			continue
```
ahalim-intel: Good catch! 
But now we are not doing anything on `case <-time.After(10 * time.Second):`. I think solution to the problem better solved by moving the Probe() before the event handling something like this:

```
	for {
		if sm.Probe() {
			resp := new(pluginapi.ListAndWatchResponse)
			for _, dev := range sm.devices {
				resp.Devices = append(resp.Devices, &pluginapi.Device{ID: dev.ID, Health: dev.Health})
			}
			glog.Infof("ListAndWatch: send devices %v\n", resp)
			if err := stream.Send(resp); err != nil {
				glog.Errorf("Error. Cannot update device states: %v\n", err)
				sm.grpcServer.Stop()
				return err
			}
		}
		
		select {
		case <-time.After(10 * time.Second):
			continue
		case <-sm.termSignal:
			glog.Infof("Terminate signal received, exiting ListAndWatch.")
			return nil
		}
	}
```
This way we can ensure Probe() invoked first when we get into to event watch loop and 10 seconds after we Probe() again. Let me know what you think

zshi-redhat: > Good catch!
> But now we are not doing anything on `case <-time.After(10 * time.Second):`. I think solution to the problem better solved by moving the Probe() before the event handling something like this:
> 
> ```
> 	for {
> 		if sm.Probe() {
> 			resp := new(pluginapi.ListAndWatchResponse)
> 			for _, dev := range sm.devices {
> 				resp.Devices = append(resp.Devices, &pluginapi.Device{ID: dev.ID, Health: dev.Health})
> 			}
> 			glog.Infof("ListAndWatch: send devices %v\n", resp)
> 			if err := stream.Send(resp); err != nil {
> 				glog.Errorf("Error. Cannot update device states: %v\n", err)
> 				sm.grpcServer.Stop()
> 				return err
> 			}
> 		}
> 		
> 		select {
> 		case <-time.After(10 * time.Second):
> 			continue
> 		case <-sm.termSignal:
> 			glog.Infof("Terminate signal received, exiting ListAndWatch.")
> 			return nil
> 		}
> 	}
> ```
> This way we can ensure Probe() invoked first when we get into to event watch loop and 10 seconds after we Probe() again. Let me know what you think

Move Probe() on top is an option, the reason why I didn't do that is because when ListAndWatch is invoked, an initial ListAndWatchResponse is sent back immediately to kubelet, there seems no need to do the Probe() right after the initial response;  but I admit that this is not a big overhead and will only happen once.

PR #38, opened 2018-11-16
avoid device-plugin to scan vfs from vfs

bmeng:
Currently, the device-plugin will scan all the devices under `/sys/class/net/`

But it looks meaningless to get the sriov_numvfs value from a vf.

Get the physical functions list and scan them only to avoid that.

### DIFF
--- a/cmd/sriovdp/sriov-device-plugin.go
+++ b/cmd/sriovdp/sriov-device-plugin.go
@@ -79,13 +79,23 @@ func getSriovPfList() ([]string, error) {
 		return sriovNetDevices, err
 	}
 
-	if len(netDevices) < 1 {
-		glog.Errorf("Error. No network device found in %s directory", netDirectory)
+	pfNetDevices := []string{}
+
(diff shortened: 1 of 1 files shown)


bmeng:
@zshi-redhat Please help review.

zshi-redhat:
Thanks Bo for working on the PR and help improve the code!
It looks that in order to avoid checking VF capability from a VF, this PR introduces another loop to first get SR-IOV capable PF. I think in theory this won't improve any code performance since it'll anyway loop against all the files, but I'm fine with it because it makes the logic more clear that we only get VF from a real PF, @ahalim-intel wdyt?

ahalim-intel:
I agree with @zshi-redhat It is not much of improvement in performance wise really. Same thing can be said against checking all network interfaces  if they are non-VF. I don't mind either way.

zshi-redhat:
Talked with @bmeng and we agreed to close this PR since it won't improve much code performance.
please feel free to re-open this PR if needed. Thank you for your contribution!

PR #39, opened 2018-11-20
Env variable name with proposed scheme

ahalim-intel:
This PR is based on [discussion](https://github.com/intel/sriov-network-device-plugin/pull/26#discussion_r229915970) on Env variable naming scheme.

Env variable with device IDs will be exported in following format: `PCIDEVICE_INTEL_COM_NET_A=0000:87:10.0;0000:87:11.6`

### DIFF
--- a/pkg/resources/genericPool.go
+++ b/pkg/resources/genericPool.go
@@ -27,15 +27,19 @@ import (
 	GetMounts()
 */
 type genericResourcePool struct {
+	resourcePool
 }
 
 func newGenericResourcePool(rc *types.ResourceConfig) types.ResourcePool {
-	return &resourcePool{
-		config:        rc,
-		devices:       make(map[string]*pluginapi.Device),
-		deviceFiles:   make(map[string]string),
-		IBaseResource: &genericResourcePool{},
+	this := &genericResourcePool{
+		resourcePool: resourcePool{
+			config:      rc,
+			devices:     make(map[string]*pluginapi.Device),
+			deviceFiles: make(map[string]string),
+		},
 	}
+	this.IBaseResource = this
+	return this
 }
 
 // Overrides GetDeviceFile() method
@@ -45,7 +49,7 @@ func (rp *genericResourcePool) GetDeviceFile(dev string) (devFile string, err er
 	return // empty string
 }
 
-func (rp *genericResourcePool) GetEnvs(resourceName string, deviceIDs []string) map[string]string {
+func (rp *genericResourcePool) GetEnvs(deviceIDs []string) map[string]string {
 	glog.Infof("generic GetEnvs() called")
 	envs := make(map[string]string)
 	values := ""
@@ -57,7 +61,7 @@ func (rp *genericResourcePool) GetEnvs(resourceName string, deviceIDs []string) 
 		}
 		values += " "
 	}
-	envs[resourceName] = values
+	envs[rp.config.ResourceName] = values
 	return envs
 }
 
--- a/pkg/resources/genericPool.go
+++ b/pkg/resources/genericPool.go
@@ -59,7 +59,7 @@ func (rp *genericResourcePool) GetEnvs(deviceIDs []string) map[string]string {
 		if i == lastIndex {
 			break
 		}
-		values += " "
+		values += ","
 	}
 	envs[rp.config.ResourceName] = values
 	return envs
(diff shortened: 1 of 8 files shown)


ahalim-intel:
@booxter  Please review this.

pkg/resources/genericPool.go
```diff
@@ -55,9 +59,9 @@ func (rp *genericResourcePool) GetEnvs(resourceName string, deviceIDs []string)
 		if i == lastIndex {
 			break
 		}
-		values += " "
+		values += ";"
```
booxter: OK so we are going back to the old format too. Good to know since I needed to patch kubevirt for the spaced format, now I need to revert back.

ahalim-intel: @booxter 
We need to replace periods  with something, as they are not acceptable in shell variable names anyway. http://tldp.org/LDP/abs/html/gotchas.html 

Spaces are not good choice as delimiters in Env Variable. In shell spaces are fine as value but no in Env. I have seen some issues already. Needed to change this regardless of the naming scheme.

booxter: I agree we need to replace dots for exactly the reason you mentioned. I haven't seen particular issues with spaces in my testing and kubevirt integration though I agree regardless that it makes sense to revert back to commas or semicolons. Though if I were you I would pick commas since that's what we used before multiendpoints branch was merged. While the updated DP is not in general backwards compatible with the previous behavior, it could help some consumers if you stick to the previous delimiter. For example, see this patch in kubevirt that we needed to adopt to the new space splitting: https://github.com/kubevirt/kubevirt/pull/1692/commits/e370ef884190a51f80f354dcdc5bd455612fea10 If you go with ; instead of , we'll need a similar patch. If you go with comma, we don't need any change.

zshi-redhat: +1 on changing the semicolon back to comma which is used in release-v1 as well.
btw, thanks for the shell scripting gotchas link!

ahalim-intel: > +1 on changing the semicolon back to comma which is used in release-v1 as well.
> btw, thanks for the shell scripting gotchas link!

Ok, we will use "'," instead.

zshi-redhat:
ok, this seems unnecessary as IBaseResource in resourcePool gets promoted automatically to the parent types(netDeviceResourcePool, etc) which I guess calling this.IBaseResource is the same as this.resourcePool.IBaseResource.

zshi-redhat:
> With this change, do we need to remove 'types.IBaseResource' from resourcePool struct in pool_stub.go and add the same into each specific resource structures, such as vfio/uio/netDevice/generic ResourcePool?

ok, this seems unnecessary as IBaseResource in resourcePool gets promoted automatically to the parent types(netDeviceResourcePool, etc) which I guess calling this.IBaseResource is the same as this.resourcePool.IBaseResource.

pkg/types/types.go
```diff
@@ -74,7 +74,7 @@ type ResourcePool interface {
 type IBaseResource interface {
 	GetDeviceFile(dev string) (devFile string, err error)
 	GetDeviceSpecs(deviceFiles map[string]string, deviceIDs []string) []*pluginapi.DeviceSpec
-	GetEnvs(resourceName string, deviceIDs []string) map[string]string
```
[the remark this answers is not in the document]
ahalim-intel: The IBaseResource interface is there so that the device specific information can be handled by some concrete resource pool types. Even though current implementation of netdevice,uio and vfio kind of send similar information, it gives more flexibility in future should we want to send more customized env var info. For example, we might want to ad _UIO" in the env var name or in the value.

To answer other question, The ResourcePool type is composition of some common methods(served by pool_stub.go) and IBaseResource interface. So when instantiate a concrete resourcPool each needs to implement its own methods for that interface. By embedding the stub resourcePool in a concrete resource allows access to its members directly.  Otherwise we need to get this information through parameters and method calls. (See how this was done before this changes)

Inheritance and polymorphism is not straight forward in Go.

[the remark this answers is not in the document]
zshi-redhat: Yep, it's reasonably if we want to leave some room for further adjustment of env var per different types of resource.

zshi-redhat:
LGTM

ahalim-intel:
Squashed one commit locally as its good idea to keep some these commits for reference.

PR #46, opened 2018-12-04
Pull the latest image for daemon set

booxter:
With policy = Never, the image won't be pulled by kubelet, meaning the
deployment will fail to start pods due to missing image in registry.


ahalim-intel:
It makes sense to remove this as we are now publishing the docker image. @zshi-redhat Can we fix the make image issue for travis?

Also, we don't have tagged version of the sriovdp image. It would be great to have some stable releases available to pull from instead of latest.

PR #47, opened 2018-12-05
fix golang package install in Dockerfile

zshi-redhat:


ahalim-intel:
Thanks for the PR. We can see it build the images in travis already. Also, I've tested it locally and see no issues.

PR #49, opened 2018-12-06
build docker image from alpine base image

ahalim-intel:
Switching base image to alpine which is lightweight and we can get golang environment.  This change also reduces the docker build time and the final image size(17.4MB instead of 521MB)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,16 @@
-FROM centos:centos7
+FROM golang:alpine as builder
 
 ADD . /usr/src/sriov-network-device-plugin
 
 ENV HTTP_PROXY $http_proxy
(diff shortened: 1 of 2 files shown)


zshi-redhat:
> Thanks @zshi-redhat for your very good point. Does travis-ci support multi-stage docker build? If it does then we could merge this PR as default docker build. This way we will have a smaller image available in docker hub should anyone want to pull that instead of building locally. We could also have an alternative Dockerfile for older versions too.

Yes, I think so, as the travis testing for this PR that use the new Dockerfile has passed.
And we could add old Dockerfile support in later PR.

booxter:
For the record, this broke the image. Fix: https://github.com/intel/sriov-network-device-plugin/pull/52

Issue #40, opened 2018-11-21
In PfOnly mode, allocated devices be reported as Unhealthy state


zshi-redhat:
When PF device is allocated to Pod, it's no longer visible in the host default namespace which may cause failure when checking link state of that device, DP in turn reports unhealthy of that device to kubelet, this will reduce one of the allocatable devices reported by kubelet and make one remaining healthy device un-allocatable.

ahalim-intel:
This has been fixed with #50

PR #50, opened 2018-12-06
fix PF only link health check

ahalim-intel:
this change will fix issue describe in #40

Change-Id: I8272bc711b73a35f6ea619bf1016bea2f408ea95

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -116,17 +116,17 @@ func GetSriovVFcapacity(pf string) int {
 	return numvfs
 }
 
-// IsNetlinkStatusUp returns 'false' if 'operstate' is not "up" for a Linux netowrk device
+// IsNetlinkStatusUp returns 'false' if 'operstate' is not "up" for a Linux network device.
+// This function will only return 'false' if the 'operstate' file of the device is readable
+// and holds value anything other than "up". Or else we assume link is up.
 func IsNetlinkStatusUp(dev string) bool {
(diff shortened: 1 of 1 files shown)


ahalim-intel:
@zshi-redhat Thanks for review. Its not ideal check but just an workaround. I'm looking into netlink based health check. But for that the device plugin needs run in host network namespace with privileged context.
Any opinion on that? Obviously there seem to be issue with SELinux enabled. #51

zshi-redhat:
> @zshi-redhat Thanks for review. Its not ideal check but just an workaround. I'm looking into netlink based health check. But for that the device plugin needs run in host network namespace with privileged context.
> Any opinion on that? Obviously there seem to be issue with SELinux enabled.
> See #51

I think for the selinux issue, it happens when device plugin runs in un-privileged securityContext, we may need to add additional selinux policy rules in order for the socket file be accessible by device plugin, those policy rules I guess are not handled by device plugin itself, rather the host level system shall take care of the configuration before SR-IOV device plugin is launched. however, I'm not familiar with how to define these rules, need further investigation. 

Meanwhile, I'm not against running device plugin with privileged mode since it eases the use of device plugin in development environment, and a lot of other features that we are adding may benefit from running with privileged mode.

booxter:
I don't think going with full priviliges for the device plugin container is the right way to go in production environment. We should be able to configure the container to use spc_t type, or even better, request creation of a new type specifically for device plugins. I am trying to understand now what's the procedure for the latter. As for spc_t, I am playing with it locally, and it seems to work (I no longer see denied messages in audit.log; currently running the fix in our kubevirt CI to validate it on OpenShift multi-node cluster.

PR #52, opened 2018-12-07
Don't use bash to run /entrypoint.sh

booxter:
Now that we switched to alpine that doesn't have bash installed, the
script fails to execute.

There were a bunch of bashisms in the script (detected by
checkbashisms tool), specifically:

1. built-in posix shell echo doesn't support -e. To overcome this, we
execute /bin/echo that is busybox version of echo guaranteed to support
it.

2. `function` keyword is not part of posix shell syntax. Just remove it.

With that, we should be able to safely switch shebang to /bin/sh.

### DIFF
--- a/images/entrypoint.sh
+++ b/images/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 set -e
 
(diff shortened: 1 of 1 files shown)


booxter:
@ahalim-intel @zshi-redhat please take a look, this un-breaks the latest image.

ahalim-intel:
Thanks for quick fix!

zshi-redhat:
++, thanks for the fix!

PR #57, opened 2019-01-08
add Dockerfile.rhel7 for OCP build

zshi-redhat:


zshi-redhat:
Should this change only impact the RHEL/OpenShift specific builds on release-v1 branch.

PR #64, opened 2019-01-24
Fix flannel-network.yaml

moshe010:
Update flannel-network.yaml with k8s.cni.cncf.io/v1
and NetworkAttachmentDefinition

Change-Id: I7916688e75b6679a2f3fa500bf3918179646c5ad


zshi-redhat:
/lgtm
Thanks for the PR!

This also reminds me that current template names under deployment folder are not that consistent, for example, for net-attach-def custom resource, we have ``sriov-crd.yaml`` and ``flannel-network.yaml``; for device plugin configuration files, there are config(2).json and config-node1.yaml (configMap), we may want to have similar conventions for each type of template file, which can be fixed with separate commits.

PR #67, opened 2019-01-27
Remove labels from pod-tc yaml

moshe010:
The doc is not containing steps to attach label to the node
so to make this pod to work we need to remove the labels
from the pod-tc yaml files

Change-Id: I11d6f24f6e8354a694cb55cbfdd17087de9f4348
Signed-off-by: Moshe Levi <moshele@mellanox.com>

### DIFF
--- a/deployments/pod-tc1.yaml
+++ b/deployments/pod-tc1.yaml
@@ -2,8 +2,6 @@ apiVersion: v1
 kind: Pod
 metadata:
   name: testpod1
-  labels:
-    env: test
   annotations:
     k8s.v1.cni.cncf.io/networks: sriov-net1
 spec:
--- a/deployments/pod-tc2.yaml
+++ b/deployments/pod-tc2.yaml
@@ -2,8 +2,6 @@ apiVersion: v1
 kind: Pod
 metadata:
   name: testpod2
-  labels:
-    env: test
   annotations:
     k8s.v1.cni.cncf.io/networks: sriov-net1
 spec:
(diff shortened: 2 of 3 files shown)


zshi-redhat:
I think these labels are decrators for Pod which should not impact scheduling of Pod on different nodes.

moshe010:
according to [1] to make it work with labels you need to first to attach the label to the node. On my setup that was the issue and when I removed the labels  it worked fine

[1] - https://kubernetes.io/docs/concepts/configuration/assign-pod-node/

zshi-redhat:
> according to [1] to make it work with labels you need to first to attach the label to the node. On my setup that was the issue and when I removed the labels it worked fine
> 
> [1] - https://kubernetes.io/docs/concepts/configuration/assign-pod-node/

yea, wondering why we need to label the node as nodeSelector is not used in Pod spec.

zshi-redhat:
> > according to [1] to make it work with labels you need to first to attach the label to the node. On my setup that was the issue and when I removed the labels it worked fine
> > [1] - https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
> 
> yea, wondering why we need to label the node as nodeSelector is not used in Pod spec.

well, I mean I didn't see why label here is a problem for creating Pod.
btw, if we decide to remove the labels in pod spec, then probably also remove them in ``deployments/kernel-net-demo``.

ahalim-intel:
These are sample deployments anyway, not necessarily will work on each and every cluster without customizing/modifying them according to the needs. Saying that, we can try to always keep these sample files with minimal constraints if possible.

I agree with @zshi-redhat we should revisit other sample deployment files and remove anything that are to specific.

PR #72, opened 2019-01-28
Fix starting device plugin issue

### LINKED ISSUE #71: Use "sh" instead of "bash" in pod-sriovdp.yaml template
Since the image that is being used is "alpine" this `command: [ "/bin/bash", "-c", "--" ]` should be `command: [ "/bin/sh", "-c", "--" ]`, also please consider updating the README.md with this change.

Mmduh-483:
Due to change to alpine image and since "bash" command not
existing it shoud be changed to "sh" command

Fixes #71

### DIFF
--- a/deployments/pod-sriovdp.yaml
+++ b/deployments/pod-sriovdp.yaml
@@ -7,13 +7,13 @@ spec:
   - name: sriov-device-plugin 
     image: nfvpe/sriov-device-plugin:latest
     imagePullPolicy: IfNotPresent
-    command: [ "/bin/bash", "-c", "--" ]
-    args: [ "while true; do sleep 300000; done;" ]
+    command: [ "/bin/sh", "-c", "--" ]
+    args: [ "sriovdp --logtostderr 10;" ]
     volumeMounts:
     - mountPath: /var/lib/kubelet/device-plugins/ 
       name: devicesock 
       readOnly: false
-    - mountPath: /sys/class/net
+    - mountPath: /sys
       name: net
       readOnly: true
   volumes:
@@ -23,6 +23,6 @@ spec:
      path: /var/lib/kubelet/device-plugins/
   - name: net
     hostPath:
-      path: /sys/class/net
+      path: /sys
   hostNetwork: true
   hostPID: true


zshi-redhat:
Thanks for the fix!  This template is so outdated.
Would you mind also fix other issues in the ``pod-sriovdp.yaml`` template?
for example:
1) the ``/etc/pcidp/config.json`` config file is not mounted to container
2) ``/sys/class/net`` is not used as mountPath to discover network devices, it should be ``/sys`` now.
3) ``pod-sriovdp.yaml`` doesn't execute ``sriovdp`` binary by default, perhaps adding a real command and args.

Mmduh-483:
@zshi-redhat 
Thanks for you comment
1- Regarding `/etc/pcidp/config.json` I did that in PR #74 
2- For the `net` volume is changed from `/sys/class/net` to `/sys` and it is confirmed to be working fine, right?
3- sriovdp binary will start working by default when create a the pod now, is there any default arguments you want to add  `/bin/sriovdp --key1 val1 --key2 val2`? or just running the binary with its default arguments `./bin/sriovdp`?

zshi-redhat:
> @zshi-redhat
> Thanks for you comment
> 1- Regarding `/etc/pcidp/config.json` I did that in PR #74
> 2- For the `net` volume is changed from `/sys/class/net` to `/sys` and it is confirmed to be working fine, right?
Yes, it should work with ``/sys`` now as we are not using ``/sys/class/net`` but ``/sys/bus/pci/device``, see [here](https://github.com/intel/sriov-network-device-plugin/blob/master/pkg/utils/utils.go#L29). If it doesn't work, then there is a bug.

> 3- sriovdp binary will start working by default when create a the pod now, is there any default arguments you want to add `/bin/sriovdp --key1 val1 --key2 val2`? or just running the binary with its default arguments `./bin/sriovdp`?
Please take a look at the [sriovdp-daemonset](https://github.com/intel/sriov-network-device-plugin/blob/master/images/sriovdp-daemonset.yaml) template which runs sriov device plugin as a DaemonSet, it has all the necessary args and mount volumes set, and it's the recommended way to run sriov device plugin. This pod-sriovdp I think is more for development testing purpose.

Mmduh-483:
@zshi-redhat I added the changes you asked except for mount `/etc/pcidp` since it is in  PR  #74

deployments/pod-sriovdp.yaml
```diff
@@ -7,13 +7,13 @@ spec:
   - name: sriov-device-plugin 
     image: nfvpe/sriov-device-plugin:latest
     imagePullPolicy: IfNotPresent
-    command: [ "/bin/bash", "-c", "--" ]
-    args: [ "while true; do sleep 300000; done;" ]
+    command: [ "/bin/sh", "-c", "--" ]
+    args: [ "sriovdp --logtostderr 10; while true; do sleep 300000; done;" ]
```
zshi-redhat: I think sriovdp will run Infinitely until it receives the ``sigCh`` signal, so perhaps remove the ``while`` loop in the command args?

ahalim-intel:
@Mmduh-483 I tried to resolve conflicts after merging #74 . Can you resolve this locally and update this PR. Better maybe create new clean PR after a rebase and close this one.

Mmduh-483:
@ahalim-intel , due to some issue with my repository I will do this in another pull request #81

PR #78, opened 2019-01-29
allow specifying NIC models to be discovered

zshi-redhat:
- add command line parameter ``nic-model``
-- ``nic-model`` parameter can be added multiple times to enable different NIC models
-- ``nic-model`` parameter is in the format of ``<vendorID>-<deviceID>``, e.g. ``0x8086-0x10fb``
- change default ``resourceName`` to ``netdev/sriov``

### DIFF
--- a/cmd/sriovdp/sriov-device-plugin.go
+++ b/cmd/sriovdp/sriov-device-plugin.go
@@ -48,9 +48,6 @@ const (
 	resourceName         = "netdev/sriov"
 )
 
-// default NIC Model to be discovered
-var defaultNicModel = []string{"0x8086-0x10fb", "0x8086-0x154c"}
-
 type arrayFlags []string
 
 func (a *arrayFlags) String() string {
@@ -195,6 +192,7 @@ func (sm *sriovManager) getPfWhiteList(pfList []string) []string {
 func (sm *sriovManager) discoverNetworks() error {
 
 	var healthValue string
+	pfWhiteList := []string{}
 	sm.rootDevices = []string{}
 
 	// Get a list of SRIOV capable NICs in the host
@@ -208,7 +206,13 @@ func (sm *sriovManager) discoverNetworks() error {
 		return nil
 	}
 
-	pfWhiteList := sm.getPfWhiteList(pfList)
+	if len(sm.cliParams.nicModel) < 1 {
+		glog.Infof("Discovering all capable and configured devices")
+		pfWhiteList = pfList
+	} else {
+		glog.Infof("Discovering all enabled NIC models")
+		pfWhiteList = sm.getPfWhiteList(pfList)
+	}
 
 	for _, dev := range pfWhiteList {
 		sriovcapablepath := filepath.Join(netDirectory, dev, "device", sriovCapable)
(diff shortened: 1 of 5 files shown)


moshe010:
Sure, @Mmduh-483 can you test it.

pkdas:
We have tested ConnectX-5 with and without DPDK using multus, sriovdp, sriov. But note that ConnectX-5 requires the kernel driver to bound always whether or not DPDK is going to be bound by an application in a container

pkdas:
Can I specify NIC model in /etc/pcidp/config.json?

zshi-redhat:
> Can I specify NIC model in /etc/pcidp/config.json?
This is not yet supported on master branch, but @ahalim-intel is working that.

zshi-redhat:
Two questions for this PR

> We have tested ConnectX-5 with and without DPDK using multus, sriovdp, sriov. But note that ConnectX-5 requires the kernel driver to bound always whether or not DPDK is going to be bound by an application in a container

Thanks @pkdas for the notes!
The testing mentioned above for Mellanox cards is more about the command line argument ``--nic-model``; ideally we would like to see that once ``--nic-model`` is specified when running sriovdp binary, the device plugin will only discover the nic models specified by ``--nic-model`` so others nic models will not be discovered or exposed by device plugin.
The default nic models if not configured will be set to ``string{"0x8086-0x10fb", "0x8086-0x154c"}`` for now.

zshi-redhat:
Two questions I got for this PR:
1) can we agree on a general ``resourceName`` format that is vendor neutral?  currently it's set to ``netdev/sriov``.
2) what should be the default behavior for device plugin to discover NIC models? for example, should it automatically discover all the configured SR-IOV devices on the host by default or use a default NIC model list?

Mmduh-483:
@zshi-redhat I have tested it with Mellanox card without DPDK and it is working fine
I noticed that when one of the VFs is allocated there will be a log message like:
```
I0130 08:54:36.919373   24185 sriov-device-plugin.go:495] DeviceID in Allocate: 0000:02:00.6
I0130 08:54:36.919424   24185 sriov-device-plugin.go:509] PCI Addrs allocated: 0000:02:00.6,
```
but when I delete the pod that is using that VF and the VF become free, there will be no logs messages
can you can add some log messages to update that?

cmd/sriovdp/sriov-device-plugin.go
```diff
@@ -45,11 +45,31 @@ const (
 	pluginMountPath      = "/var/lib/kubelet/device-plugins"
 	kubeletEndpoint      = "kubelet.sock"
 	pluginEndpointPrefix = "sriovNet"
-	resourceName         = "intel.com/sriov"
+	resourceName         = "netdev/sriov"
 )
 
+// default NIC Model to be discovered
+var defaultNicModel = []string{"0x8086-0x10fb", "0x8086-0x154c"}
```
moshe010: I am not sure why we need a default card here. It think it will be better to let to user to configure what he want to use and if it empty is should take all cards. What do you think?

zshi-redhat: Agree, I think we want a general device plugin that works with various NIC models and not tied to certain NIC types, having a default list may cause confusion and require frequent change when adding more NIC models. Will submit a new version of commit to fix it.

zshi-redhat: Fixed with a follow-up commit, defaults to discover all capable & configured SR-IOV devices when ``nic-model`` is not specified.

zshi-redhat:
Tested locally, it worked in two ways:
1) when --nic-model is provided, device plugin only discovers devices of specified model
2) when --nic-model is not provided (default), device plugin discovers all the SR-IOV configured devices.
@moshe010 please let me know if any further comments. Thanks!

PR #82, opened 2019-02-10
WIP: Add rdma device type support

Mmduh-483:


zshi-redhat:
Did we have any discussion or conclusion on whether to support rdma as a new configurable option or new type? I saw there was an item in weekly resource management agenda, but no details.

zshi-redhat:
Any thoughts of how rdma configuration might fit with the new selector approach? for example, could it be a new driver type?

moshe010:
I not sure it fit driver, I was thinking to add selector of capabilities. I need to discuss this internally in mellanox . I hope to update on the next meeting

zshi-redhat:
> I not sure it fit driver, I was thinking to add selector of capabilities. I need to discuss this internally in mellanox . I hope to update on the next meeting

Is mounting of "/dev/infiniband" the only requirement to use rdma devices in container?
If so, is it possible to automatically detect whether device supports rdma or not, and always mount "/dev/infiniband" during pod creation?

PR #83, opened 2019-02-10
Fix typo in Dockerfile.rhel

Mmduh-483:


zshi-redhat:
LGTM
Thanks for the fix!

PR #84, opened 2019-02-10
Modify pod-sriovdp.yaml to run device plugin when create the pod

Mmduh-483:


zshi-redhat:
/LGTM
Thanks for the fix!
Later we might want to change the Readme file and use Daemonset for device plugin launching which is the recommended way.

PR #86, opened 2019-02-11
Support multi arch travis-ci job and docker image

Mmduh-483:
Add CI job for ppc64le machine and create a manifest for the image
in both amd64 and ppc64le archs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,19 @@
 language: go
 
+matrix:
+  include:
+  - name: amd64
+    os: linux
+    env: ARCH=amd64
+  - name: ppc64le
+    os: linux-ppc64le
+    env: 
+      - ARCH=ppc64le
+      - DOCKERFILE=images/Dockerfile.ppc64le
+
 before_install:
   - sudo apt-get update -qq
+  - sudo systemctl restart docker
 
 script:
   - make test
@@ -16,7 +28,14 @@ deploy:
   # Push image to Dockerhub on merge to master
   - provider: script
     skip_cleanup: true
-    script: docker push nfvpe/sriov-device-plugin
+    script: >
+     bash -c '
+     if [ "${ARCH}" == "amd64" ]; then
+       docker push nfvpe/sriov-device-plugin;
+     fi
+     docker tag nfvpe/sriov-device-plugin nfvpe/sriov-device-plugin:$ARCH;
+     docker push nfvpe/sriov-device-plugin:$ARCH;
+     echo done'
     on:
       branch: master
   # Push image to Dockerhub on tag
@@ -24,9 +43,13 @@ deploy:
     skip_cleanup: true
     script: >
       bash -c '
-      docker tag nfvpe/sriov-device-plugin nfvpe/sriov-device-plugin:$TRAVIS_TAG;
-      docker push nfvpe/sriov-device-plugin:$TRAVIS_TAG;
-      echo foo'
+      if [ "${ARCH}" == "amd64" ]; then
+        docker tag nfvpe/sriov-device-plugin nfvpe/sriov-device-plugin:"$TRAVIS_TAG";
+        docker push nfvpe/sriov-device-plugin:"$TRAVIS_TAG";
+      fi
+      docker tag nfvpe/sriov-device-plugin nfvpe/sriov-device-plugin:"$TRAVIS_TAG-$ARCH";
+      docker push nfvpe/sriov-device-plugin:"$TRAVIS_TAG-$ARCH";
+      echo done'
     on:
       tags: true
       all_branches: true
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ export GOBIN
 
 # Docker
 IMAGEDIR=$(BASE)/images
-DOCKERFILE=$(CURDIR)/images/Dockerfile
+DOCKERFILE?=$(CURDIR)/images/Dockerfile
 TAG=nfvpe/sriov-device-plugin
 # Accept proxy settings for docker 
 DOCKERARGS=
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@
       - [Deploy demo Pod connecting to pre-created SR-IOV networks](#deploy-demo-pod-connecting-to-pre-created-sr-iov-networks)
       - [Verify status and the network connections of the demo Pod](#verify-status-and-the-network-connections-of-the-demo-pod)
     - [Pod device information](#pod-device-information)
+- [Multi Architecture Support](#multi-architecture-support)
 - [Issues and Contributing](#issues-and-contributing)
 
 ## SRIOV Network Device Plugin
@@ -450,6 +451,22 @@ The allocated device information is exported in Container's environment variable
 
 For example, if 2 devices are allocated from `intel.com/sriov` extended resource then the allocated device information will be found in following env variable:
 `PCIDEVICE_INTEL_COM_SRIOV=0000:03:02.1,0000:03:04.3`
+
+## Multi Architecture Support
+
+The supported architectures:
+* AMD64
+* PPC64LE
+
+Buiding image for AMD64:
+```
+$ DOCKERFILE=Dockerfile make image 
+```
+
+Buiding image for PPC64LE:
+```
+$ DOCKERFILE=images/Dockerfile.ppc64le TAG=nfvpe/sriov-device-plugin:ppc64le make image        
+```
 
 ## Issues and Contributing
 
--- a/deployments/k8s-v1.10-v1.15/sriovdp-daemonset.yaml
+++ b/deployments/k8s-v1.10-v1.15/sriovdp-daemonset.yaml
@@ -60,3 +60,59 @@ spec:
             items:
             - key: config.json
               path: config.json
+
+---
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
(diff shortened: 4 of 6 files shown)


ahalim-intel:
@s1061123 If you have time, can you please review this? Thanks!

Mmduh-483:
@ahalim-intel do you have any comments for this PR?

s1061123:
@ahalim-intel The diff seems to valid to me. BTW I am not good at sr-iov release strategy.
For now,  multus does release only x64 and ppc/arm are not supported yet.

Do we align with multus side as well (and arm could be also target?)

moshe010:
@s1061123, the plan is to do changes for muluts and sriov-cni

s1061123:
Do we need to change base image in Dockerfile because `FROM alpine` is for x64, right?

Mmduh-483:
@s1061123 no need to change the Dockerfile

.travis.yml
```diff
@@ -1,7 +1,22 @@
...
+  - name: amd64
+    os: linux
+    env: ARCH=amd64
+  - name: ppc64le
+    os: linux-ppc64le
+    env: ARCH=ppc64le
+
 before_install:
   - sudo apt-get update -qq
+  - sudo apt-get remove docker-ce
```
zshi-redhat: why ``remove`` and ``install`` docker-ce?

.travis.yml
```diff
@@ -1,7 +1,22 @@
...
+  - sudo apt-get install docker-ce
+  - sudo mkdir -p ~/.docker
+  - sudo echo '{"experimental":"enabled"}' | sudo tee ~/.docker/config.json
```
zshi-redhat: Is it necessary to config both ``experimental:enabled`` and ``experimental:true``?

.travis.yml
```diff
@@ -19,16 +34,28 @@ deploy:
...
   # Push image to Dockerhub on tag
   - provider: script
     skip_cleanup: true
     script: >
       bash -c '
-      docker tag nfvpe/sriov-device-plugin nfvpe/sriov-device-plugin:$TRAVIS_TAG;
-      docker push nfvpe/sriov-device-plugin:$TRAVIS_TAG;
+      docker tag nfvpe/sriov-device-plugin nfvpe/sriov-device-plugin:"$TRAVIS_TAG-$ARCH";
+      docker push nfvpe/sriov-device-plugin:"$TRAVIS_TAG-$ARCH";
+      sudo docker manifest create -a nfvpe/sriov-device-plugin:$TRAVIS_TAG nfvpe/sriov-device-plugin:"$TRAVIS_TAG-amd64" nfvpe/sriov-device-plugin:"$TRAVIS_TAG-ppc64le";
```
zshi-redhat: what does ``docker manifest`` do here? do we want to use a experimental command here?

.travis.yml
Mmduh-483: we need the docker version 18.09.0 or later which is supporting `docker manifest`, but the docker version provided from Travis is a bit older

.travis.yml
Mmduh-483: one for the server side `/etc/docker/daemon.json` and one for the client side `~/.docker/config.json`

.travis.yml
Mmduh-483: It is creating like a map when the user run `docker pull image:latest` it will return the image:amd64 if the machine arch is amd64 or ppc64le if it is the arch. of the machine

Mmduh-483:
> Can you also paste the multi-arch change in Multus and SR-IOV CNI?

It is on lt plan but first, we need approval for this one

zshi-redhat:
> > Can you also paste the multi-arch change in Multus and SR-IOV CNI?
> 
> It is on lt plan but first, we need approval for this one

@Mmduh-483 would you please first rebase this PR to see if it can pass CI?

Mmduh-483:
> @Mmduh-483 [here](https://github.com/intel/sriov-network-device-plugin/pull/117) is the change to remove glide install in `travis.yaml`; I'll rebase it to current `feature/device_selectors` branch, after which, all the PRs proposed against `feature/device_selectors` will need rebase again. does it sound good to you?

it's ok

.travis.yml
zshi-redhat: gotcha, can this be done with single cmd similar to ``apt-get upgrade docker-ce``? I barely use ubuntu, please forgive me if it's a silly question:-)

Dockerfile
```diff
@@ -5,6 +5,7 @@ ADD . /usr/src/sriov-network-device-plugin
 ENV HTTP_PROXY $http_proxy
 ENV HTTPS_PROXY $https_proxy
 RUN apk add --update make && \
+    apk add build-base && \
```
zshi-redhat: Why this package is required in Dockerfile, Is it needed to build multi-arch images?

Mmduh-483: To avoid the error in the image
travis error

Mmduh-483: for docker upgrade, I updated the command,
`apt-get install --only-upgrade docker-ce`

zshi-redhat:
@Mmduh-483 , post a question from @s1061123 : can we use the same entrypoint.sh script (with /bin/sh) for ppc64le arch in sriovdp image?

Mmduh-483:
> @Mmduh-483 , post a question from @s1061123 : can we use the same entrypoint.sh script (with /bin/sh) for ppc64le arch in sriovdp image?

yes

Mmduh-483:
@ahalim-intel is there any comments for this PR?

Mmduh-483:
@zshi-redhat @ahalim-intel can you guys please review the changes?

zshi-redhat:
> @zshi-redhat @ahalim-intel can you guys please review the changes?

It looks like `docker manifest` is not used in Multus, is there a particular reason for this?
@s1061123 any comments?

s1061123:
@zshi-redhat multus does not use it because we could filter to run daemonset in kubernetes side. In the future, we could support it when the feature is getting mature.

s1061123:
In addition, you may need to test that the container image could be run in each architecture....

Mmduh-483:
@zshi-redhat multus specified a different tag for each architecture in the daemonset file explicitly, `docker manifest` do the same thing but it is implicit, please give it a test

Mmduh-483:
@adrianchiris

adrianchiris:
> In addition, you may need to test that the container image could be run in each architecture....

@s1061123 either way (use docker manifest or split at k8s level) you will need to test the container image actually works right ?

@zshi-redhat  do you have a specific preference ?

adrianchiris:
i wonder if utilizing the manifest is container runtime specific ...

Mmduh-483:
We had an internal discussion and decided to not use `docker manifest` since it is still [experimental](https://docs.docker.com/engine/reference/commandline/manifest/)

Mmduh-483:
@zshi-redhat any other comments?

Dockerfile
```diff
@@ -4,7 +4,7 @@ ADD . /usr/src/sriov-network-device-plugin
 
 ENV HTTP_PROXY $http_proxy
 ENV HTTPS_PROXY $https_proxy
-RUN apk add --update --virtual build-dependencies build-base linux-headers && \
+RUN apk add --update --virtual build-dependencies build-base linux-headers make && \
```
zshi-redhat: Do we need explicitly specify `make`?

Makefile
```diff
@@ -20,7 +20,7 @@ export GOBIN
 
 # Docker
 IMAGEDIR=$(BASE)/images
-DOCKERFILE=$(CURDIR)/Dockerfile
+DOCKERFILE?=$(CURDIR)/Dockerfile
```
zshi-redhat: so this change is to allow `DOCKERFILE` be configured manually(e.g. to Dockerfile.ppc64le) when running `make`?

Dockerfile
Mmduh-483: for ppc it is needed

Makefile
Mmduh-483: yes, yes, another idea would be `make image_ppc64le`, not sure which one would be better

zshi-redhat: I'm fine with either way, perhaps worth document the build steps later in developer doc.

Dockerfile
zshi-redhat: So Dockerfile.ppc64le is for building ppc64le image, this one would be for amd64 and should not require `make`?

Mmduh-483: I agree

Mmduh-483:
@zshi-redhat Any comments for this PR?

images/sriovdp-daemonset.yaml
```diff
@@ -60,3 +60,59 @@ spec:
...
+            path: /var/lib/kubelet/
+        - name: log
+          hostPath:
+            path: /var/log
+        - name: config-volume
+          configMap:
+            name: sriovdp-config
+            items:
+            - key: config.json
+              path: config.json
```
ahalim-intel: What about having a separate damonset file for other architecture? I'm concerned about the size of this file if we start adding them all in one file.

Mmduh-483: @ahalim-intel in multus-cni both of the DS are in the same file, is ok for me to separate them?

Mmduh-483:
@zshi-redhat Are there other comments?

zshi-redhat:
@ahalim-intel would you please take a look?

PR #90, opened 2019-02-27
Add Basic Support for RDMA

Mmduh-483:
Check if i the device is supporting the rdma by checking if the path ` /sys/class/pci_bus/<busAddress>/device/<pciAddress>/infiniband` exists and if so, mount the directory `/dev/infiniband` to the pod requesting the resource

### DIFF
--- a/cmd/sriovdp/sriov-device-plugin.go
+++ b/cmd/sriovdp/sriov-device-plugin.go
@@ -38,8 +38,10 @@ import (
 
 const (
 	netDirectory    = "/sys/class/net/"
+	pciBusDirectory = "/sys/bus/pci/devices/"
 	sriovCapable    = "/sriov_totalvfs"
 	sriovConfigured = "/sriov_numvfs"
+	rdmaDevices     = "/dev/infiniband"
 
 	// Device plugin settings.
 	pluginMountPath      = "/var/lib/kubelet/device-plugins"
@@ -64,11 +66,16 @@ type cliParams struct {
 	nicModel arrayFlags
 }
 
+type vfDevice struct {
+	k8sDevice pluginapi.Device
+	isRdma    bool
+}
+
 // sriovManager manages sriov networking devices
 type sriovManager struct {
 	cliParams
 	socketFile  string
-	devices     map[string]pluginapi.Device // for Kubelet DP API
+	devices     map[string]vfDevice // for Kubelet DP API
 	rootDevices []string
 	grpcServer  *grpc.Server
 	termSignal  chan bool
@@ -79,7 +86,7 @@ func newSriovManager(cp *cliParams) *sriovManager {
 
 	return &sriovManager{
 		cliParams:   *cp,
-		devices:     make(map[string]pluginapi.Device),
+		devices:     make(map[string]vfDevice),
 		socketFile:  fmt.Sprintf("%s.sock", pluginEndpointPrefix),
 		termSignal:  make(chan bool, 1),
 		stopWatcher: make(chan bool),
@@ -188,6 +195,14 @@ func (sm *sriovManager) getPfWhiteList(pfList []string) []string {
 	return pfwl
 }
 
+func isRdmaDevice(pciAddr string) bool {
+	pciBusPath := filepath.Join(pciBusDirectory, pciAddr, "infiniband")
+	if _, err := os.Stat(pciBusPath); os.IsNotExist(err) {
+		return false
+	}
+	return true
+}
+
 //Reads DeviceName and gets PCI Addresses of VFs configured
 func (sm *sriovManager) discoverNetworks() error {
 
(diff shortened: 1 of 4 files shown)


cmd/sriovdp/sriov-device-plugin.go
```diff
@@ -515,6 +518,17 @@ func (sm *sriovManager) Allocate(ctx context.Context, rqt *pluginapi.AllocateReq
 		envmap["SRIOV-VF-PCI-ADDR"] = pciAddrs
 
 		containerResp.Envs = envmap
+
+		if isRdmaDirExists {
```
zshi-redhat: Is there a solid way to identify whether a certain device is rdma capable ?
I don't think we want to mount rdma device path to every pod even if device allocated to pod doesn't support rdma.
for example, if there are two types of device on the same host, one is rdma capable, the other is not, then we should avoid mounting rdma path to the other device.

Mmduh-483: @zshi-redhat I updated the code to check if the device is supporting rdma by checking the full path of  ` /sys/class/pci_bus/<pciBus>/device/<pciAddress>/infiniband`
e.g. ` /sys/class/pci_bus/0000\:02/device/0000:02:00.4/infiniband`

moshe010:
@Mmduh-483, let change the check if it /sys/bus/pci/devices/<pci address>/infiniband
for example /sys/bus/pci/devices/0000\:02\:00.0/infiniband exist then it is RDMA.
It will be much more simpler

cmd/sriovdp/sriov-device-plugin.go
```diff
@@ -515,6 +527,17 @@ func (sm *sriovManager) Allocate(ctx context.Context, rqt *pluginapi.AllocateReq
 		envmap["SRIOV-VF-PCI-ADDR"] = pciAddrs
 
 		containerResp.Envs = envmap
+
+		if isRdmaDevice(id){
```
zshi-redhat: This will only configure the rdma device if the last device in ``Allocate`` request is rdma capable. perhaps we shall move this into the for loop against ``container.DevicesIDs`` above. Further improvement might be that the rdma could be a property of device that discovered and recorded in ``discoverNetworks``  function and we only need to check whether the device has rdma property in the ``Allocate`` call, but this requires changes in the main ``sriovManager`` data structure.

zshi-redhat:
As a general guidance, I think we should wait until master branch has ``rdma`` supported, then merge this PR in release-v1 branch. @ahalim-intel wdyt?

deployments/pod-sriovdp.yaml
```diff
@@ -13,7 +13,7 @@ spec:
     - mountPath: /var/lib/kubelet/device-plugins/ 
       name: devicesock 
       readOnly: false
-    - mountPath: /sys/class/net
```
zshi-redhat: This change shall also go to ``images/sriovdp-daemonset.yaml`` file.

deployments/pod-sriovdp.yaml
```diff
@@ -23,6 +23,6 @@ spec:
      path: /var/lib/kubelet/device-plugins/
   - name: net
     hostPath:
-      path: /sys/class/net
```
zshi-redhat: ditto.

moshe010:
> As a general guidance, I think we should wait until master branch has `rdma` supported, then merge this PR in release-v1 branch. @ahalim-intel wdyt?
@zshi-redhat, as a guideline I agree, but we are  currently blocked to push this to master due to expected changes on master to support the HOMOGENEOUS DEPLOYMENT feature, which is doesn't have pull request yet. Given this, the eventual changes to the master would regardless be different due to the expected changes to the device plugin to support the new deployment feature.

We would really like for openshift 4.1 to have RDMA support, and need to find a path together to get this in. It will be really disappointing if we can't get this important feature into openshift 4.1...

cmd/sriovdp/sriov-device-plugin.go
```diff
@@ -493,20 +511,24 @@ func (sm *sriovManager) GetDevicePluginOptions(ctx context.Context, empty *plugi
 func (sm *sriovManager) Allocate(ctx context.Context, rqt *pluginapi.AllocateRequest) (*pluginapi.AllocateResponse, error) {
 	resp := new(pluginapi.AllocateResponse)
 	pciAddrs := ""
+	id := ""
```
ahalim-intel: Why the var `id` moved up instead of local to the loop? I don't see it's being used outside of the loop anyway. Maybe I'm missing something?

ahalim-intel:
> As a general guidance, I think we should wait until master branch has `rdma` supported, then merge this PR in release-v1 branch. @ahalim-intel wdyt?

As long as it doesn't create new issues with non-rdma devices I see no problem with it. But for realease-v1 I'll let you decide. I believe you'll need to test it thoroughly if any new features are added to this branch.

cmd/sriovdp/sriov-device-plugin.go
Mmduh-483: fixed

PR #93, opened 2019-02-28
Use device plugin image with release-v1 tag

### LINKED ISSUE #92: Deploying device plugin with  release-v1
when try to deploy the device plugin with `pod-sriovdp.yaml` file it will fail getting the following error

```
I0228 11:32:55.284854    6741 server.go:460] Event(v1.ObjectReference{Kind:"Pod", Namespace:"default", Name:"sriov-device-plugin", UID:"8bfe019c-3b4c-11e9-beb8-0025904fd47a", APIVersion:"v1", ResourceVersion:"118717", FieldPath:"spec.containers{sriov-device-plugin}"}): type: 'Warning' reason: 'Failed' Error: failed to start container "sriov-device-plugin": Error response from daemon: OCI runtime create failed: container_linux.go:348: starting container process caused "exec: \"/bin/bash\": stat /bin/bash: no such file or directory": unknown
```
This happen because the image `nfvpe/sriov-device-plugin:latest` is created from the master branch and use `alpine` image which doesn't have `/bin/bash`

Mmduh-483:
fixes #92

### DIFF
--- a/deployments/pod-sriovdp.yaml
+++ b/deployments/pod-sriovdp.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   containers:
   - name: sriov-device-plugin 
-    image: nfvpe/sriov-device-plugin:latest
+    image: nfvpe/sriov-device-plugin:release-v1
     imagePullPolicy: IfNotPresent
     command: [ "/bin/bash", "-c", "--" ]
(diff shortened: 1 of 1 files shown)


zshi-redhat:
Thanks for the fix! Yes, we should use ``release-v1`` image.

PR #96, opened 2019-03-01
add feature: dynamic discovery and grouping

ahalim-intel:
Not ready for merge; yet!
More updates and unit tests fixes on it's way.


ahalim-intel:
TODOs:

- Fix unit tests
- Update README
- Update sample deployment files
-

cmd/sriovdp/manager.go
```diff
@@ -143,21 +159,122 @@ func (rm *resourceManager) validConfigs() bool {
...
 
-	return true
+	if len(ifNames) > 0 { // there's at least one interface name found
+		for _, ifName := range ifNames {
+			link, err := netlink.LinkByName(ifName)
+			if err != nil {
+				glog.Errorf("expected to get valid host interface with name %s: %q", ifName, err)
+			}
+
+			routes, err := netlink.RouteList(link, netlink.FAMILY_V4) // IPv6 routes: all interface has at least one link local route entry
```
zshi-redhat: Is there way to detect if vfio/uio devices are in use?

pkg/resources/pciNetDevice.go
```diff
@@ -0,0 +1,115 @@
...
+		Health: pluginapi.Healthy,
+	}
+
+	// 			4. Create pciNetDevice object with all relevent info
+	return &pciNetDevice{
+		pciDevice:   pciDevice,
+		ifName:      "", // TO-DO: Get this using utils pkg
+		pfName:      "", // TO-DO: Get this using utils pkg
+		driver:      driverName,
+		vfID:        0,  // TO-DO: Get this using utils pkg if needed
```
zshi-redhat: This looks like a meaningful default num :-)

pkg/resources/factory.go
```diff
@@ -49,16 +50,70 @@ func (rf *resourceFactory) GetResourceServer(rp types.ResourcePool) (types.Resou
...
-	case "netdevice":
-		return newNetDevicePool(rc)
+		return newUioResourcePool()
 	default:
-		return newGenericResourcePool(rc)
+		return newNetDevicePool()
 	}
 }
+
+// GetResourcePool returns and instance of ResourcePool for a ResourceConfig
```
zshi-redhat: GetResourcePool -> GetSelector

zshi-redhat:
The changes of network device discovery and selector based configuration make me feel that it's now capable of supporting host devices.

pkg/resources/factory.go
```diff
@@ -49,16 +50,70 @@ func (rf *resourceFactory) GetResourceServer(rp types.ResourcePool) (types.Resou
 }
 
 // GetResourcePool returns and instance of ResourcePool for a ResourceConfig
-func (rf *resourceFactory) GetResourcePool(rc *types.ResourceConfig) types.ResourcePool {
-	glog.Infof("Resource pool type: %s", rc.DeviceType)
-	switch rc.DeviceType {
-	case "vfio":
-		return newVfioResourcePool(rc)
+func (rf *resourceFactory) GetInfoProvider(name string) types.DeviceInfoProvider {
```
przemeklal: GetResourcePool -> GetInfoProvider in the comment

pkg/resources/pool_stub.go
```diff
@@ -50,63 +45,59 @@ func (rp *resourcePool) InitDevice() error {
...
+
+func (rp *resourcePool) GetDeviceSpecs(deviceIDs []string) []*pluginapi.DeviceSpec {
+	glog.Infof("GetDeviceSpecs(): for devices: %v", deviceIDs)
+	devSpecs := make([]*pluginapi.DeviceSpec, 0)
+
+	// Add vfio group specific devices
+	for _, id := range deviceIDs {
+		if dev, ok := rp.devicePool[id]; ok {
+			ds := dev.GetDeviceSpecs()
+			devSpecs = append(devSpecs, ds...)
```
przemeklal: For vfio devices `dev.GetDeviceSpecs()` each device spec returned contains its specific mount and one for the default `/dev/vfio/vfio`. When multiple devices are passed here, in the final returned DeviceSpec this default `/dev/vfio/vfio` spec appears multiple times. Can this be an issue?

pkg/utils/utils.go
```diff
@@ -237,3 +270,37 @@ func GetUIODeviceFile(dev string) (devFile string, err error) {
...
+	return names, nil
+}
+
+// GetDriverName returns current driver attached to a pci device from its pci address
+func GetDriverName(pciAddr string) (string, error) {
+	driverLink := filepath.Join(sysBusPci, pciAddr, "driver")
+	driverInfo, err := os.Readlink(driverLink)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return pciAddr, nil
```
przemeklal: is it intended behaviour to return PCI address as the driver name in case the file doesn't exist? or is it leftover from copy-pasted `GetPfAddr` function?

README.md
```diff
@@ -71,8 +69,8 @@ The following  NICs were tested with this implementation. However, other SRIOV c
 	- VF driver: v3.2.2-k
 > please refer to Intel download center for installing latest [Intel-® 82599ES 10 Gigabit Ethernet](https://ark.intel.com/products/41282/Intel-82599ES-10-Gigabit-Ethernet-Controller) drivers
 
-- Mellanox ConnectX�-4 Lx EN Adapter
-- Mellanox ConnectX�-5 Adapter
+- Mellanox ConnectX�-4 Lx EN Adapter
```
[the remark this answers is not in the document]
ahalim-intel: > unrecognized special characters?

not sure what happen there, possibly editor's auto-formatting. Will revert it back. Thanks for flagging this.

README.md
```diff
@@ -243,6 +251,16 @@ We assume that you have working K8s cluster configured with Mutlus meta plugin f
...
 The [images](./images) directory contains example Docker file, sample specs along with build scripts to deploy the SRIOV device plugin as daemonset. Please see [README.md](./images/README.md) building docker the image.
 
+````
+# Create ConfigMap
+$ kubectl create -f images/configMap.yaml
+
+
+# Create sriov-device-plugin-daemonset
+$ kubectl create -f images/sriovdp-daemonset.yaml
+pod "testpod1" created
```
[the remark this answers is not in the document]
ahalim-intel: will fix this

cmd/sriovdp/manager.go
```diff
@@ -143,21 +159,122 @@ func (rm *resourceManager) validConfigs() bool {
...
+	}
+
+	devices := pci.ListDevices()
+	if len(devices) == 0 {
+		return fmt.Errorf("discoverDevices(): error: could not retrive PCI info: %v,", err)
+	}
+	for _, device := range devices {
+		devClass, err := strconv.ParseInt(device.Class.ID, 16, 64)
+		if err != nil {
+			return fmt.Errorf("discoverDevices(): error parsing device class %v", err)
```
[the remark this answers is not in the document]
ahalim-intel: Good point. I will change that. But we expect get device class info, an error here would be an exception really.

cmd/sriovdp/manager.go
```diff
@@ -143,21 +159,122 @@ func (rm *resourceManager) validConfigs() bool {
...
+	glog.Infoln("discovering host network devices")
+
+	pci, err := ghw.PCI()
+	if err != nil {
+		return fmt.Errorf("discoverDevices(): error getting PCI info: %v", err)
+	}
+
+	devices := pci.ListDevices()
+	if len(devices) == 0 {
+		return fmt.Errorf("discoverDevices(): error: could not retrive PCI info: %v,", err)
```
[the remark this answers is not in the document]
ahalim-intel: sure we could do that.

cmd/sriovdp/manager.go
```diff
@@ -143,21 +159,122 @@ func (rm *resourceManager) validConfigs() bool {
...
+				aPF := utils.IsSriovPF(device.Address)
+				aVF := utils.IsSriovVF(device.Address)
+
+				if aPF || !aVF {
+					// add to linkWatchList
+					rm.addToLinkWatchList(device.Address)
+				}
+
+				if aPF && utils.SriovConfigured(device.Address) {
+					// do not add this device in net device list
```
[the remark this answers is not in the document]
ahalim-intel: Yes, thats the idea. If a PF is configured with VFs then we should leave the PF out as allocable  resource. (We need the PF always reside in init namespace for CNI to setup it's VFs). Allowing a PF to move in a Pod will make the VF unusable. If PF is desired to be allocated directly in a Pod then no VF should be configured.

cmd/sriovdp/manager.go
ahalim-intel: I couldn't think of any other than querying the kubelet resource api. This may get little bit complicated

przemeklal:
@ahalim-intel I opened a PR on your fork with the tests for the resources package here https://github.com/ahalim-intel/sriov-network-device-plugin/pull/2

Currently tests are adjusted to make this pass https://github.com/intel/sriov-network-device-plugin/pull/96#discussion_r262849759 
Let me know if this needs to be updated or not.

In the meanwhile I'm starting to work on the new functions in the utils package if that's OK.

pkg/resources/pciNetDevice.go
ahalim-intel: The vfID is not being used at the moment. This is just a placeholder to set correct value later. even if we don't set it it will be '0' by default.

pkg/resources/factory.go
ahalim-intel: fixed

pkg/resources/factory.go
ahalim-intel: fixed

pkg/resources/pool_stub.go
ahalim-intel: Good point! Functionally this shouldn't be any issue. As this will just add the same device file into the same cgroup again. Ideally we should add a device file once in the overall deviceSpec. We can add a check - only add deviceSpec if its not in the spec list. I will add a fix. Thanks

pkg/utils/utils.go
ahalim-intel: We should return error for all errors during driver name look up. The driver name is required to make decision on creating resource pool. I will fix this. Thanks

cmd/sriovdp/manager.go
```diff
@@ -143,21 +159,123 @@ func (rm *resourceManager) validConfigs() bool {
...
+	if err != nil {
+		return false, fmt.Errorf("error trying get net device name for device %s", pciAddr)
 	}
 
-	return true
+	if len(ifNames) > 0 { // there's at least one interface name found
+		for _, ifName := range ifNames {
+			link, err := netlink.LinkByName(ifName)
+			if err != nil {
+				glog.Errorf("expected to get valid host interface with name %s: %q", ifName, err)
```
przemeklal: I know it's unlikely in a real-world scenario, but in the unit tests, if the `link` variable is `nil` here and then you call `netlink.RouteList`, `routes` array will contain full routing table of the host and the function returns `true`, meaning that the device is in use. I'd expect this function to return an error in such scenario.

ahalim-intel: if `link` is nil means there could be error looking up the link and it is safe to assume this is in use -> do not add in any resource pool.  The route list don't matter I suppose.

przemeklal: okay makes sense, the tests will assume that in this scenario expected outcome is that the link is in use, thanks

README.md
```diff
@@ -145,32 +143,38 @@ This plugin creates device plugin endpoints based on the configurations given in
...
-| "sriovMode"    | No       | Whether the root devices are SRIOV capable or not | `bool` - true OR false[default]                                     | `true`                   |
-| "deviceType"   | No       | Device driver type                                | `string` - "netdevice"\|"uio"\|"vfio"                               | `"netdevice"`            |
+
+
+|     Field      | Required |        Description        |                       Type - Accepted values                        |                          Example/Accepted values                          |
+|----------------|----------|---------------------------|---------------------------------------------------------------------|---------------------------------------------------------------------------|
+| "resourceName" | Yes      | Endpoint resource name    | `string` - must be unique and should not contain special characters | "sriov_net_A"                                                             |
+| "selectors"    | No       | A map of device selectors | Each selector is a map of  `string` list.                           | "vendors": ["8086"], "devices": ["154c", "10ed"], "drivers": ["vfio-pci"] |
```
moshe010: what is the default behaviour if you don't use selectors? Does it select all or it won't select anyone?

README.md
```diff
@@ -145,32 +143,38 @@ This plugin creates device plugin endpoints based on the configurations given in
...
-    "resourceList":
-    [
-        {
-            "resourceName": "sriov_net_A",
-            "rootDevices": ["02:00.0", "02:00.2"],
-            "sriovMode": true,
-            "deviceType": "netdevice"
+    "resourceList": [{
+            "resourceName": "intel_sriov_netdevice",
+            "selectors": {
```
moshe010: we have support code that update the  selectors to be a list of dict. 
so we can do 
            "selectors": [{
                "vendors": ["8086"],
                "devices": ["154c", "10ed"],
                "drivers": ["vfio-pci"]
            }, {
                "vendors": ["15b3"],
                "devices": ["1015"],
                "drivers": ["mlx5_core"]
            }
Can we add this to this PR?

moshe010: Also we have selector per netdevice which can be useful as well.

README.md
```diff
@@ -243,6 +251,21 @@ We assume that you have working K8s cluster configured with Multus meta plugin f
 
 The [images](./images) directory contains example Docker file, sample specs along with build scripts to deploy the SRIOV device plugin as daemonset. Please see [README.md](./images/README.md) building docker the image.
 
+````
+# Create ConfigMap
+$ kubectl create -f images/configMap.yaml
```
moshe010: shouldn't the configMap.yaml be under deployments folder as it just an example?

deployments/pod-sriovdp.yaml
```diff
@@ -9,6 +9,8 @@ spec:
     imagePullPolicy: IfNotPresent
     command: [ "/bin/sh", "-c", "--" ]
     args: [ "sriovdp --logtostderr 10;" ]
+    securityContext:
+      privileged: true
```
moshe010: why this is required? what  is required privileged

deployments/pod-sriovdp.yaml
```diff
@@ -19,16 +21,29 @@ spec:
     - mountPath: /etc/pcidp/
       name: config
       readOnly: true
+    - name: hwdata
+      mountPath: /usr/share/hwdata
```
moshe010: you are assume the lspci is installed? it is install in all OS on the same location?

README.md
ahalim-intel: > what is the default behaviour if you don't use selectors? Does it select all or it won't select anyone?

A resourcPool with no selectors will include all network devices initially discovered by host.

moshe010: ok so it wasn't clear in the readme, unless I missed ...

deployments/pod-sriovdp.yaml
ahalim-intel: > you are assume the lspci is installed? it is install in all OS on the same location?

The 'ghw' library uses /usr/share/hwdata/pci.ids file. This file is part of hwdata package available in rpm and deb. As a fallback the library will try to download this database from https://pci-ids.ucw.cz/v2.2/pci.ids.gz. 
We may as well add the 'hwdata' as a prerequisites on the host.

README.md
ahalim-intel: > we have support code that update the selectors to be a list of dict.
> so we can do
> "selectors": [{
> "vendors": ["8086"],
> "devices": ["154c", "10ed"],
> "drivers": ["vfio-pci"]
> }, {
> "vendors": ["15b3"],
> "devices": ["1015"],
> "drivers": ["mlx5_core"]
> }
> Can we add this to this PR?

Lets discuss this use case in out weekly call. A demo/slide would be helpful.

ahalim-intel: > Also we have selector per netdevice which can be useful as well.

I thinking of creating maintaining a new branch until this is merged into master. So that we can include feature improvements, bug tracking/fix etc.

deployments/pod-sriovdp.yaml
moshe010: in my ubuntu is it under /usr/share/misc/pci.ids

ahalim-intel: Which release?
On 17.04 I can see there's a symlink
```
root@silv-ah-ubuntu1704:~# ll /usr/share/hwdata/
total 72
drwxr-xr-x   2 root root  4096 Mar 25 11:49 ./
drwxr-xr-x 306 root root 12288 Mar 25 11:49 ../
lrwxrwxrwx   1 root root    15 Jul  3  2016 pci.ids -> ../misc/pci.ids
-rw-r--r--   1 root root 55055 Jul  3  2016 pnp.ids
lrwxrwxrwx   1 root root    15 Jul  3  2016 usb.ids -> ../misc/usb.ids
```

moshe010: Ubuntu 18.04.2 LTS

deployments/pod-sriovdp.yaml
ahalim-intel: The device plugin daemonset needs to be run  in privilege mode because it needs to query host forwarding table. We also need to enable this should we add features to watch Link status. 

Have you managed to run this with `privileged: true`?

ahalim-intel:
Merging this PR into a feature/device_selector branch and continue the development in there. Our target is to merge this feature into a new release.

PR #103, opened 2019-03-19
remove vendor dependency for bulid/test/lint rules in Makefile

zshi-redhat:
Normally when building the binary and running ``make``, it's expected to build from local vendor dependencies instead of having to retrieve dependency update remotely. With this change, it allows user to build locally without having a internet connection. For glide update, we could still use ``make vendor``.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ $(GOBIN):
 $(BUILDDIR): | $(BASE) ; $(info Creating build directory...)
 	@cd $(BASE) && mkdir -p $@
 
-build: vendor $(BUILDDIR)/$(BINARY_NAME) ; $(info Building $(BINARY_NAME)...)
+build: $(BUILDDIR)/$(BINARY_NAME) | vendor ; $(info Building $(BINARY_NAME)...)
 	$(info Done!)
 
 $(BUILDDIR)/$(BINARY_NAME): $(GOFILES) | $(BUILDDIR)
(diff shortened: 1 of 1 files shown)


ahalim-intel:
@zshi-redhat One little concern with this.
For example, say, someone updated their vendor with newer packages and trigger vendor update with `make vendor`. The subsequent `make` won't build the binary unless `make clean` is run because vendor is not a prerequisites. But I see your point there, build should be run with whatever in the vendor/ and it should not require to fetch them during build. But at minimum we should have a check that `vendor/` exist before any build or else build may fail.
We could leave the `vendor` as _order-only-prerequisites_ as described [here](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html). Wdyt?

zshi-redhat:
> @zshi-redhat One little concern with this.
> For example, say, someone updated their vendor with newer packages and trigger vendor update with `make vendor`. The subsequent `make` won't build the binary unless `make clean` is run because vendor is not a prerequisites.

Yes, you're right, I didn't think that far, we'll have to first ``clean`` the environment and ``make`` after vendor packages be updated. do you think we can add another rule that is specific to vendor file update and replace current vendor rule with the new one in build/lint/test dependencies?

But I see your point there, build should be run with whatever in the vendor/ and it should not require to fetch them during build. But at minimum we should have a check that `vendor/` exist before any build or else build may fail.
> We could leave the `vendor` as _order-only-prerequisites_ as described [here](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html). Wdyt?

+1, this is better than just removing the vendor dependencies.

zshi-redhat:
> > @zshi-redhat One little concern with this.
> > For example, say, someone updated their vendor with newer packages and trigger vendor update with `make vendor`. The subsequent `make` won't build the binary unless `make clean` is run because vendor is not a prerequisites.
> 
> Yes, you're right, I didn't think that far, we'll have to first `clean` the environment and `make` after vendor packages be updated. do you think we can add another rule that is specific to vendor file update and replace current vendor rule with the new one in build/lint/test dependencies?

Considering the number of files under ``vendor`` is large, detecting the change of each file is time-consuming when just running ``make``;  I assume it's better not have rule for vendor file update.

> 
> But I see your point there, build should be run with whatever in the vendor/ and it should not require to fetch them during build. But at minimum we should have a check that `vendor/` exist before any build or else build may fail.
> 
> > We could leave the `vendor` as _order-only-prerequisites_ as described [here](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html). Wdyt?
> 
> +1, this is better than just removing the vendor dependencies.

zshi-redhat:
@ahalim-intel Just realized that with ``vendor`` in the order-only-prerequisites, it will still be evaluated and corresponding commands will be executed. perhaps we want to just remove the ``vendor`` from rules.

PR #106, opened 2019-03-27
Add filtering By netDev Names

Mmduh-483:
Filtering the resources by the network device name


pkg/types/types.go
```diff
@@ -32,6 +32,7 @@ type ResourceConfig struct {
 		Vendors []string `json:"vendors,omitempty"`
 		Devices []string `json:"devices,omitempty"`
 		Drivers []string `json:"drivers,omitempty"`
+		NetDevs []string `json:"netDevs,omitempty"`
```
ahalim-intel: This is probably a minor thing, but "netDevs" is not intuitive from user perspective in my opinion. How about "ifnames" as in short for "Interface name"? Also it is worth updating the README file under selector section.

zshi-redhat: Or does the ``NetDevs`` intentionally applied to net device only, not includes user space device like uio or vfio?

pkg/resources/deviceSelectors.go
```diff
@@ -64,6 +65,28 @@ func (s *driverSelector) Filter(inDevices []types.PciNetDevice) []types.PciNetDe
...
+}
+
+type netDevSelector struct {
+	netDevs []string
+}
+
+func (s *netDevSelector) Filter(inDevices []types.PciNetDevice) []types.PciNetDevice {
+	filteredList := make([]types.PciNetDevice, 0)
+	for _, dev := range inDevices {
+		if netDevs, err := utils.GetNetNames(dev.GetPciAddr()); err == nil {
```
zshi-redhat: Is it worth adding the ability to filter user space device name here, for example: dev/vfio/<num>?

pkg/types/types.go
moshe010: yes

moshe010:
> The interface name of kernel network device may be changed after it's released by SR-IOV cni, in which case the config file with `NetDev` filter may not be applied as expected. This could happen when VF is used once by Pod && device plugin gets restarted.
Yes you are right. This will require changes in SR-IOV cni like caching the interface name

pkg/types/types.go
zshi-redhat: Perhaps using ``NetDevNames``If it's for kernel network device only?

pkg/resources/deviceSelectors.go
zshi-redhat: For Mellanox card, this may not be necessary as there is no pre binding of devices to vfio or uio driver. @ahalim-intel do you think we need a user space device name filter for other NICs?

ahalim-intel:
> > The interface name of kernel network device may be changed after it's released by SR-IOV cni, in which case the config file with `NetDev` filter may not be applied as expected. This could happen when VF is used once by Pod && device plugin gets restarted.
> > Yes you are right. This will require changes in SR-IOV cni like caching the interface name

We are doing this in this PR: https://github.com/intel/sriov-cni/pull/71
SRIOV-CNI should reset the interface name to original name.

pkg/resources/deviceSelectors.go
ahalim-intel: With userspace driver there won't be any kernel interface name available.

zshi-redhat: Yes, there won't be kernel interface name for userspace driver, what about the device spec name ``/dev/vfio/<num>``? is it worth filtering? 
Another problem(also discussed in #109 ) is that the ``netDevs`` filter may contain a node specific interface name that may not applied to other  nodes in a heterogeneous environment.

Mmduh-483:
> what will happen if the PF netdevice is passed in inDevices ?
> is there such a scenario ? or are PF net devices always excluded from inDevices.


> GetPfName

Mmduh-483:
> what will happen if the PF netdevice is passed in inDevices ?
> is there such a scenario ? or are PF net devices always excluded from inDevices.

if the device is not a VF then GetPfName will return the net dev name for the device itself (it is in the comment above the function declaration)

Mmduh-483:
> GetPfNetName ?
> 
> Also, is there a reason why this is not implemented by calling: GetNetNames(GetPfName(pciAddr))[0] ?

There is no function with this name `GetPfNetName`

Mmduh-483:
> GetPfNetName ?
> 
> Also, is there a reason why this is not implemented by calling: GetNetNames(GetPfName(pciAddr))[0] ?

done

Mmduh-483:
> enp0s0f2 interface name is usually assigned to a VF net device right ?
> 
> the filter should be according to PF netdev, lets give a different example in the readme
> 
> e.g enp1s0f0, enp2s2f0

done

zshi-redhat:
please see comments [here](https://github.com/intel/sriov-network-device-plugin/pull/110#issuecomment-483497087) and [here](https://github.com/intel/sriov-network-device-plugin/pull/110#issuecomment-483497867)

Mmduh-483:
moved to https://github.com/intel/sriov-network-device-plugin/pull/122

PR #110, opened 2019-04-01
Add support for mounting rdma resources

Mmduh-483:
If isRdma flag is true and the device is supporting the rdma then
the rdma dir '/dev/infiniband' will be mounted to the container


README.md
```diff
@@ -153,6 +154,7 @@ This plugin creates device plugin endpoints based on the configurations given in
         },
         {
             "resourceName": "intel_sriov_dpdk",
+            "isRdma": true,
```
zshi-redhat: Shall we have a separate resourceName example here for Mellanox rdma devices?
The current one seems confusing as Intel NIC doesn't support rdma, right?

moshe010: absolutely this example is misleading we should have on for intel and one for mellanox

ahalim-intel:
@Mmduh-483 
Same comment as in #111

Mmduh-483:
> What will happen if additional operations are required other than mounts ?
> 
> Should we have in addition, something like : PrepareRdma(PciDevice) in RdmaSpec or in RdmaManager ?

PrepareRdma it seems like you want to write configuration but the device plugin is responsible to discover not the configure new things, and AFAIK there is no case mellanox cards to prepare rdma.
@ahalim-intel @zshi-redhat what do you think guys?

Mmduh-483:
> also i think we need a unit test for the RDMA spec infra and this module

I will add that after it is ok for the approach of rdmaManager

Mmduh-483:
> if the only role of rdmaManager is to return an RdmaSpec object
> this can be simply: func RdmaSpecFactory(vendor, pciAddrs string) types.RdmaSpec
> 
> i feel this is over-abstracting abit, unless there is a valid use-case.

it is the same as the factory Interface

Mmduh-483:
@zshi-redhat @ahalim-intel I refactored the code, please review it, I will add the unit tests later

pkg/resources/factory.go
```diff
@@ -119,3 +120,14 @@ func (rf *resourceFactory) GetResourcePool(rc *types.ResourceConfig, deviceList
 	rPool := newResourcePool(rc, apiDevices, devicePool)
 	return rPool, nil
 }
+
+// GetRdmaSpec returns the vendors rdma specs
+func (rf *resourceFactory) GetRdmaSpec(vendor, pciAddrs string) types.RdmaSpec {
+	switch vendor {
+	// Mellanox device
+	case "15b3":
```
zshi-redhat: Does this have to be vendor specific creation or there can be a general way to discovering rdma capability of a network pci device as a kernel device?

pkg/resources/rdma/genericRdma.go
```diff
@@ -0,0 +1,26 @@
+package rdma
+
+import (
+	pluginapi "k8s.io/kubernetes/pkg/kubelet/apis/deviceplugin/v1beta1"
+
+	"github.com/intel/sriov-network-device-plugin/pkg/types"
+)
+
+type genericRdmaSpec struct {
```
zshi-redhat: Is it possible to detect the rdma mode for a particular device and add a field in this RdmaSpec? for example, whether it's in RoCE, IB or iWARP mode.

zshi-redhat:
I still can see the rebase conflict notification with current handle zshi-redhat.. if I switch to another account, then it shows no conflict.

zshi-redhat:
I see there are 7 commits included in this PR where two of them are for the rdma feature.
would you please do a rebase locally with current head of feature/device_selectors and only apply the 2 rdma related commits above, then force push it again? if this still doesn't work, perhaps we need to create another PR for it.

pkg/resources/rdma/mlnxRdma.go
```diff
@@ -0,0 +1,36 @@
...
+type mlnxRdmaSpec struct {
+	isRdma bool
+	mounts []*pluginapi.Mount
+}
+
+// NewMlnxRdmaSpec returns rdmaSpec for Mellanox NetDev
+func NewMlnxRdmaSpec(pciAddrs string) types.RdmaSpec {
+	isRdma := len(rdmamap.GetRdmaDevicesForPcidev(pciAddrs)) > 0
+	mounts := make([]*pluginapi.Mount, 0)
+	mounts = append(mounts, &pluginapi.Mount{
```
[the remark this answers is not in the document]
Mmduh-483: There is no specific case, is there any advantages of using DeviceSpec?

pkg/resources/factory.go
Mmduh-483: Vendors will mostly have only one approach to support rdma and have there own resources paths to mount

pkg/resources/pool_stub.go
```diff
@@ -102,6 +102,10 @@ func (rp *resourcePool) GetMounts(deviceIDs []string) []*pluginapi.Mount {
 	for _, id := range deviceIDs {
 		if dev, ok := rp.devicePool[id]; ok {
 			mnt := dev.GetMounts()
+			if rp.config.IsRdma && dev.IsRdma() {
```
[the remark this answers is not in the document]
Mmduh-483: I see in `netDevicePool` that there are no mounts being returned, is there a reason for that?

Mmduh-483:
> I see there are 7 commits included in this PR where two of them are for the rdma feature.
> would you please do a rebase locally with current head of feature/device_selectors and only apply the 2 rdma related commits above, then force push it again? if this still doesn't work, perhaps we need to create another PR for it.
> btw, there are two rebases (two PRs are rebased and added) in feature/device_selectors branch after this PR was created, but they don't appear in the commits of this PR.

I rebased the code, can you please recheck?

pkg/resources/pool_stub.go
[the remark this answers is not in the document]
zshi-redhat: There is currently no need to mount a certain filesystem path for network device in order for it to work; rdma is the first case that we need to mount for a network device.

pkg/resources/rdma/mlnxRdma.go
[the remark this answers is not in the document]
zshi-redhat: ``pluginapi.[DeviceSpec](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/deviceplugin/v1beta1/api.proto#L150)`` is for mounting device to container, it has different permission settings than ``pluginapi.[Mount](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/deviceplugin/v1beta1/api.proto#L139)`` which is for mounting host volume to container.

pkg/resources/factory.go
zshi-redhat: I was thinking if we can use a generic filesystem path to discover whether a kernel device is capable of doing rdma, for example, read a certain path under ``/sys/bus/pci/<id>`` to determine if that pci device supports rdma. Is this part vendor specific?

zshi-redhat:
> > I see there are 7 commits included in this PR where two of them are for the rdma feature.
> > would you please do a rebase locally with current head of feature/device_selectors and only apply the 2 rdma related commits above, then force push it again? if this still doesn't work, perhaps we need to create another PR for it.
> > btw, there are two rebases (two PRs are rebased and added) in feature/device_selectors branch after this PR was created, but they don't appear in the commits of this PR.
> 
> I rebased the code, can you please recheck?

This still has conflict, I can see the conflict when running the following instructions:
```
git clone https://github.com/Mmduh-483/sriov-network-device-plugin.git 
cd sriov-network-device-plugin/ 
git checkout device_selector/rdma-support 
git remote add upstream https://github.com/intel/sriov-network-device-plugin.git 
git fetch upstream 
git pull --rebase upstream feature/device_selectors 
#here, you can see the conflicts....
```

Mmduh-483:
move to https://github.com/intel/sriov-network-device-plugin/pull/120

PR #120, opened 2019-04-21
Add support for mounting rdma resources

Mmduh-483:
If isRdma flag is true and the device is supporting the rdma then
the rdma device Spec will be mounted to the container


pkg/resources/rdma/mlnxRdma.go
```diff
@@ -0,0 +1,36 @@
...
+type mlnxRdmaSpec struct {
+	isRdma     bool
+	deviceSpec []*pluginapi.DeviceSpec
+}
+
+// NewMlnxRdmaSpec returns the Mellanox vendor RdmaSpec
+func NewMlnxRdmaSpec(pciAddrs string) types.RdmaSpec {
+	isRdma := len(rdmamap.GetRdmaDevicesForPcidev(pciAddrs)) > 0
+	deviceSpec := make([]*pluginapi.DeviceSpec, 0)
+	deviceSpec = append(deviceSpec, &pluginapi.DeviceSpec{
```
adrianchiris: if isRdma == False i dont think we need to append a DeviceSpec object

pkg/resources/pool_stub.go
```diff
@@ -68,6 +68,11 @@ func (rp *resourcePool) GetDeviceSpecs(deviceIDs []string) []*pluginapi.DeviceSp
 	for _, id := range deviceIDs {
 		if dev, ok := rp.devicePool[id]; ok {
 			newSpecs := dev.GetDeviceSpecs()
+			rdmaSpec := dev.GetRdmaSpec()
+			if rp.config.IsRdma && rdmaSpec.IsRdma() {
```
adrianchiris: what will happen if a resource is marked as isRdma in the config file but rdmaSpec.IsRdma() is false ?
I think we should at least  log the error as the container will have no RDMA capabilities although the user requested it.

pkg/resources/rdma/mlnxRdma.go
Mmduh-483: fixed

pkg/resources/pool_stub.go
Mmduh-483: I will add a new filter in the `deviceSelector,go` to fix it

adrianchiris: Also i recommend to log an error if we get to this point and (rp.config.IsRdma != rdmaSpec.IsRdma())

Mmduh-483: ok

pkg/resources/factory.go
```diff
@@ -103,6 +104,16 @@ func (rf *resourceFactory) GetResourcePool(rc *types.ResourceConfig, deviceList
 		}
 	}
 
+	// filter for rdma devices
+	if rc.IsRdma {
+		rdmaDevices := make([]types.PciNetDevice, 0)
+		for _, dev := range filteredDevice {
+			if dev.GetRdmaSpec().IsRdma() {
+				rdmaDevices = append(rdmaDevices, dev)
```
zshi-redhat: Looks like ``rdmaDevices`` is created but not used anywhere, shall it be assigned back to ``filteredDevice``?

images/configMap.yaml
```diff
@@ -21,6 +21,15 @@ data:
                     "devices": ["154c", "10ed"],
                     "drivers": ["vfio-pci"]
                 }
+            },
+            {
+                "resourceName": "mlnx_sriov_rdma",
+                "isRdma": true,
```
zshi-redhat: Is there any thought on how to add various ``rdma`` modes support based on the ``isRdma`` config?  like how do you specify the rdma mode(RoCE, IB, etc) in the config.

adrianchiris: How RDMA is implemented and transported depends on the fabric and the driver.
Basically you can say, from an application point of view(assuming it was written correctly and does no assumptions on the underlying fabric), that its "auto-selected".

pkg/resources/factory.go
Mmduh-483: fixed

ahalim-intel:
What happens if a resource pool configured as `"isRdma": true` but the resource pool contains devices that do not support RDMA(insufficient selectors)? Are you ensuring that only RDMA capable devices should be included for a resource pool with `"isRdma": true`?

images/configMap.yaml
Mmduh-483: > What happens if a resource pool configured as `"isRdma": true` but the resource pool contains devices that do not support RDMA(insufficient selectors)? Are you ensuring that only RDMA capable devices should be included for a resource pool with `"isRdma": true`?

This approach only includes the RDMA capable devices, we may add an additional configuration's flag `"mountRdma": bool` this will be to mount RDMA resources if the device is RDMA capable, or we can mount the RDMA resource if the device is RDMA capable by default and remove `isRdma` flag from the configuration file, what do you think is the best approach?

Mmduh-483:
This approach only includes the RDMA capable devices, we may add an additional configuration's flag "mountRdma": bool this will be to mount RDMA resources if the device is RDMA capable, or we can mount the RDMA resource if the device is RDMA capable by default and remove isRdma flag from the configuration file, what do you think is the best approach?

ahalim-intel:
As long as a resource pool does not include non-rdma device then this is Ok.  I don't think we need another flag for that. But I think what is important there is to be able to specify the mode of RDMA. In my understanding there are 3 modes(or 4 maybe), Infiniband, iWARP, RoCE, and RoCEv2.  So, in the factory.go GetRdmaSpec() we should switch on RDMA modes instead of vendor ID there(some vendor/nic may support more than one modes).

Maybe we should replace the flag "isRdma": bool with "rdmaMode": string. Once this flag is set then we can assume RDMA mode is desired and the type of mode is specified in its value("infiniband", "iwarp", "roce", "rocev2" etc.). The GetRdmaSpec() method should get right type of RDMA spec provider. And during the resource pool creation it should be taken care of that only devices with supported RDMA added into the resourc pool.

images/configMap.yaml
ahalim-intel: > How RDMA is implemented and transported depends on the fabric and the driver.
> Basically you can say, from an application point of view(assuming it was written correctly and does no assumptions on the underlying fabric), that its "auto-selected".

That is correct, but other modes of rdma may need different device specs/mounts to be exported. And we may not want to mix different types of RDMA capbale devices into same resource pool either(for different underlying fabrics. performance, latency, CPU utilization etc).

paravmellanox:
@ahalim-intel Link layer details (such as roce, rocev2, iwarp, IB) should not be exposed to the device plugin. Kernel treats them using unified stack and user apis.
rdma devices can be single or dual port.
In fact most mellanox devices supports roce, rocev2 on single port, second port can be IB or both can be IB or RoCE.
rdmamap library https://github.com/Mellanox/rdmamap also is link layer agnostic as it should be just like kernel stack and other user space packages.

So it should be just a bool flag.

ahalim-intel:
@paravmellanox If this information is not exposed for the user, how they will know what underlying mode the RDMA is operating on? This maybe important for many reason. I know that from application perspective it doesn't matter. 

I don't have enough info yet, but for making a VF available in a container in other modes may need some other specific files in the container env. Just like for infiniband the "/dev/infiniband" is being exported. With this implementation its being decided by looking at the vendor id("15b3"). This is problematic because same vendor/NIC may  support different modes. How do you handle other modes other than infiniband? Or you are going to assume "isRdma": true and vendor id is "15b3" so just export "/dev/infiniband" for any modes? My question is how we are going handle mode specific device specs  in the device plugin?

paravmellanox:
@ahalim-intel from kernel side, regardless of vendor type kernel ABI remains the same.
for example if vendor != 15b3, than instead of /dev/infiniband/ will we have /dev/iwarp? 
Answer is no. rdma subsystem is not build that way. We have clear definitions of how devices are exposed regardless of their underlying link layer or vendor.
Currently we have several rdma drivers from broadcom, mellanox, cavium, intel. All exposes same type of devices to users and same user interface.

for an unknown reason in future, if a need arise to do very vendor specific things, a plugin should execute vendor specific plugin calls. That framework for extending device plugin should be build at that point of time.

paravmellanox:
@ahalim-intel device plugin are expected to configure rdma devices correctly regardless of their link layer type. 

User may give additional explicit hint the container yaml file that they want rdma device of IB type or not.
So for example if cluster has 3 IB PFs and 4 iWarp PFs, all treated equally as rdma and things will fail.
However, lets say even if we don't have 4 iWarp PFs, who ensures that those 3 IB PFs are inter connected at fabric level? This is the fabric/network configuration problem.
So future extension can give additional hints where user says resource_type=rdma, rdma_protocol=ib/roce/iwarp. But it shouldn't be the explicit or mandatory resource definition.
And there may be more sub features (say rdma tag matching, or gpudirect enabled) can be requested to pick right node (PF) to choose.

README.md
```diff
@@ -158,6 +158,15 @@ This plugin creates device plugin endpoints based on the configurations given in
                 "devices": ["154c", "10ed"],
                 "drivers": ["vfio-pci"]
             }
+        },
+        {
+            "resourceName": "mlnx_sriov_rdma",
```
paravmellanox: it shouldn't say mlnx here.
vendor and device id describe what vendor it is.

resource name should be just 'rdmaDevice'.

README.md
```diff
@@ -158,6 +158,15 @@ This plugin creates device plugin endpoints based on the configurations given in
...
+            "isRdma": true,
+            "selectors": {
+                "vendors": ["15b3"],
+                "devices": ["1018"],
+                "drivers": ["mlx5_core"]
```
paravmellanox: @Mmduh-483 please remove driver from here.
Only vendor and devide should be mentioned.

rdma device are not created by mlx5_core driver either. they are created by mlx5_ib driver and pci device is owned by mlx5_core driver.

pkg/resources/pciNetDevice.go
```diff
@@ -42,6 +43,7 @@ func NewPciNetDevice(pciDevice *ghw.PCIDevice, rFactory types.ResourceFactory) (
 	dSpecs := infoProvider.GetDeviceSpecs(pciAddr)
 	mnt := infoProvider.GetMounts(pciAddr)
 	env := infoProvider.GetEnvVal(pciAddr)
+	rdmaSpec := rFactory.GetRdmaSpec(pciDevice.Vendor.ID, pciDevice.Address)
```
paravmellanox: @Mmduh-483 as this is done correctly, you didn't use the driver parameter, hence driver should not be specified as I explained in previous comment.

pkg/resources/rdma/mlnxRdma.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"github.com/Mellanox/rdmamap"
+	pluginapi "k8s.io/kubernetes/pkg/kubelet/apis/deviceplugin/v1beta1"
+
+	"github.com/intel/sriov-network-device-plugin/pkg/types"
+)
+
+// mlnxRdmaDir path to rdma resources
+const mlnxRdmaDir = "/dev/infiniband"
```
paravmellanox: remove mlnx prefix from here and this const completely.

Add RdmaDeviceDir = "/dev/infiniband" to https://github.com/Mellanox/rdmamap/blob/master/rdma_map.go and use is here as rdmamap.RdmaDeviceDir

pkg/resources/rdma/mlnxRdma.go
```diff
@@ -0,0 +1,38 @@
...
+	return &mlnxRdmaSpec{isRdma: isRdma, deviceSpec: deviceSpec}
+}
+
+func (r *mlnxRdmaSpec) IsRdma() bool {
+	return r.isRdma
+}
+
+func (r *mlnxRdmaSpec) GetRdmaDeviceSpec() []*pluginapi.DeviceSpec {
+	return r.deviceSpec
+}
```
paravmellanox: @Mmduh-483 there is nothing mellanox vendor specific done here, rename it to rdma/rdma.go

pkg/resources/factory.go
```diff
@@ -119,3 +131,13 @@ func (rf *resourceFactory) GetResourcePool(rc *types.ResourceConfig, deviceList
 	rPool := newResourcePool(rc, apiDevices, devicePool)
 	return rPool, nil
 }
+
+func (rf *resourceFactory) GetRdmaSpec(vendor, pciAddrs string) types.RdmaSpec {
+	switch vendor {
+	// Mellanox device
+	case "15b3":
```
paravmellanox: this case should not be there either. it should be just based on the pciAddr

pkg/resources/rdma/mlnxRdma.go
paravmellanox: @Mmduh-483 addition of RdmaDeviceDir is taken care in the https://github.com/Mellanox/rdmamap/blob/master/rdma_map.go 

Please update and use that directly.

ahalim-intel:
> @ahalim-intel from kernel side, regardless of vendor type kernel ABI remains the same.
> for example if vendor != 15b3, than instead of /dev/infiniband/ will we have /dev/iwarp?

This is exactly what my point is. This decision should not be made for a vendor ID if the /dev/infiniband/ is going to be common(?) for any rdma modes. 

> So for example if cluster has 3 IB PFs and 4 iWarp PFs, all treated equally as rdma and things will fail.

So it makes more sense to have different resource pools for different underlying protocols.

> However, lets say even if we don't have 4 iWarp PFs, who ensures that those 3 IB PFs are inter connected at fabric level? This is the fabric/network configuration problem.

We are not concerned about connectivity related problems here. But at minimal, we could describe in a resource pool what type of devices it's representing. I'm sure the K8s resource class would provide a lot more information about hardware resources a device plugin exposes.

paravmellanox:
> 
> 
> > @ahalim-intel from kernel side, regardless of vendor type kernel ABI remains the same.
> > for example if vendor != 15b3, than instead of /dev/infiniband/ will we have /dev/iwarp?
> 
> This is exactly what my point is. This decision should not be made for a vendor ID if the /dev/infiniband/ is going to be common(?) for any rdma modes.
> 
@ahalim-intel, right.
I think /dev/infiniband containing 'infiniband' created the confusion.
But that is just legacy name existed. We can ignore the name part, it does represents ib,roce/iwarp.

> > So for example if cluster has 3 IB PFs and 4 iWarp PFs, all treated equally as rdma and things will fail.
> 
> So it makes more sense to have different resource pools for different underlying protocols.
> 
To begin with there should be abstract rdma resource that 'user' can annotate in the container yaml file as resource he/she wants to request.
(user != administrator)
So selector should be 'mlnx_rdma" or I prefer 'rdma', but fine for now there is vendor name prefixed.
In future, for a given resource there can be more properties such as, 
linklayer=ib/roce/iwarp
gpudirect=support/dontcare
tagmaching=supported/dontcare
more...

Putting all the combinations as individual resource name "mlnx_rdma_ib" or "mlnx_rdma_roce" "mlnx_gpudirect_ib", mlnx_gpudirect_roce", doesn't extend well. There are many combinations that can be created...
A given k8s resource description should have one or more such fields and user should be able to request/give hint of what is desired.
So to begin with 'mlxn_rdma' is good start.

> > However, lets say even if we don't have 4 iWarp PFs, who ensures that those 3 IB PFs are inter connected at fabric level? This is the fabric/network configuration problem.
> 
> We are not concerned about connectivity related problems here. But at minimal, we could describe in a resource pool what type of devices it's representing. I'm sure the K8s resource class would provide a lot more information about hardware resources a device plugin exposes.
device type or link layer type should be a sub field.

So my input summary is:
1. drop mlnxrdma.go file, do it generic way as rdma.go that is what kernel also does.
2. At present start with /dev/infiniband defined in https://github.com/Mellanox/rdmamap/blob/master/rdma_map.go#33
(a) later on Make use of API rdmamap.GetRdmaDevicesForPcidev()
https://github.com/Mellanox/rdmamap/blob/master/rdma_map.go#344
3. resource definition as 'mlnx_rdma"
4. driver string to be updated as ["mlx5_ib", "mlx5_core"]

In future when any vendor or link layer specific to be done, rdmamap should be updated to provide different devices etc by making right kernel calls using rdma netlink sockets and/or sysfs file layout which is well defined.

README.md
Mmduh-483: @paravmellanox this is an example for Mellanox's RDMA devices

README.md
Mmduh-483: same as above

pkg/resources/rdma/mlnxRdma.go
Mmduh-483: updated

pkg/resources/rdma/mlnxRdma.go
Mmduh-483: rdma/<vendor>Rdma.go is for vendor-specific RDMA if the vendor wants to initialize, configure, ...etc. anything for RDMA it will be here, @moshe010 @ahalim-intel is it good to add another function here like 
`initRdmaDeviceSpec` or `setupRdmaDeviceSpec`

pkg/resources/factory.go
Mmduh-483: @paravmellanox how is the device plugin going to get the rdmaDeviceSpec from the PCI address?

pkg/resources/rdma/mlnxRdma.go
rajatchopra: Any init/setup can be done within the `New` itself. Is there a need that init/setup be called from outside this file?

pkg/resources/factory.go
```diff
@@ -119,3 +130,7 @@ func (rf *resourceFactory) GetResourcePool(rc *types.ResourceConfig, deviceList
 	rPool := newResourcePool(rc, apiDevices, devicePool)
 	return rPool, nil
 }
+
+func (rf *resourceFactory) GetRdmaSpec(vendor, pciAddrs string) types.RdmaSpec {
```
moshe010: @Mmduh-483  remove the vendor you are not using it anyway

ahalim-intel: Ok, lets remove the "vendor" from this method's parameter for now. Once we get it verified by @zshi-redhat again that it works with all the new changes we can go ahead and merge it.
But please keep in mind we may need changes in future to facilitate other modes of RDMA.

Mmduh-483: done

zshi-redhat:
Run some tests with ib_send_bw and verified the base functionality of RDMA:   RDMA connection can be established from SR-IOV VF device within Pod to remote PF device on another host.  Paste the test performance below:
```
=== server side, remote PF ===
[root@nfvpe-07 ]# ib_send_bw -d mlx5_2 -i 1 -x 0

************************************
* Waiting for client to connect... *
************************************
---------------------------------------------------------------------------------------
                    Send BW Test
 Dual-port       : OFF		Device         : mlx5_2
 Number of qps   : 1		Transport type : IB
 Connection type : RC		Using SRQ      : OFF
 RX depth        : 512
 CQ Moderation   : 100
 Mtu             : 1024[B]
 Link type       : Ethernet
 GID index       : 0
 Max inline data : 0[B]
 rdma_cm QPs	 : OFF
 Data ex. method : Ethernet
---------------------------------------------------------------------------------------
 local address: LID 0000 QPN 0x01d2 PSN 0xf7273b
 GID: 254:128:00:00:00:00:00:00:154:03:155:255:254:151:56:222
 remote address: LID 0000 QPN 0x04cc PSN 0xf60b2
 GID: 254:128:00:00:00:00:00:00:192:119:11:255:254:95:29:181
---------------------------------------------------------------------------------------
 #bytes     #iterations    BW peak[MB/sec]    BW average[MB/sec]   MsgRate[Mpps]
Conflicting CPU frequency values detected: 999.932000 != 3040.649000. CPU Frequency is not max.
 65536      1000             0.00               4372.03		   0.069952
---------------------------------------------------------------------------------------

```

```
==== client side, SR-IOV VF inside pod ===

[root@testpod1 /]# ib_send_bw -d mlx5_4 -i 1 10.56.217.100  -x 0
---------------------------------------------------------------------------------------
                    Send BW Test
 Dual-port       : OFF		Device         : mlx5_4
 Number of qps   : 1		Transport type : IB
 Connection type : RC		Using SRQ      : OFF
 TX depth        : 128
 CQ Moderation   : 100
 Mtu             : 1024[B]
 Link type       : Ethernet
 GID index       : 0
 Max inline data : 0[B]
 rdma_cm QPs	 : OFF
 Data ex. method : Ethernet
---------------------------------------------------------------------------------------
 local address: LID 0000 QPN 0x04cc PSN 0xf60b2
 GID: 254:128:00:00:00:00:00:00:192:119:11:255:254:95:29:181
 remote address: LID 0000 QPN 0x01d2 PSN 0xf7273b
 GID: 254:128:00:00:00:00:00:00:154:03:155:255:254:151:56:222
---------------------------------------------------------------------------------------
 #bytes     #iterations    BW peak[MB/sec]    BW average[MB/sec]   MsgRate[Mpps]
Conflicting CPU frequency values detected: 999.932000 != 3525.946000. CPU Frequency is not max.
 65536      1000             4361.50            4361.46		   0.069783
---------------------------------------------------------------------------------------
```
I think we are good to go with this version of RDMA support.

README.md
```diff
@@ -169,9 +178,10 @@ This plugin creates device plugin endpoints based on the configurations given in
 
 
 |     Field      | Required |        Description        |                       Type - Accepted values                        |                          Example/Accepted values                                                   |
-|----------------|----------|---------------------------|---------------------------------------------------------------------|---------------------------------------------------------------------------|
+|----------------|----------|---------------------------|---------------------------------------------------------------------|---------------------------------------------------------------------------.........................|
```
ahalim-intel: This formatting breaks the table structure when you view the README.md.

I've updated the table with new contents below. Can you please fix this(copy & paste)?

```
|     Field      | Required |        Description        |                      Type - Accepted values                       |                                      Example/Accepted values                                       |
|----------------|----------|---------------------------|-------------------------------------------------------------------|----------------------------------------------------------------------------------------------------|
| "resourceName" | Yes      | Endpoint resource name    | string - must be unique and should not contain special characters | "sriov_net_A"                                                                                      |
| "selectors"    | No       | A map of device selectors | Each selector is a map of string list.                            | "vendors": ["8086"], "devices": ["154c", "10ed"], "drivers": ["vfio-pci"], "pfNames": ["enp2s2f0"] |
| "isRdma"       | No       | Mount RDMA resources      | `bool` - boolean value true or false                              | "isRdma": true                                                                                     |
```

PR #121, opened 2019-04-21
Add multi selector

Mmduh-483:
Provides the end user to create more filtering options


README.md
```diff
@@ -145,20 +145,27 @@ This plugin creates device plugin endpoints based on the configurations given in
...
+            "selectors": [
+                {
+                    "vendors": ["8086"],
+                    "devices": ["154c", "10ed"],
+                    "drivers": ["vfio-pci"]
+                },
+                {
+                    "vendors": ["15b3"],
+                    "devices": ["1014"],
+                    "drivers": ["mlx5_core"]
```
zshi-redhat: nit. Although this won't cause any issue with above configuration (using ``mlx5_core`` under ``intel_sriov_dpdk`` resourcename), it might be worth changing it to intel kernel driver like ``i40evf``;  and we could add ``mlx5_core`` driver in the first example with mlx_sriov as the resource name:
```
{
    "resourceList": [{
	            "resourceName": "mlx_sriov",
	            "selectors": [{
	                "vendors": ["15b3"],
	                "devices": ["1014"],
	                "drivers": ["mlx5_core"]
	            }]
	        },
                {
	            "resourceName": "intel_sriov",
                    "selectors": [
	                {
	                    "vendors": ["8086"],
	                    "devices": ["154c", "10ed"],
	                    "drivers": ["vfio-pci"]
	                },
	                {
	                    "vendors": ["8086"],
	                    "devices": ["154c"],
	                    "drivers": ["i40evf"]
                       }
	            ]
                }
	    ]
}
...

```

Mmduh-483: fixed

README.md
```diff
@@ -144,21 +144,28 @@ This plugin creates device plugin endpoints based on the configurations given in
...
+                    "devices": ["154c", "10ed"],
+                    "drivers": ["vfio-pci"]
+                },
+                {
+                    "vendors": ["8086"],
+                    "devices": ["154c"],
+                    "drivers": ["i40evf"]
+                }
+            ]
+       }
```
ahalim-intel: This approach will process same filters multiple times. For example, the "driver" filter already support multiple driver names can can be processed in one go. And by introducing selector lists we are essentially filtering them all over again. A lot of redundant processing. As we add new filters this can be expensive and complicated in future. Since the intent is to allow multiple vendor/device combination in a resource pool what about we introduce something like a vendorDevice struct instead?

Let me show you an example below:

```
{
    "resourceName": "intel_sriov",
    "selectors": [
        {
            "vendorDevice": [
            	{"vendorID": "8086", "deviceID": "154c"},
            	{"vendorID": "8086", "deviceID": "10ed" },
            	{"vendorID": "15b3", "deviceID": "1014" },
            ],
            "drivers": ["i40evf", "ixgbevf", mlx5_core" ]
        },
    ]
}
```
The above configuration will replace "vendor" and "device" selectors introduce a new struct combining them. With this approach the deviceID could be omitted. If deviceID is not present in a vendorDevice entry then this selector will match devices with only vendorID.

```
type VendorDevice struct {
	VendorID string `json:"vendorID"`
	DeviceID string `json:"deviceID,omitempty"`
}
```
Let me know what do you think about the above approach instead. Will that cover your use-cases?

Mmduh-483: I agree with you to exclude the drivers out of the multi selector, but is it guaranteed that the drivers' names' will never be similar between 2 vendors? or even 2 devices? we may need to use a specific driver with some devices id because others may be incompatible or having issues like:

vendor: A
device: d1, d2
drivers: dr1, dr2
and after some time the support for driver dr2 got deprecated for d1 or even have an issue or a bug, this case is not covered in your approach, another approach may be for testing specific drivers compatibility. 

In case we have one vendor with many devices, let's say 5 devices, the in the configuration's file is something like:
```
{
    "resourceName": "sriov",
    "selectors": {
            "vendorDevice": [
            	{"vendorID": "1234", "deviceID": "1"},
            	{"vendorID": "1234", "deviceID": "2" },
            	{"vendorID": "1234", "deviceID": "3" },
                {"vendorID": "1234", "deviceID": "4"},
            	{"vendorID": "1234", "deviceID": "5" }
            ],
            "drivers": ["a", "b", "c" ]
        }
}
```
this is not something the user may like.

Also, you are overlapping the filters for the vendors and the devices, to check if the device is in the white list, you need to check for the vendor and the device's id in the same filter, so you are doing a full loop for the vendors and devices in the white list for each device and this is the same complexity as now.

Mmduh-483:
@ahalim-intel is there any comments for this PR?

README.md
ahalim-intel: @Mmduh-483 
The "drivers" is just an example, any other filter we may introduce in future will face similar consequences. The "vendorDevice" is not ideal solution either - it's just a quick thought on handling multiple vendor+device combos in same resource pool.

> vendor: A
device: d1, d2
drivers: dr1, dr2
and after some time the support for driver dr2 got deprecated for d1 or even have an issue or a bug, this case is not covered in your approach, another approach may be for testing specific drivers compatibility.

If a driver is not supported or does not exist that should not cause any problem. We need to remember the order of device filter evaluation important here. Each filter will narrow down the device list for the next. Any device that do not match filter simply wont be included for that pool.

IMHO, current implementation can aleady handle the multi vendor+device without much issues. We can just simply add correct vendor ids, device ids that we wish to add into a pool. Only concern of a device ids overlapping with other vendors maybe an issue. But so far I haven't seen any common device id's for supported SR-IOV NIC's yet. If this becomes an issue then it's better to handle with a tuple like struct as shown with "vendorDevice" with additional repetition of same vendors for different devices.  But this will limit the repetition for vendor and devices only, not for other filters. 

Maybe I am not getting the true intention of the PR. Are we just trying to facilitate different vendor+devices in a pool? OR to be able to create a pool with devices that meets any of the listed "selector" group? i.e. "selector group 1" OR "selector group 2" OR "selector group 3" Or ... so on and evaluates all selectors that in the group?

PR #122, opened 2019-04-21
Add filtering By PF Names

Mmduh-483:
Filtering the resources by the PF device name

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,8 @@ This plugin creates device plugin endpoints based on the configurations given in
             "selectors": {
                 "vendors": ["8086"],
                 "devices": ["154c", "10ed"],
-                "drivers": ["vfio-pci"]
+                "drivers": ["vfio-pci"],
+                "pfNames": ["enp0s0f0","enp2s2f1"]
             }
         }
     ]
@@ -167,10 +168,10 @@ This plugin creates device plugin endpoints based on the configurations given in
 
 
 
-|     Field      | Required |        Description        |                       Type - Accepted values                        |                          Example/Accepted values                          |
+|     Field      | Required |        Description        |                       Type - Accepted values                        |                          Example/Accepted values                                                   |
 |----------------|----------|---------------------------|---------------------------------------------------------------------|---------------------------------------------------------------------------|
-| "resourceName" | Yes      | Endpoint resource name    | `string` - must be unique and should not contain special characters | "sriov_net_A"                                                             |
-| "selectors"    | No       | A map of device selectors | Each selector is a map of  `string` list.                           | "vendors": ["8086"], "devices": ["154c", "10ed"], "drivers": ["vfio-pci"] |
+| "resourceName" | Yes      | Endpoint resource name    | `string` - must be unique and should not contain special characters | "sriov_net_A"                                                                                      |
+| "selectors"    | No       | A map of device selectors | Each selector is a map of  `string` list.                           | "vendors": ["8086"], "devices": ["154c", "10ed"], "drivers": ["vfio-pci"], "pfNames": ["enp2s2f0"] |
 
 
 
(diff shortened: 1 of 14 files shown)


pkg/resources/pciNetDevice.go
```diff
@@ -30,11 +30,25 @@ func NewPciNetDevice(pciDevice *ghw.PCIDevice, rFactory types.ResourceFactory) (
...
 	// 			2. Get driver info
+	var ifName string
 	pciAddr := pciDevice.Address
 	driverName, err := utils.GetDriverName(pciAddr)
 	if err != nil {
 		return nil, err
 	}
+	netDevs, err := utils.GetNetNames(pciAddr)
+	if err != nil {
+		return nil, err
```
ahalim-intel: We need to consider some VF may not have a kernel netdev name(with uio, vfio drivers attached). In those cases we still want to add the devices into the device list. Maybe ignore the error here and only add ifname if it is available.

Mmduh-483: won't this make errors or bugs with other devices types?

ahalim-intel: > won't this make errors or bugs with other devices types?

This will break existing vfio/uio device support. What type of bugs/errors are we concerned about?

Mmduh-483: fixed

README.md
```diff
@@ -167,10 +169,10 @@ This plugin creates device plugin endpoints based on the configurations given in
...
 |----------------|----------|---------------------------|---------------------------------------------------------------------|---------------------------------------------------------------------------|
-| "resourceName" | Yes      | Endpoint resource name    | `string` - must be unique and should not contain special characters | "sriov_net_A"                                                             |
-| "selectors"    | No       | A map of device selectors | Each selector is a map of  `string` list.                           | "vendors": ["8086"], "devices": ["154c", "10ed"], "drivers": ["vfio-pci"] |
+| "resourceName" | Yes      | Endpoint resource name    | `string` - must be unique and should not contain special characters | "sriov_net_A"                                                                                      |
+| "selectors"    | No       | A map of device selectors | Each selector is a map of  `string` list.                           | "vendors": ["8086"], "devices": ["154c", "10ed"], "drivers": ["vfio-pci"], "ifnames": ["enp2s2f0"] |
```
ahalim-intel: Please a description of this selector  at the end of the section "Assumptions".

pkg/types/types.go
```diff
@@ -37,6 +37,7 @@ type ResourceConfig struct {
 		Vendors []string `json:"vendors,omitempty"`
 		Devices []string `json:"devices,omitempty"`
 		Drivers []string `json:"drivers,omitempty"`
+		NetDevs []string `json:"ifnames,omitempty"`
```
ahalim-intel: The filed name `NetDevs` is ambiguous. if its not too much to ask, lets make this field name(and related methods) obvious to what this selector is doing. I think "pfNames" goes well with this selector. What do you think? (I know the original PR was intended for actual NetDev name for VFs and you're probably reusing the var names etc. Lets try to keep the identifiers meaningful so less comments/docs are needed to explain them).

deployments/config.json
```diff
@@ -4,15 +4,17 @@
             "selectors": {
                 "vendors": ["8086"],
                 "devices": ["154c", "10ed"],
-                "drivers": ["i40evf", "ixgbevf"]
+                "drivers": ["i40evf", "ixgbevf"],
+                "ifnames": ["ens0", "enp2s0f1"]
```
ahalim-intel: Lets keep the sample deployment configs simple and more general so that user can run deploy them easily. For advanced use-cases proper documentation will be fine I'd say.

README.md
Mmduh-483: fixed

pkg/types/types.go
Mmduh-483: fixed

deployments/config.json
Mmduh-483: "pfNames" is removed from the first example

README.md
```diff
@@ -218,7 +219,7 @@ The device plugin will initially discover all PCI network resources in the host
 1. "vendors" - The vendor hex code of device
 2. "devices" - The device hex code of device
 2. "drivers" - The driver name the device is registered with
-
+4. "ifnames" - The Physical funtion name
```
ahalim-intel: I think this need to be changed to "pfNames" instead

PR #124, opened 2019-04-25
exclude default route interface only

zshi-redhat:


ahalim-intel:
We might have interface used by host for other routes not necessarily the default route. Is it not safer to assume any interface has a route entry; regardless default or not, are in use? Have you come across on any issue?

zshi-redhat:
${title} is addressed in another [PR](https://github.com/intel/sriov-network-device-plugin/pull/156), closing this one.

PR #126, opened 2019-05-07
Add pci.ids.gz file to image to avoid downloading it at runtime

pliurh:
`ghw.PCI()` trys to download pci.ids.gz form Internet if it cannot be found locally. We should not assume the container always has Internet access available. 

Event when the container has Internet access, due to there is no ca certificates installed, the following error pops up.
"error discovering host network devicesdiscoverDevices(): error getting PCI info: Get https://pci-ids.ucw.cz/v2.2/pci.ids.gz: x509: certificate signed by unknown authority"

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ RUN apk add --update --virtual build-dependencies build-base linux-headers && \
     make build
 
 FROM alpine
+ADD https://pci-ids.ucw.cz/v2.2/pci.ids.gz /usr/share/misc/pci.ids/
 COPY --from=builder /usr/src/sriov-network-device-plugin/build/sriovdp /usr/bin/
 WORKDIR /
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN apk add --update --virtual build-dependencies build-base linux-headers && \
     make build
 
 FROM alpine
-ADD https://pci-ids.ucw.cz/v2.2/pci.ids.gz /usr/share/misc/pci.ids/
+RUN apk add hwdata-pci
 COPY --from=builder /usr/src/sriov-network-device-plugin/build/sriovdp /usr/bin/
 WORKDIR /
 


Dockerfile
```diff
@@ -10,6 +10,7 @@ RUN apk add --update --virtual build-dependencies build-base linux-headers && \
     make build
 
 FROM alpine
+ADD https://pci-ids.ucw.cz/v2.2/pci.ids.gz /usr/share/misc/pci.ids/
```
ahalim-intel: Lets just install the package "hwdata-pci" instead.
`RUN apk add hwdata-pci`

pliurh: Fixed.

zshi-redhat: What is the equivalent package of hwdata-pci on centos or rhel?

ahalim-intel: @zshi-redhat "hwdata" should work on centos rhel. Looks like there's no hwdata-pci for centos which is a subpackage of hwdata in apk (include both pci.ids, usb.ids db). I suggested hwdata-pci for smaller image size.

zshi-redhat: @ahalim-intel thanks for checking it on centos!  sure, there is no problem using hwdata-pci for alpine image, just make sure that we'll have a equivalent package in RHEL so that it can be easily backported to openshift.

Dockerfile
```diff
@@ -10,6 +10,7 @@ RUN apk add --update --virtual build-dependencies build-base linux-headers && \
     make build
 
 FROM alpine
+RUN apk add hwdata-pci
```
zshi-redhat: Where will the package be installed, `/usr/share/misc/pci.ids/`?

ahalim-intel: > Where will the package be installed, `/usr/share/misc/pci.ids/`?

It will be installed at `/usr/share/hwdata/pci.ids`

Mmduh-483:
Downloading this package was causing the error attached in the image when using the daemonset with host ubuntu,
A workaround was to add the following code to line 17:
```
...
apk add ca-certificates
...
```
certificate err ubuntu

zshi-redhat:
> Downloading this package was causing the error attached in the image when using the daemonset with host ubuntu,
> A workaround was to add the following code to line 17:
> 
> ```
> ...
> apk add ca-certificates
> ...
> ```
> 
> To skip downloading it would be a better solution
> certificate err ubuntu

I was about to post the same result on centos:

```
[root]# kubectl logs kube-sriov-device-plugin-amd64-zbzw5 -n kube-system
I0528 11:02:22.997951   26315 manager.go:70] Using Kubelet Plugin Registry Mode
I0528 11:02:22.998424   26315 main.go:44] resource manager reading configs
I0528 11:02:22.998699   26315 manager.go:98] ResourceList: [{ResourceName:intel_sriov_netdevice Selectors:{Vendors:[8086] Devices:[154c 10ed] Drivers:[i40evf ixgbevf]}}]
I0528 11:02:22.998762   26315 main.go:60] Discovering host network devices
I0528 11:02:22.998776   26315 manager.go:179] discovering host network devices
E0528 11:02:23.256607   26315 main.go:62] error discovering host network devicesdiscoverDevices(): error getting PCI info: Get https://pci-ids.ucw.cz/v2.2/pci.ids.gz: x509: certificate signed by unknown authority
```
Maybe `hwdata` package is installed by default in `/usr/share/hwdata` not `/use/share/misc` which leads to downloading of the package again?

zshi-redhat:
Also in the `sriov-daemonset.yaml`, it mounts `/usr/share/hwdata` host path to container; perhaps we shall remove it if the `hwdata` is installed in the image?

ahalim-intel:
We need to remove the volume mount `/usr/share/hwdata` from `sriov-daemonset.yaml` as this will override the docker images internal `/usr/share/hwdata`. 
The "ghw" library looks into  `/usr/share/hwdata/` and  `/usr/share/misc/` for pci.ids database, if none found then it will proceed to download. 

So, if we install the hwdata package in docker image and remove the mount this issue should be gone.

PR #128, opened 2019-05-28
remove host mount for /usr/share/hwdata

ahalim-intel:
Since the hwdata package is being installed in docker image to
avoid downloading the pci.ids at runtime, the host mount is no
longer needed.

### DIFF
--- a/deployments/pod-sriovdp.yaml
+++ b/deployments/pod-sriovdp.yaml
@@ -21,9 +21,6 @@ spec:
     - mountPath: /etc/pcidp/
       name: config
       readOnly: true
-    - name: hwdata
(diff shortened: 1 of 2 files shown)


zshi-redhat:
@ahalim-intel thanks for the PR!
There is another pod [spec](https://github.com/intel/sriov-network-device-plugin/blob/feature/device_selectors/deployments/pod-sriovdp.yaml#L39) file we used to run device plugin in a regular pod instead of daemonset, perhaps also change that one? or shall we remove the regular pod spec which is somehow duplicated with the daemonset.

ahalim-intel:
Updated now

PR #142, opened 2019-06-26
Filtering devices of interfaces that do not belong namespace

fedepaol:
With docker in docker setup, what happens is we have the device listed under /sys/bus/pci/devices/ but the /net link is empty and it does not contain the interface name. This makes the plugin to crash (as described in https://github.com/intel/sriov-network-device-plugin/issues/141). 

With this PR, we filter out eventual host devices that are not listed in the current network namespace.

### DIFF
--- a/cmd/sriovdp/manager.go
+++ b/cmd/sriovdp/manager.go
@@ -207,9 +207,11 @@ func (rm *resourceManager) discoverHostDevices() error {
 				productName = string([]byte(productName)[0:37]) + "..."
 			}
 			glog.Infof("discoverDevices(): device found: %-12s\t%-12s\t%-20s\t%-40s", device.Address, device.Class.ID, vendorName, productName)
+			isUsed, _ := isInUse(device.Address)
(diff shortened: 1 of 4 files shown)


ahalim-intel:
@fedepaol Thanks for the PR. Since it was a runtime error coming in from the GetPFName() function I've submitted a [fix ](#143 )for that. With proper error returned from this function should resolve this issue. This error will be sufficient to skip devices that do not have interface name available. Let me know if this fixes your issue otherwise we will consider this PR.

fedepaol:
Sounds like it will solve the problem. Will try tomorrow and will let you
know.

Il mer 26 giu 2019, 21:06 Abdul Halim <notifications@github.com> ha scritto:

fedepaol:
(Thanks)

booxter:
The link in the comment to another PR is broken, here it is: https://github.com/intel/sriov-network-device-plugin/pull/143

fedepaol:
Guess this can be closed

Issue #141, opened 2019-06-26
Crash with docker in docker setup


fedepaol:
Hi, I am experiencing a crash of the plugin with a docker in docker setup. The cluster is running in [kind](https://github.com/kubernetes-sigs/kind) and I moved only the relevant sriov devices into the node namespace.

The thing is, the devices are present inside the container sysfs **but** the link name is not present inside the `/net` subfolder.

I think the dp plugin could filter the devices that do not belong to the current namespace.

This is from the host:
```
[root@xxx kubevirt]# ls /sys/bus/pci/devices/0000:01:00.0/net/
eno1
```
The same path `/sys/bus/pci/devices/0000:01:00.0/net/` exists inside the container (and gets fetched by the dp) but it's empty, causing the crash.

The crash stack / log is the following:
```
I0626 09:50:41.900492      14 manager.go:72] Using Deprecated Devie Plugin Registry Path
I0626 09:50:41.900550      14 main.go:44] resource manager reading configs
I0626 09:50:41.900768      14 manager.go:98] ResourceList: [{ResourceName:sriov IsRdma:false Selectors:{Vendors:[] Devices:[] Drivers:[] PfNames:[]}}]
I0626 09:50:41.900840      14 main.go:60] Discovering host network devices
I0626 09:50:41.900855      14 manager.go:179] discovering host network devices
I0626 09:50:41.992936      14 manager.go:209] discoverDevices(): device found: 0000:01:00.0	02          	Broadcom Inc. and...	NetXtreme BCM5720 2-port Gigabit Ethe...
panic: runtime error: index out of range

goroutine 1 [running]:
github.com/intel/sriov-network-device-plugin/pkg/utils.GetPfName(0xc0008844f5, 0xc, 0xf964e8, 0x0, 0x0, 0x0)
	/usr/src/sriov-network-device-plugin/.gopath/src/github.com/intel/sriov-network-device-plugin/pkg/utils/utils.go:69 +0x310
github.com/intel/sriov-network-device-plugin/pkg/resources.NewPciNetDevice(0xc0008be340, 0xb24540, 0xc00013c8a0, 0x0, 0x0, 0x4, 0x4)
	/usr/src/sriov-network-device-plugin/.gopath/src/github.com/intel/sriov-network-device-plugin/pkg/resources/pciNetDevice.go:46 +0xdb
main.(*resourceManager).discoverHostDevices(0xc0000d5b80, 0x0, 0xa656fa)
[...]

PR #143, opened 2019-06-26
fix a runtime error if PF interface name not found

ahalim-intel:
The GetPFName() is used to determine a device's PF name if it was
an SR-IOV VF. For non SR-IOV devices, it would try to get device's
corresponding kernel net interface name. This name is unavailable
if the interface is not in the same namespace(see #141) or the
device is enabled for dpdk mode.
In those cases this function should return appropriate errors to
the caller if no interface name is found.

Change-Id: Id7c6662a2a86f20eb91fcfc4d4f965cd31f9d9b0

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -65,13 +65,20 @@ func GetPfName(pciAddr string) (string, error) {
 		if os.IsNotExist(err) {
 			path := filepath.Join(sysBusPci, pciAddr, "net")
 			files, err = ioutil.ReadDir(path)
-			if err == nil {
-				return files[0].Name(), nil
+			if err != nil {
+				return "", err
 			}
+			if len(files) < 1 {
+				return "", fmt.Errorf("no interface name found for device %s", pciAddr)
+			}
+			return files[0].Name(), nil
 		}
 		return "", err
+	} else if len(files) > 0 {
+		return files[0].Name(), nil
 	}
-	return files[0].Name(), nil
+	return "", fmt.Errorf("no interface name found for device %s", pciAddr)
+
 }
 
 // IsSriovPF check if a pci device SRIOV capable given its pci address
--- a/pkg/utils/utils_test.go
+++ b/pkg/utils/utils_test.go
@@ -357,4 +357,37 @@ var _ = Describe("In the utils package", func() {
 			"0000:01:10.0", []string{"fake0", "fake1"}, false,
 		),
 	)
+
+	DescribeTable("getting PF names",
+		func(fs *FakeFilesystem, device string, expected string, shouldFail bool) {
+			defer fs.Use()()
+			actual, err := GetPfName(device)
+			Expect(actual).To(Equal(expected))
+			assertShouldFail(err, shouldFail)
+		},
+		Entry("device doesn't exist", &FakeFilesystem{}, "0000:01:10.0", nil, true),
+		Entry("device is a VF and interface name exists",
+			&FakeFilesystem{Dirs: []string{"sys/bus/pci/devices/0000:01:10.0/physfn/net/fakePF"}},
+			"0000:01:10.0", "fakePF", false,
+		),
+		Entry("device is a VF and interface name does not exist",
+			&FakeFilesystem{Dirs: []string{"sys/bus/pci/devices/0000:01:10.0/physfn/net/"}},
+			"0000:01:10.0", "", true,
+		),
+		Entry("device is a PF and interface name exists",
+			&FakeFilesystem{Dirs: []string{"sys/bus/pci/devices/0000:01:10.0/net/fakeIF"}},
+			"0000:01:10.0", "fakeIF", false,
+		),
+		Entry("device is a PF interface name does not exist",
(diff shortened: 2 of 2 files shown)


zshi-redhat:
From the code, it looks like `GetPfName()` will return an error if `/net` path doesn't container any interface name. but what I understood @fedepaol request is to silently ignore interfaces(or not returning any error) that don't have a name in that path and exclude those devices in the resource pool.

fedepaol:
Well, the main thing is to avoid the plugin to crash :-)

Said that, my patch was intentionally skipping devices not belonging to current namespace, but the final result of treating this as an error is the same (the device not added to the list).

pkg/utils/utils.go
```diff
@@ -66,12 +66,19 @@ func GetPfName(pciAddr string) (string, error) {
 			path := filepath.Join(sysBusPci, pciAddr, "net")
 			files, err = ioutil.ReadDir(path)
 			if err == nil {
-				return files[0].Name(), nil
+				if len(files) > 0 {
```
fedepaol: The extra nesting could be avoided to improve readability, by handling the errors first.
That would match the go tendency to have the happy path on the leftmost side.

```
   if err != nil {
       return "", err
   }
   if len(files) <= 0 {
       return "", fmt.Errorf("no interface name found for device %s", pciAddr)
   }
   return files[0].Name(), nil
```

pkg/utils/utils.go
```diff
@@ -66,12 +66,19 @@ func GetPfName(pciAddr string) (string, error) {
...
+					return files[0].Name(), nil
+				}
+				return "", fmt.Errorf("no interface name found for device %s", pciAddr)
 			}
 		}
 		return "", err
+	} else if len(files) > 0 {
+		return files[0].Name(), nil
+	} else {
+		return "", fmt.Errorf("no interface name found for device %s", pciAddr)
```
fedepaol: We could skip the final else leaving the last return as the main exit point of the function as in
```
        } else if len(files) > 0 {
		return files[0].Name(), nil
	}
	return "", fmt.Errorf("no interface name found for device %s", pciAddr)
```

fedepaol:
Well, the main thing is to avoid the plugin to crash :-)
Said that, my patch was more intentional in skipping devices not belonging to current namespace. The final result is almost the same (apart from logged errors) which is, the device is not added to the list.

zshi-redhat:
> Well, the main thing is to avoid the plugin to crash :-)
> 
> Said that, my patch was intentionally skipping devices not belonging to current namespace, but the final result of treating this as an error is the same (the device not added to the list).

ok, so the final result would be devices are not added in the list and there will be an error message for each excluded device: `glog.Errorf("discoverDevices() error adding new device: %q", err)`
I was thinking `discoverHostDevices` will eventually return the error print when device name not found. but it seems it just prints error message and doesn't return immediately.

fedepaol:
> > Well, the main thing is to avoid the plugin to crash :-)
> > Said that, my patch was intentionally skipping devices not belonging to current namespace, but the final result of treating this as an error is the same (the device not added to the list).
> 
> ok, so the final result would be devices are not added in the list and there will be an error message for each excluded device: `glog.Errorf("discoverDevices() error adding new device: %q", err)`
> I was thinking `discoverHostDevices` will eventually return the error print when device name not found. but it seems it just prints error message and doesn't return immediately.

Yes exactly

ahalim-intel:
@fedepaol @zshi-redhat & @booxter Thanks a bunch for your reviews.   I've update the PR. Some notes  below:

> ok, so the final result would be devices are not added in the list and there will be an error message for each excluded device: glog.Errorf("discoverDevices() error adding new device: %q", err)
I was thinking discoverHostDevices will eventually return the error print when device name not found. but it seems it just prints error message and doesn't return immediately.

The PF name is not mandatory, but we have this filed available if needed for VF health monitoring. We used to do VFs health check with their corresponding PF. For non SR-IOV nic this would be their own interface name. Without the kernel net interface it's not possible to determine device health.
So, if we couldn't get the interface names(errors, not available) we simply do not add those devices in the device pool.

zshi-redhat:
> @fedepaol @zshi-redhat & @booxter Thanks a bunch for your reviews. I've update the PR. Some notes below:
> 
> > ok, so the final result would be devices are not added in the list and there will be an error message for each excluded device: glog.Errorf("discoverDevices() error adding new device: %q", err)
> > I was thinking discoverHostDevices will eventually return the error print when device name not found. but it seems it just prints error message and doesn't return immediately.
> 
> The PF name is not mandatory, but we have this filed available if needed for VF health monitoring. We used to do VFs health check with their corresponding PF. For non SR-IOV nic this would be their own interface name. Without the kernel net interface it's not possible to determine device health.
> So, if we couldn't get the interface names(errors, not available) we simply do not add those devices in the device pool.

Agree with @ahalim-intel on filtering out those devices at this point when interface name is not found, as device health check rely on the name to find the link state. This might be changed in the future when we have a new solution on how to check the health device.

/lgtm

ahalim-intel:
Thank you all for the reviews

PR #148, opened 2019-07-11
Add Link Type selector

adrianchiris:
Add the ability to filter according to a device's link type.
This will enable a user to filter SR-IOV resources according
to link type, allowing the creation of resource pools based on
the underlying fabric a device is connected to
(e.g ethernet, infiniband).

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ This plugin creates device plugin endpoints based on the configurations given in
 |     Field      | Required |        Description        |                      Type - Accepted values                       |                                      Example/Accepted values                                       |
 |----------------|----------|---------------------------|-------------------------------------------------------------------|----------------------------------------------------------------------------------------------------|
 | "resourceName" | Yes      | Endpoint resource name    | string - must be unique and should not contain special characters | "sriov_net_A"                                                                                      |
-| "selectors"    | No       | A map of device selectors | Each selector is a map of string list.                            | "vendors": ["8086"], "devices": ["154c", "10ed"], "drivers": ["vfio-pci"], "pfNames": ["enp2s2f0"] |
+| "selectors"    | No       | A map of device selectors | Each selector is a map of string list.                            | "vendors": ["8086"], "devices": ["154c", "10ed"], "drivers": ["vfio-pci"], "pfNames": ["enp2s2f0"], "linkTypes": ["ether"] |
 | "isRdma"       | No       | Mount RDMA resources      | `bool` - boolean value true or false                              | "isRdma": true                                                                                     |
 
 
@@ -230,6 +230,7 @@ The device plugin will initially discover all PCI network resources in the host 
 2. "devices" - The device hex code of device
 3. "drivers" - The driver name the device is registered with
 4. "pfNames" - The Physical funtion name
+5. "linkTypes" - The link type of the net device associated with the PCI device.
 
 ### Workflow
 
--- a/README.md
+++ b/README.md
@@ -168,6 +168,13 @@ This plugin creates device plugin endpoints based on the configurations given in
                 "devices": ["1018"],
                 "drivers": ["mlx5_ib"]
             }
+        },
+        {
+            "resourceName": "infiniband_rdma_netdevs",
+            "isRdma": true,
+            "selectors": {
+                "linkTypes": ["infiniband"]
+            }
         }
     ]
 }
--- a/pkg/resources/factory.go
+++ b/pkg/resources/factory.go
@@ -75,6 +75,8 @@ func (rf *resourceFactory) GetSelector(attr string, values []string) (types.Devi
 		return newDriverSelector(values), nil
 	case "pfNames":
 		return newPfNameSelector(values), nil
+	case "linkTypes":
+		return newLinkTypeSelector(values), nil
 	default:
 		return nil, fmt.Errorf("GetSelector(): invalid attribute %s", attr)
 	}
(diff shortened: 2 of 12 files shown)


adrianchiris:
above force push is due to accidental file permission changes of the original PR

pkg/resources/pciNetDevice.go
```diff
@@ -60,6 +61,15 @@ func NewPciNetDevice(pciDevice *ghw.PCIDevice, rFactory types.ResourceFactory) (
 		Health: pluginapi.Healthy,
 	}
 
+	linkType := ""
+	if len(ifName) > 0 {
```
ahalim-intel: Also need to check if ifName is not empty("")

pkg/resources/pciNetDevice.go
```diff
@@ -60,6 +61,15 @@ func NewPciNetDevice(pciDevice *ghw.PCIDevice, rFactory types.ResourceFactory) (
...
+		la, err := utils.GetLinkAttrs(ifName)
+		if err != nil {
+			return nil, err
```
ahalim-intel: I'd prefer not to return nil for not detecting linktype here. Maybe add linktype info only if we can get it otherwise leave it empty like we do for ifName. There are times when a VF may not be represented as a netdevice thus those info won't be available. But "selectors" will be able to capture only those devices that have these information when applied.

pkg/resources/pciNetDevice.go
adrianchiris: If ifName is empty then its length would be zero.

https://play.golang.org/p/cDCkPWTJydn

pkg/resources/pciNetDevice.go
adrianchiris: For ifName, indeed you may not have a netdevice in case the device is not bounded to a network driver (hence ifName will be empty).
However, if we get into this condition then we have a netdevice (ifName is not empty)
Failing to get link attributes may point to some system fault, are we OK with ignoring that ?

pkg/resources/pciNetDevice.go
ahalim-intel: duh! For some reason I was thinking ifName was a slice of names(we had that in sriov-cni where a VF could have more than one interface name)

adrianchiris: :)

pkg/resources/pciNetDevice.go
ahalim-intel: Good call. I think we can safely assume that if we can get a valid interface name then link attributes should be available. I think it's good idea to return the err. The discovery method logs this error anyway. Maybe we could add some more info into the error before returning?

adrianchiris: So, utils.GetLinkAttrs() returns an informative error in case of failure that includes the netlink error:

`func GetLinkAttrs(ifName string) (*nl.LinkAttrs, error) {
	link, err := getLinkByName(ifName)
	if err != nil {
		return nil, fmt.Errorf("error getting link attributes for net device %s %v", ifName, err)
	}
	return link.Attrs(), nil
}`

It will propagate up the stack and will be reflected in the discovery method.

pkg/resources/factory.go
```diff
@@ -112,6 +114,16 @@ func (rf *resourceFactory) GetResourcePool(rc *types.ResourceConfig, deviceList
 		}
 	}
 
+	// filter by linkTypes list
+	if rc.Selectors.LinkTypes != nil && len(rc.Selectors.LinkTypes) > 0 {
+		if len(rc.Selectors.LinkTypes) > 1 {
+			glog.Warningf("Link type selector should have a single value.")
```
zshi-redhat: If this should be a single value,  why using a `list` instead of `string`?

README.md
```diff
@@ -180,7 +180,7 @@ This plugin creates device plugin endpoints based on the configurations given in
 |     Field      | Required |        Description        |                      Type - Accepted values                       |                                      Example/Accepted values                                       |
 |----------------|----------|---------------------------|-------------------------------------------------------------------|----------------------------------------------------------------------------------------------------|
 | "resourceName" | Yes      | Endpoint resource name    | string - must be unique and should not contain special characters | "sriov_net_A"                                                                                      |
-| "selectors"    | No       | A map of device selectors | Each selector is a map of string list.                            | "vendors": ["8086"], "devices": ["154c", "10ed"], "drivers": ["vfio-pci"], "pfNames": ["enp2s2f0"] |
+| "selectors"    | No       | A map of device selectors | Each selector is a map of string list.                            | "vendors": ["8086"], "devices": ["154c", "10ed"], "drivers": ["vfio-pci"], "pfNames": ["enp2s2f0"], "linkTypes": ["ether"] |
```
zshi-redhat: what about adding an example config in above `resourceList`?
Q: Is there any other `linkType` other than `ether`, `infiniband`?

pkg/resources/factory.go
adrianchiris: providing multiple link types will technically work (although i dont think its useful - hence the warning)
The reason i chose to keep it as a list of strings is two fold:
1. Keep consistent user API for specifying a selector
2. Supporting a single string value as a selector (instead of a list of strings) will require a refactor
    as it does not fit the current selector model.

README.md
adrianchiris: Sure, will add a new entry .

in regards to link types, netlink go lib defines the string values in: https://github.com/vishvananda/netlink/blob/b1e98597921720cee4147d6476110fc8fc56072d/nl/nl_linux.go#L111

which are derived from: https://github.com/torvalds/linux/blob/master/include/uapi/linux/if_arp.h

adrianchiris: Sure, will add a new entry  in the resourceList above.

in regards to link types, netlink go lib defines the string values in: https://github.com/vishvananda/netlink/blob/b1e98597921720cee4147d6476110fc8fc56072d/nl/nl_linux.go#L111

which are derived from: https://github.com/torvalds/linux/blob/master/include/uapi/linux/if_arp.h

pkg/resources/factory.go
zshi-redhat: ok

PR #156, opened 2019-08-08
only exclude default route interface

zshi-redhat:
When PF is connected to a dhcp network, creation of VFs
will result in IP address being assigned to each VF. Current
Device Plugin ignore such devices and doesn't count it as
allocatable devices.

This commit changes such behavior by checking and filtering
out device with default route, leaving other devices as
available resource.

### DIFF
--- a/cmd/sriovdp/manager.go
+++ b/cmd/sriovdp/manager.go
@@ -209,7 +209,7 @@ func (rm *resourceManager) discoverHostDevices() error {
 			glog.Infof("discoverDevices(): device found: %-12s\t%-12s\t%-20s\t%-40s", device.Address, device.Class.ID, vendorName, productName)
 
 			// exclude device in-use in host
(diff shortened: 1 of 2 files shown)


ahalim-intel:
I understood the problem. Just wondering just checking for default route is enough to determine if a link is in use or not. There will be just one interface with default route entry. How do we exclude some interfaces that might be used by the host to connect to different network, not necessarily the default route one?

zshi-redhat:
> I understood the problem. Just wondering just checking for default route is enough to determine if a link is in use or not. There will be just one interface with default route entry. How do we exclude some interfaces that might be used by the host to connect to different network, not necessarily the default route one?

If it's a kubernetes managed node, I'd imagine that all these resources will be first exposed by device plugin (for non-SR-IOV devices, this may be done via another device plugin), then used in a pod, ideally these devices will not be configured directly by user. 
If this does happen, user has the flexibility to filter devices based on interface name if used.

PR #158, opened 2019-08-13
Add DANM as a compatible meta plugin option to the README, in addition to Multus

Levovar:
The document was generalized to mention that the DP works together with any meta plugin supporting the DP based network provisioning.

Deployment example continue to mention Multus in the first place, but also added DANM to showcase the flexibility of the plugin.

Also touched up the document at a couple of places: broken links deleted, out of date information updated, some English corrected.

I rearranged the order of proposed installation because IMO it is more logical to start with the CNI and the DP first, but can change it back if you prefer the original way :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ The SRIOV network device plugin is Kubernetes device plugin for discovering and 
 
 To deploy workloads with SRIOV VF this plugin needs to work together with the following two CNI components:
 
-- A compatible CNI meta plugin (Multus CNI, or DANM)
+- Any CNI meta plugin supporting Device Plugin based network provisioning (Multus CNI, or DANM)
 
   - Retrieves allocated network device information of a Pod
 
@@ -102,10 +102,10 @@ $ make
 $ make image
 ``` 
 
-### Install a compatible CNI meta plugin
+### Install one compatible CNI meta plugin 
 
-#### Multus
-This section explains an example deployment of SRIOV Network device plugin in Kubernetes with Multus.
+#### Option 1 - Multus
+This section explains an example deployment of SRIOV Network device plugin in Kubernetes if you choose Multus as your meta plugin.
 Required YAML files can be found in [deployments/](deployments/) directory.
 
 #### Build and configure Multus
@@ -135,8 +135,8 @@ Multus uses Custom Resource Definitions(CRDs) for defining additional network at
 $ kubectl create -f deployments/sriov-crd.yaml
 ```
 
-#### DANM
-This section explains an example deployment of SRIOV Network device plugin in Kubernetes with DANM.
+#### Option 2 - DANM
+This section explains an example deployment of SRIOV Network device plugin in Kubernetes if you choose DANM as your meta plugin.
 
 #### Install DANM 
 Refer to [DANM documentation](https://github.com/nokia/danm#getting-started) for detailed instructions.
@@ -239,7 +239,7 @@ Usage of ./sriovdp:
 
 This plugin does not bind or unbind any driver to any device whether it's PFs or VFs. It also doesn't create Virtual functions either. Usually, the virtual functions are created at boot time when kernel module for the device is loaded. Required device drivers could be loaded on system boot-up time by white-listing/black-listing the right modules. But plugin needs to be aware of the driver type of the resources(i.e. devices) that it is registering as K8s extended resource so that it's able to create appropriate Device Specs for the requested resource.
 
-For exmaple, if the driver type is uio(i.e. igb_uio.ko) then there are specific device files to add in Device 
+For example, if the driver type is uio(i.e. igb_uio.ko) then there are specific device files to add in Device 
 Spec. For vfio-pci, device files are different. And if it is Linux kernel network driver then there is no device file to be added.
 
 The idea here is, user creates a resource config for each resource pool as shown in [Config parameters](#config-parameters) by specifying the resource name, a list resource "selectors".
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,37 +127,24 @@ $ cp deployments/cni-conf.json /etc/cni/net.d/
...
-```
-2. Create the SRIOV Network CRD
+Multus uses Custom Resource Definitions(CRDs) for defining additional network attachements. These network attachment CRDs follow the standards defined by K8s Network Plumbing Working Group(NPWG). Please refer to [Multus documentation](https://github.com/intel/multus-cni/blob/master/README.md) for more information.
+1. Create the SRIOV Network CRD
 ```
 $ kubectl create -f deployments/sriov-crd.yaml
 ```
 
-### Build and run SRIOV network device plugin
+#### DANM
```
zshi-redhat: It might be good to have an example as Multus in Readme file on how to build the meta plugin and create sriov network attachment. This may require creating templates in the deployment folder.

README.md
```diff
@@ -101,37 +127,24 @@ $ cp deployments/cni-conf.json /etc/cni/net.d/
...
-```      
-> On successful build the `sriovdp` executable can be found in `./build` directory. It is recommended to run the plugin in a container or K8s Pod. The follow on steps cover how to build and run the Docker image of the plugin.
+#### Install DANM 
+Refer to [DANM documentation](https://github.com/nokia/danm#getting-started) for detailed instructions.
 
- 3. Build docker image
- ```
-$ make image
-``` 
+#### Create SR-IOV type networks
```
zshi-redhat: Is this similar to using sriov-cni to create network attachments?

zshi-redhat:
@Levovar any update?

README.md
Levovar: Those are already available in the DANM repository, so I chose simply link them, rather than duplicate them here.
I generally don't like duplicating things, because one of them will ultimately fall out of synch.
do you see any issue with simply linking the appropriate artifacts?

README.md
Levovar: it is a CRD based API if that's what you are asking. so in a sense, it is similar, yes.
it is different API though. we don't expose the specifics of each and every CNI backend as Multus does, instead we provide one, generic, abstract API

as above, the necessary manifests showing off how one can create an SR-IOV type network is linked to the README enhancement

Levovar:
> @Levovar any update?

my bad, I was on my summer holiday :) back now, answered the comments. basically I opted to link manifests rather than copy them over here, but if you prefer the duplication I can do that

README.md
ahalim-intel: +1
I think link to the DANM is good for its own documentations and sample deployment.

ahalim-intel:
@zshi-redhat Do you have any further comments for this PR?

README.md
zshi-redhat: I may have different thought on this, if we list DANM as an alternative meta plugin of Multus, then we should treat them equally, this may include:
1) provide simple example that prove to be functional in device plugin doc
2) add new tests where fits for using device plugin with DANM, this may require adding new tests in CI (ongoing work)
3) evaluate other parts of doc to make sure all the description to meta plugin also applies to DANM, otherwise, we need to add note or change the doc.

Above may not be needed if the goal is just to point user for another meta plugin solution and we don't guarantee it always works. If that's the case, we perhaps also want to add a note for that and set the right expectation for users.

Levovar: I don't consider myself as a part of the community, so the decision is yours! I'm okay either way :)
couple questions to your points:
1: examples meaning manifests? or proof from a real installation that it works?
2: Unit Tests, or functional tests? because from unit test perspective I don't think the meta plugin has any interface to the DP.
if you mean functional tests, and already have an automated K8s + Multus + DP installation sure, we can work on creating one such setup with DANM too.
I think it might take more time though, so maybe we could approach this iteratively. Or maybe use the CNCF Testbed project for this?
3: if you see other parts which need some face lifting feel free to point out! I went through and I think I updated whatever needed updating, but if you see anything else I will change that too

I will try and join the next netowrking / resource management scrum call on Monday, we could discuss the details and additional action items online!

ahalim-intel: Ok, I went through the entire document flow again. I can see why @zshi-redhat is asking about the example deployments. 

With this change, the "Install a compatible CNI meta plugin" section doesn't tell user to pick ONE meta plugin - Multus or DANM, not both. Users who are new to this may as well proceed with both steps.  ¯\\\_(ツ)\_/¯

Worse even, the "Example deployments" section doesn't mention that it is based on the assumption that K8s is setup with Multus(at least this was the case before we introduced DANM here). Users who may want to try this out with DANM may think that, this example deployments should work with either meta plugin.

Maybe we should try to make it clearer to the users that the example deployments in this document assume that Multus is configured. To deploy with DANM they should look into relevant section provided by DANM instead adding entire DANM deployments on this doc.

Levovar: Sure! I will make the first point more explicit, and maybe fork the example output section too and add some outputs with DANM (thought most of the printouts will be similar)

Levovar: done, please re-check. the example manifests are still live under the DANM repository, but the README here now contains very detailed step-by-step examples how one can deploy them

also added example test results to showcase it works :)

Levovar:
bump
@ahalim-intel , @zshi-redhat any feedback on the new commit?

zshi-redhat:
> bump
> @ahalim-intel , @zshi-redhat any feedback on the new commit?

@Levovar thanks for the update! 
/lgtm

PR #166, opened 2019-08-22
update sample deployment manifests

ahalim-intel:
This update corrects the resource references that match with existing configMap.
Also fixes a problem where Network Status not being populated due to cniVersion was missing
for sriov network CRD.

Change-Id: I2ef7104ec3b94627698eb21b0708a79adcce7f59


zshi-redhat:
@ahalim-intel  what about also changing the other two pods templates (pod-tc2.yaml, pod-tc3.yaml)?

ahalim-intel:
> @ahalim-intel what about also changing the other two pods templates (pod-tc2.yaml, pod-tc3.yaml)?

Good question. The other 2 pods specs are same as pod-tc1.yaml, just wondering if we need them at all or can we just remove them?

zshi-redhat:
> > @ahalim-intel what about also changing the other two pods templates (pod-tc2.yaml, pod-tc3.yaml)?
> 
> Good question. The other 2 pods specs are same as pod-tc1.yaml, just wondering if we need them at all or can we just remove them?

I sometime use them for multiple pod testing, for example, to check the network connectivity between two pods on same and different nodes. but this can be achieved by copying pod-tc1.yml.
so let's remove the same pod files, or maybe replace them with dpdk examples.

ahalim-intel:
> I sometime use them for multiple pod testing, for example, to check the network connectivity between two pods on same and different nodes. but this can be achieved by copying pod-tc1.yml.
> so let's remove the same pod files, or maybe replace them with dpdk examples.

The dpdk example needs different configMap and net CRD. Let's keep the pod-tc2.yaml, with correct resource reference of course and remove other one.

zshi-redhat:
> > I sometime use them for multiple pod testing, for example, to check the network connectivity between two pods on same and different nodes. but this can be achieved by copying pod-tc1.yml.
> > so let's remove the same pod files, or maybe replace them with dpdk examples.
> 
> The dpdk example needs different configMap and net CRD. Let's keep the pod-tc2.yaml, with correct resource reference of course and remove other one.

+1

PR #168, opened 2019-08-31
add go mod support

zshi-redhat:


ahalim-intel:
@zshi-redhat Thank for this PR. Do we have to continue support the vendoring or this is for a transitioning period?

zshi-redhat:
> @zshi-redhat Thank for this PR. Do we have to continue support the vendoring or this is for a transitioning period?

For a transitioning period.

This is mainly for backward compatibility support, if user 
1) doesn't use new go versions(1.11 or 1.12),
2) GO111MODUELS is not enabled,
3) no internet connection
Then vendor pkgs are required to built sriovdp binary.

I think we will eventually remove all the vendor pkgs in the repo(it can be easily re-generated via go module),wdyt?

ahalim-intel:
Ok, make sense. 
So, can we remove the glide dependency and use "go mod -v" to generate vendor where needed or do we will still need glide for older Go? Just so that a developer do not need to update glide files any longer.

zshi-redhat:
> Ok, make sense.
> So, can we remove the glide dependency and use "go mod -v" to generate vendor where needed or do we will still need glide for older Go? Just so that a developer do not need to update glide files any longer.

Yes, we can remove glide dependency.
I didn't try with older go version, but I think for developer who needs to update package dependency, we can enforce to use go module instead of glide, then remove all the glide files.
For user who wants to build sriovdp binary locally, we can keep vendor pkg there so they don't need to self generate vendor pkgs which requires new go versions.

PR #173, opened 2019-09-05
remove glide files & update travis-ci for go mod

ahalim-intel:
Since we have added go module support with #168 the glide files
are no longer needed to manage vendor.

Also added travis-ci build matrix for go v1.11 and v1.12 with
GO111MODULE set 'on' and 'off'.

Change-Id: Iec4108e50c557904f31a60a412d644332a039425


ahalim-intel:
@zshi-redhat  Should we remove the glide files or keep them? As mentioned in #160 there seem to be issues with adding kubernetes device manager api dependencies using go mod vendor.

Then again keeping both glide and go mod to manage the vendor may result in conflicts.

zshi-redhat:
> @zshi-redhat Should we remove the glide files or keep them? As mentioned in #160 there seem to be issues with adding kubernetes device manager api dependencies using go mod vendor.
> 
> Then again keeping both glide and go mod to manage the vendor may result in conflicts.

Yes, I think we should use either glide or go mode to manage the vendor.
We changed to use go mod in #168 which updated all vendor pkgs, if move back to glide, then we will need to revert vendor pkg back to previous state using glide.
Re the issue in #160, it seems a problem of kubernetes pkg, not go mod. I'd say let's track that issue in https://github.com/kubernetes/kubernetes/issues/82437. If we want to have a version that supports NUMA for testing, would it be good to create a separate development branch to hold that patch?

ahalim-intel:
Closing this PR as this has been resolved by #225

PR #175, opened 2019-09-11
Fixing all entries in the README's table of contents

Levovar:
After the merge of the README update I realized I forgot to update the table of contents.
Then when I started fixing it I realized others also forgot to do it :)

I rechecked all entries and everything should be now up-to-date, clickable, and properly indented!


Levovar:
@ahalim-intel @zshi-redhat

PR #179, opened 2019-09-18
update TopologyInfo to send list of NUMANode

ahalim-intel:
The K8s device manager api has been updated to include multiple NUMANode infos instead of a single one. Please see [here](https://github.com/kubernetes/kubernetes/pull/82020) and [here](https://github.com/kubernetes/kubernetes/pull/80570).


zshi-redhat:
@ahalim-intel do you know if there is use case for SR-IOV interface to be associated with more than one NUMA node?
Also I checked the environment I tested yesterday, the `repeated field` change are there, but curious why I didn't hit any failure without this PR.

ahalim-intel:
Hi @zshi-redhat the rational for this changes discussed in lengths you can find them on the K8s PRs listed in this PRs description. 

@klueska is the author of those PR and might be able to explain it better. But in my understanding from the discussions - some device can be associated with more than one NUMA node. But most PCI devices(like SR-IOV) shoul have a single NUMA. So, we are good to send one instance of NUMANode in the TopologyInfo.

> Also I checked the environment I tested yesterday, the kubernetes code has repeated field change, but curious why I didn't hit any failure without this PR.

You might have slightly older commit for K8s. If you build it from v1.16 it should fail or ignored because unknown field.

cc @klueska

klueska:
> @ahalim-intel do you know if there is use case for SR-IOV interface to be associated with more than one NUMA node?

I'm not sure if there is a use case for an SR-IOV interface to ever be associated with multiple NUMA nodes, but we designed the Device Plugin API extension to be able to work with other devices that might need this.

Here is a link to the discussion on this: https://github.com/kubernetes/kubernetes/pull/74423#discussion_r317580396

zshi-redhat:
Tested with release-1.16 with single-numa-node policy, device plugin works as expected. All sriov pods can get VF resource from allocated from kubelet.
/lgtm

Issue #186, opened 2019-11-28
make did not build docker image by default Makefile


xuguozhen:
.PHONY: all
all: fmt lint build
https://github.com/intel/sriov-network-device-plugin/blob/master/Makefile#L49
All did not contain image building,make could not build docker image.

zshi-redhat:
@xuguozhen `make image` should be able to build docker image, `make` only build the binary.

killianmuldoon:
@zshi-redhat I think this is still the appropriate behaviour. Can we close this?

PR #187, opened 2019-11-28
enable Makefile building docker image by default

xuguozhen:
Fixes：#186
Signed-off-by： xuguozhen <xuguozhen@inspur.com>


killianmuldoon:
@xuguozhen is there a use case where you feel this is necessary? Right now for development building the image takes an order of magnitude longer than building the binary - so by default it's quicker to use make this way.

killianmuldoon:
Closing as stale.

PR #194, opened 2020-01-20
make resource prefixes configurable for individual resource pools

przemeklal:
* Introduce new optional configuration field "resourcePrefix" that allows
to override resource prefix on a resource pool basis instead of setting
it globablly for all resource pools.
* Support configuration with duplicate resource names, but distinct
prefixes, e.g. "prefixA.com/10G" and "prefixB.com/10G" won't return
validation error, as prefixes are now taken into account as well.
* Include resource prefix in the socket name, so that resources with the
same name but different prefix are supported.
* Update ResourcePool mock.
* Fix tests.
* Update README.

Signed-off-by: Przemyslaw Lal <przemyslawx.lal@intel.com>


przemeklal:
TODOs:
- [x] initial implementation
- [ ] small fix in the validation code (use global prefix as well when comparing names)
- [ ] fix tests
- [ ] document new configuration option

przemeklal:
Main reason for this PR is a newly introduced SRIOV Network Service Mesh forwarder (work in progress) that takes use of the SRIOV Network Device Plugin for SRIOV VF resources allocation.
With configurable resourcePrefixes, injection of resources via NSM admission webhook can be greatly simplified. Also, plan is to make this config field fully optional here, so that we don't break existing configurations in the process. Corresponding PR for NSM can be found here:
https://github.com/networkservicemesh/networkservicemesh/pull/1925

przemeklal:
I'm marking this as a draft PR for the moment as there's couple of todos left.
Intent is to make this available early for the folks that want to play with the current implementation of the SRIOV forwarder in NSM.

przemeklal:
Let's put this on hold as there's ongoing design related discussion in the NSM community.

cmd/sriovdp/manager.go
```diff
@@ -153,23 +153,33 @@ func (rm *resourceManager) stopAllServers() error {
...
 		// check if name contains acceptable characters
 		if !utils.ValidResourceName(conf.ResourceName) {
 			glog.Errorf("resource name \"%s\" contains invalid characters", conf.ResourceName)
 			return false
 		}
-		// check resource names are unique
-		_, ok := resourceName[conf.ResourceName]
-		if ok {
+
+		// resourcePrefix might be is overriden for a given resource pool
```
ahalim-intel: nit: " might be is" > "might be"

przemeklal: good catch, fixed

PR #195, opened 2020-01-27
Add pciAddresses selectors for VM usecase

pperiyasamy:
This makes device plugin to support VM usecase in which
vf (or) virtio devices can be enumerated into a device
pool using its pci addresses.

Signed-off-by: Periyasamy Palanisamy <periyasamy.palanisamy@est.tech>

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,11 +232,12 @@ The "deviceType" value determines which selectors are supported for that device.
 #### Common selectors
 All device types support following common device selectors.
 
-|   Field   | Required |                Description                |         Type/Defaults          |   Example/Accepted values   |
-|-----------|----------|-------------------------------------------|--------------------------------|-----------------------------|
-| "vendors" | N        | Target device's vendor Hex code as string | `string` list Default: `null`  | "vendors": ["8086"]         |
-| "devices" | N        | Target Devices' device Hex code as string | `string` list Default: `null`  | "devices": ["154c", "10ed"] |
-| "drivers" | N        | Target device driver names as string      | `string` list Default: `null`  | "drivers": ["vfio-pci"]     |
+|   Field        | Required |                Description                |         Type/Defaults          |   Example/Accepted values        |
+|----------------|----------|-------------------------------------------|--------------------------------|----------------------------------|
+| "vendors"      | N        | Target device's vendor Hex code as string | `string` list Default: `null`  | "vendors": ["8086"]              |
+| "devices"      | N        | Target Devices' device Hex code as string | `string` list Default: `null`  | "devices": ["154c", "10ed"]      |
+| "drivers"      | N        | Target device driver names as string      | `string` list Default: `null`  | "drivers": ["vfio-pci"]          |
+| "pciAddresses" | N        | Target device's pci address as string     | `string` list Default: `null`  | "pciAddresses": ["0000:03:02.0"] |
 
 
 #### Extended selectors for device type "netDevice"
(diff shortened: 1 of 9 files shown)


zshi-redhat:
Hi @pperiyasamy thanks for your PR?
Is this to group individual VF into one resource pool for VM use case where PF is not accessable ?
Is it possible to use vendor and device selectors instead of pfNames selector to achieve the same purpose?

pperiyasamy:
@zshi-redhat Yes, this is for the use case where PF is not accessible and left/right VF (or virtio) devices having same vendor and device id. hence we need to enumerate using their names to create left and right resource pools.

zshi-redhat:
> @zshi-redhat Yes, this is for the use case where PF is not accessible and left/right VF (or virtio) devices having same vendor and device id. hence we need to enumerate using their names to create left and right resource pools.

From current implemention of `GetPfName`, it looks like if PF doesn't exist, it will return VF interface name. Would you mind trying to use VF interface name in `pfName` selector to see if it can satisfy your request? I didn't try that myself before, just thought that might be an option when looking at the PR.

pperiyasamy:
> From current implemention of `GetPfName`, it looks like if PF doesn't exist, it will return VF interface name.

@zshi-redhat I don't think so. because it looks for net directory inside physfn which may not exist inside the VM. Moreover there is no interface name associated with the VF/virtio device bound with dpdk driver.

pperiyasamy:
@zshi-redhat any update on this ?

zshi-redhat:
I think we probably don't want to use PCI address as a fake PF name. It might be good to come up a solution to support using DP in VM instead of having small fixes for each use case.

pperiyasamy:
> I think we probably don't want to use PCI address as a fake PF name. It might be good to come up a solution to support using DP in VM instead of having small fixes for each use case.

Okay, Can we properly fix it by adding vfNames in selectors?  Here is the issue #205

lynic:
> I think we probably don't want to use PCI address as a fake PF name. It might be good to come up a solution to support using DP in VM instead of having small fixes for each use case.

HI zshi, do you have any new ideas on this PR? We also need this filter to support VM case, any update for the moment?

pperiyasamy:
> > I think we probably don't want to use PCI address as a fake PF name. It might be good to come up a solution to support using DP in VM instead of having small fixes for each use case.
> 
> HI zshi, do you have any new ideas on this PR? We also need this filter to support VM case, any update for the moment?

@lynic I'm still hoping to rework on this PR depending on feedback [here](https://github.com/intel/sriov-network-device-plugin/issues/205).

ahalim-intel:
@pperiyasamy  I think a selector for device pci address is desirable and possblu address the VM usecases. Let's come up with more meaningful parameter name instead of "vfDevices". This selector could be applied for any pci devices -regardless of VF or PF.

Could you please update this PR?

pperiyasamy:
> @pperiyasamy I think a selector for device pci address is desirable and possbly address the VM usecases. Let's come up with more meaningful parameter name instead of "vfDevices". This selector could be applied for any pci devices -regardless of VF or PF.
> 
> Could you please update this PR?

@ahalim-intel Thanks for your attention now, updated the PR replacing "vfDevices" parameter with "pciAddresses". Hope this is more relevant and generic.

zshi-redhat:
might be worth updating the readme file.

pperiyasamy:
@zshi-redhat Thanks, will update the readme. I just noticed `pciAddresses` is part of `netDevice` selectors. shouldn't it move into `Common` selectors ?

ahalim-intel:
> @zshi-redhat Thanks, will update the readme. I just noticed `pciAddresses` is part of `netDevice` selectors. shouldn't it move into `Common` selectors ?

Yes, this selector should be common and go in `DeviceSelectors` struct instead.

pperiyasamy:
Hi, addressed the comments on `readme `file and moving `pciAddresses `parameter. 
can you have a look now ?

ahalim-intel:
updated commit header

lynic:
This PR looks ready to be merged, is there any plan to get it in?

PR #196, opened 2020-01-30
SRIOV support for non-network device (FEC)

damiankopyto:
SRIOV feature is not just applicable for Network interfaces but any IO device that support PCIe SIG SRIOV features. One such device is based on Intel® PAC N3000 FPGA which has Network Functionality of 2x25G and FPGA part that can be programmed to support hardware offload of the  Wireless (4G/5G) Forward Error Correction (FEC). This Product has already been released by Intel and is being used/deployed by customers. Intel FPGA Device plugin (https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/cmd/fpga_plugin/README.md) does not support such devices as it dose not cover SRIOV devices using DPDK and also differ in the FPGA image that gets programmed. As part of this PR we are recommending to extend the SRIOV support for non-networking devices and also rename the device plugin from “sriov-network-device-plugin” to “sriov-device-plugin”. For the Intel® PAC N3000 FPGA supporting vRAN or Cloud-native RAN the FPGA PF that supports FEC is bound to DPDK igb_uio or vfio_pci. The Virtual Function resources are allocated to the PODs running RAN/FlexRAN like workload and uses vfio_pci.

### DIFF
--- a/cmd/sriovdp/manager.go
+++ b/cmd/sriovdp/manager.go
@@ -31,8 +31,9 @@ import (
 
 const (
 	socketSuffix = "sock"
-	netClass     = 0x02 // Device class - Network controller.	 ref: https://pci-ids.ucw.cz/read/PD/02 (for Sub-Classes)
-	fecClass     = 0x12 // Device class - Processing accelerator
+	netClass     = 0x02   // Device class - Network controller.	 ref: https://pci-ids.ucw.cz/read/PD/02 (for Sub-Classes)
+	accClass     = 0x12   // Device class - Processing accelerator
+	rsuDevice    = 0x0b30 // RSU device - only used for programming the FPGA
 )
 
 /*
@@ -47,7 +48,7 @@ Network controller subclasses. ref: https://pci-ids.ucw.cz/read/PD/02
 		07	Infiniband controller
 		08	Fabric controller
 		80	Network controller
-		12  Processing Accelerator
+		12	Processing Accelerator
 */
 
 type cliParams struct {
@@ -196,8 +197,8 @@ func (rm *resourceManager) discoverHostDevices() error {
 			continue
 		}
 
-		// only interested in network class + patch for detecting FEC processing accelerator
-		if devClass == netClass || devClass == fecClass {
+		// only interested in network class and accelerator class
+		if devClass == netClass || devClass == accClass {
 			vendor := device.Vendor
 			vendorName := vendor.Name
 			if len(vendor.Name) > 20 {
@@ -212,55 +213,12 @@ func (rm *resourceManager) discoverHostDevices() error {
 
 			// Network devices
 			if devClass == netClass {
-				// exclude device in-use in host
-				if isDefaultRoute, _ := hasDefaultRoute(device.Address); !isDefaultRoute {
-
-					aPF := utils.IsSriovPF(device.Address)
-					aVF := utils.IsSriovVF(device.Address)
-
-					if aPF || !aVF {
-						// add to linkWatchList
-						rm.addToLinkWatchList(device.Address)
-					}
-
-					if aPF && utils.SriovConfigured(device.Address) {
-						// do not add this device in net device list
-						continue
-					}
-
-					if newDevice, err := resources.NewPciNetDevice(device, rm.rFactory); err == nil {
-						rm.netDeviceList = append(rm.netDeviceList, newDevice)
-					} else {
-						glog.Errorf("discoverDevices() error adding new device: %q", err)
-					}
-
-				}
+				rm.netDevHandler(device)
 			}
 
-			// FEC processing accelerators
-			if devClass == fecClass {
-				//Check if device is a physical function (PF)
-				aPF := utils.IsSriovPF(device.Address)
-
-				//If dev ID = 0b30 it means it is a RSU device and not FEC
-				if device.Product.ID != "0b30" {
-					if aPF && utils.SriovConfigured(device.Address) {
-						// If it is a PF and SRIOV is configured do not add to device list
-						continue
-					}
-
-					if newDevice, err := resources.NewPciNetDevice(device, rm.rFactory); err == nil {
-						if newDevice.GetDriver() == "igb_uio" || newDevice.GetDriver() == "vfio-pci" {
-							rm.netDeviceList = append(rm.netDeviceList, newDevice)
-						} else {
-							glog.Infof("Excluding FEC device %-12s because it is not bound to userspace driver.", device.Address)
-						}
-					} else {
-						glog.Errorf("discoverDevices() error adding new device: %q", err)
-					}
-				} else {
-					glog.Infof("Excluding device %-12s , it is a RSU device.", device.Address)
-				}
+			// Processing accelerators
+			if devClass == accClass {
+				rm.accDevHandler(device)
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


Levovar:
+1 for the functionality :)

cmd/sriovdp/manager.go
```diff
@@ -208,28 +210,57 @@ func (rm *resourceManager) discoverHostDevices() error {
 			}
 			glog.Infof("discoverDevices(): device found: %-12s\t%-12s\t%-20s\t%-40s", device.Address, device.Class.ID, vendorName, productName)
 
-			// exclude device in-use in host
-			if isDefaultRoute, _ := hasDefaultRoute(device.Address); !isDefaultRoute {
+			// Network devices
+			if devClass == netClass {
```
Levovar: consider refactoring the two branches into their own handler functions for increased readability

cmd/sriovdp/manager.go
```diff
@@ -208,28 +210,57 @@ func (rm *resourceManager) discoverHostDevices() error {
...
+
+			// FEC processing accelerators
+			if devClass == fecClass {
+				//Check if device is a physical function (PF)
+				aPF := utils.IsSriovPF(device.Address)
 
-				if newDevice, err := resources.NewPciNetDevice(device, rm.rFactory); err == nil {
-					rm.netDeviceList = append(rm.netDeviceList, newDevice)
+				//If dev ID = 0b30 it means it is a RSU device and not FEC
+				if device.Product.ID != "0b30" {
```
Levovar: use a well named constant instead of hard-coding a hex in the middle of the code

BTW is this code manufacturer -i.e. Intel- specific?

cmd/sriovdp/manager.go
```diff
@@ -32,6 +32,7 @@ import (
 const (
 	socketSuffix = "sock"
 	netClass     = 0x02 // Device class - Network controller.	 ref: https://pci-ids.ucw.cz/read/PD/02 (for Sub-Classes)
+	fecClass     = 0x12 // Device class - Processing accelerator
```
nicolas-chautru: Please rename FecClass to a more generic name as this can appy vof devices exposing more generic baseband processing accelerator. "accClass" should be enough to distinguish from "netClass"

zshi-redhat:
Hi, @damiankopyto thanks for adding new device support!

I have some basic questions regarding enabling this new type of device.
1) Is this a network only accelerator or it can also be used for non-network purposes? From the N3000 data sheet, it looks like so, but I'd like to confirm.
2) Does it require PF be bond to userspace driver in order to perform management and control actions on the device? Is there or will there be any kernel interface developed for device management?
3) Will there be any need to load FPGA image per VF device? How do you handle FPGA image loading?

cmd/sriovdp/manager.go
cbf123: According to [here](https://www.intel.com/content/www/us/en/programmable/documentation/xgz1560360700260.html) it's the PCI device ID for the N3000 FPGA device.

cmd/sriovdp/manager.go
```diff
@@ -46,6 +48,7 @@ Network controller subclasses. ref: https://pci-ids.ucw.cz/read/PD/02
 		07	Infiniband controller
 		08	Fabric controller
 		80	Network controller
+		12	Processing Accelerator
```
cbf123: I think this is actually wrong.  Number 12 is correct for the PCI device class for processing accelerators as specified at https://pci-ids.ucw.cz/read/PD/

However, it does not appear to be a valid option for a *network* controller subclass as described at the URL listed on line 40.

cmd/sriovdp/manager.go
```diff
@@ -208,28 +211,14 @@ func (rm *resourceManager) discoverHostDevices() error {
...
-				} else {
-					glog.Errorf("discoverDevices() error adding new device: %q", err)
-				}
+			// Network devices
+			if devClass == netClass {
+				rm.netDevHandler(device)
+			}
 
+			// Processing accelerators
+			if devClass == accClass {
```
cbf123: It seems a little strange to have two totally separate if statements where only one of them can be true...normally I'd expect to see an "else if" here.

cmd/sriovdp/manager.go
```diff
@@ -290,3 +279,58 @@ type linkWatcher struct {
...
+
+func (rm *resourceManager) accDevHandler(dev *ghw.PCIDevice) {
+	//Check if device is a physical function (PF)
+	aPF := utils.IsSriovPF(dev.Address)
+	devID, err := strconv.ParseInt(dev.Product.ID, 16, 64)
+	if err != nil {
+		glog.Warningf("discoverDevices(): unable to parse device ID for device %+v %q", dev, err)
+		return
+	}
+	//Check if device ID = 0b30, it means it is a RSU (Remote System Update) device and not accelerator
```
cbf123: It seems messy to need to be worrying about vendor-specific PCI device IDs in the middle of this code. Is there nothing else we can look at to determine the class of a device, or to make this more generic so that adding other accelerators would be simpler?

cmd/sriovdp/manager.go
```diff
@@ -31,7 +31,9 @@ import (
 
 const (
 	socketSuffix = "sock"
-	netClass     = 0x02 // Device class - Network controller.	 ref: https://pci-ids.ucw.cz/read/PD/02 (for Sub-Classes)
+	netClass     = 0x02   // Device class - Network controller.	 ref: https://pci-ids.ucw.cz/read/PD/02 (for Sub-Classes)
+	accClass     = 0x12   // Device class - Processing accelerator
+	rsuDevice    = 0x0b30 // RSU device - only used for programming the FPGA
```
cbf123: What does "RSU" mean in this context, and do we need to document that somewhere?

cmd/sriovdp/manager.go
```diff
@@ -290,3 +279,58 @@ type linkWatcher struct {
...
+		if aPF && utils.SriovConfigured(dev.Address) {
+			// If it is a PF and SRIOV is configured do not add to device list
+			return
+		}
+
+		if newDevice, err := resources.NewPciNetDevice(dev, rm.rFactory); err == nil {
+			if newDevice.GetDriver() == "igb_uio" || newDevice.GetDriver() == "vfio-pci" {
+				rm.netDeviceList = append(rm.netDeviceList, newDevice)
+			} else {
+				glog.Infof("Excluding FEC device %-12s because it is not bound to userspace driver.", dev.Address)
```
cbf123: I thought we were dropping the "FEC" terminology?

cmd/sriovdp/manager.go
```diff
@@ -290,3 +279,58 @@ type linkWatcher struct {
...
+	}
+	//Check if device ID = 0b30, it means it is a RSU (Remote System Update) device and not accelerator
+	if devID != rsuDevice {
+		if aPF && utils.SriovConfigured(dev.Address) {
+			// If it is a PF and SRIOV is configured do not add to device list
+			return
+		}
+
+		if newDevice, err := resources.NewPciNetDevice(dev, rm.rFactory); err == nil {
+			if newDevice.GetDriver() == "igb_uio" || newDevice.GetDriver() == "vfio-pci" {
```
cbf123: This looks like more device-specific logic.  Can we make this more generic to support other accelerators in the future?

cmd/sriovdp/manager.go
```diff
@@ -290,3 +279,58 @@ type linkWatcher struct {
...
+				rm.netDeviceList = append(rm.netDeviceList, newDevice)
+			} else {
+				glog.Infof("Excluding FEC device %-12s because it is not bound to userspace driver.", dev.Address)
+			}
+		} else {
+			glog.Errorf("discoverDevices() error adding new device: %q", err)
+		}
+	} else {
```
steven-webster: It might be better to check if devID == rsuDevice { log and return } at line 318.  I'm thinking of the (future) case of any other accelerator classed PCI devices that would not be appropriate to discover here.  It might be useful to have a list of 'blacklisted' device Ids per class that can be looped through rather than individual checks.

ahalim-intel:
We are currently assessing the impact of supporting this devices with this device plugin. What I see from this changes, this was quick and easy 'fix' to add support for this specific fpga device by adding them into the same global 'pciNetDevice' list.
First of all, these devices should be not added in the pciNetDevice, in fact they should added into their own type of pciDevice which  can then have their own set of 'selector's instead the 'selector's targeted for net devices. As some of the netDevice selectors won't be meaningful for other devices(e.g. "pfNames").
If we do agree to add support for fpga devices with this plugin, then there are other ways to do it instead of making changes in pciNetDevice specific code.

zshi-redhat:
Found this link which answers some of my questions above. Paste it here in case anyone is interestedl in understanding detailed configuration needed for N3000 device, https://github.com/open-ness/specs/blob/master/doc/enhanced-platform-awareness/openness-fpga.md

kad:
>  Intel FPGA Device plugin (https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/cmd/fpga_plugin/README.md) does not support such devices as it dose not cover SRIOV devices using DPDK and also differ in the FPGA image that gets programmed. 

Please fix that sentence, it is incorrect. FPGA device plugin supports FPGA IP block of N3000 card in pre-programmed mode. Programming FPGA IP block is missing in the intel-fpga kernel driver for N3000, and as soon it will be added, FPGA plugin will be able to handle it in orchestration programmed mode as well.

ahalim-intel:
Closing this PR since this feature have been implemented by #214

PR #207, opened 2020-02-14
Docs and files for bond cni usage

killianmuldoon:
Added documentation on how to set up bond cni with sriov - as per [comment  ](https://github.com/intel/bond-cni/pull/4#issuecomment-582868917)

### DIFF
--- a/docs/bond/README.md
+++ b/docs/bond/README.md
@@ -1,0 +1,99 @@
+# Bond interfaces with SR-IOV
+
+The below shows a method to create a pod with 2 SR-IOV interfaces bonded into a single bond interface inside a container using [Bond CNI](https://github.com/intel/bond-cni)
+
+This style of configuration can be used to create workloads with high availability - i.e. with two VFs from different PFs connected to the pod. Note that Bond CNI will only work with interfaces utilising the kernel driver. Workloads that leverage userspace drivers on VFs - such as DPDK workloads - are not covered by this guide or by the Bond CNI.
+
+## Prerequisites
+This guide assumes the quick set up of the repo has been followed. That is that Multus, SRIOV CNI and SRIOV Device Plugin have been installed.
+
+Additionally at least two SRIOV Virtual Functions of the type 'intel_sriov_netdevice' should be available on a target node.
+
+## Installing Bond CNI
+In order to get Bond CNI working in your cluster first clone the repo:
+
+`git clone https://github.com/intel/bond-cni`
+
+Next move to the directory and build the binary:
+
+`cd bond-cni && ./build.sh`
+
+As with all CNI directories this binary will need to be placed in the CNI directory on each node in order to be used by the cluster. By default this directory is at `/opt/cni/bin`
+
+For a single node this can simply be copied:
+``cp bin/bond /opt/cni/bin``
+
+For each additional host in the cluster run:
+``scp bin/bond <USER>@<HOST_IP>:/opt/cni/bin``
+
+## Network Attachment Definitions
+Next create the network attachment definitions. The SRIOV CRD is similar to the one in the main repo - but it omits the IP address management as that will be handled by the bond interface. Note that this isn't obligatory but is standard when bonding interfaces.
+
+``kubectl apply -f sriov-net.yaml``
+
+Next create the bonded network attachment definition:
+
+``kubectl apply -f bond-crd.yaml``
+
+### Pod deployment
+With the above network definitions in place the following pod should deploy to a node with 2 SRIOV interfaces available and set up the SRIOV and Bond interfaces correctly.
+```
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test-pod
+  annotations:
+    k8s.v1.cni.cncf.io/networks: '[
+{"name": "sriov-net",
+"interface": "net1"
+},
+{"name": "sriov-net",
+"interface": "net2"
+},
+{"name": "bond-net",
+"interface": "bond0"
+}]'
+spec:  # specification of the pod's contents
+  restartPolicy: Never
+  containers:
+  - name: bond-test
+    image: alpine:latest
+    command:
+      - /bin/sh
+      - "-c"
+      - "sleep 60m"
+    imagePullPolicy: IfNotPresent
+    resources:
+      requests:
+        intel.com/intel_sriov_netdevice: '2'
+      limits:
+        intel.com/intel_sriov_netdevice: '2'
+```
+
+Multus will trigger the above CRDs, sriov-net and bond-net1 in the order in which they're listed. This will result in first the creation of a pod with two SRIOV VFs interfaces named net1 and net2 as specified in the networks annotation. It will then create a bond interface named bond0, using the interfaces named net1 and net2 as specified in the bond network attachment definition.
+
+The pod spec contains explicit naming of each of the interfaces in the networks annotation.
+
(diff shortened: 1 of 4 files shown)


docs/bond/README.md
```diff
@@ -0,0 +1,98 @@
+# Bond interfaces with SR-IOV
+
+The below shows a method to create a pod with 2 SR-IOV interfaces bonded into a single bond interface inside a container using [Bond CNI](https://github.com/intel/bond-cni)
+
+This style of configuration can be used to create workloads with high availability - i.e. with two VFs from different PFs connected to the pod.
+
+## Prerequisites
+This guide assumes the quick set up of the repo has been followed. That is that Multus SRIOV CNI and SRIOV Device Plugin have been installed.
```
zshi-redhat: add comma between Multus and SR-IOV.

docs/bond/README.md
```diff
@@ -0,0 +1,98 @@
...
+  - name: bond-test
+    image: alpine:latest
+    command:
+      - /bin/sh
+      - "-c"
+      - "sleep 60m"
+    imagePullPolicy: IfNotPresent
+    resources:
+      requests:
+        intel.com/intel_sriov_netdevice: '2'
```
zshi-redhat: In a real deployment, this doesn't guarantee two VF resources are from different PFs. We may want to hightlight this as user would assume this is sufficient to create a SR-IOV pod with HA.

docs/bond/README.md
```diff
@@ -0,0 +1,98 @@
...
+With the above network definitions in place the following pod should deploy to a node with 2 SRIOV interfaces available and set up the SRIOV and Bond interfaces correctly.
+``
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test-pod
+  annotations:
+    k8s.v1.cni.cncf.io/networks: '[
+{"name": "sriov-net",
+"interface": "net1"
```
zshi-redhat: Although Multus will name the interface with netx in order, but I think it's good to explicitly list the interface name here which maps to the bond-cni net-attach-def.

docs/bond/README.md
pkdas: sriovdp is aware of VFs from different PFs.   We extended sriov-dp and sriov CNI to implement bonding(active-active and across PFs and any number of VFs). It also required that LACP bonding is created on the host with the correct PFs. There is additional plumbing required to remove a VF when link is up but LACP is down.  The deployment requirement was MLAG with LACP to the TOR.

zshi-redhat: @pkdas I think bonding using LACP mode may be a separate question that worth creating a dedicated issue in bond-cni or device plugin. but thanks for mentioning it here!

The comment I made was specific to the example in this bond doc that uses active-backup bond mode. Since it's possible that VFs allocated from single resource pool may belongs to same PF.

pkdas: Having 2 VFs from the same PF and bond them together does not have much value. If link goes down on the PF,  all VFs will have link down and entire bond is down. There is no benefit having 2 VFs from the same PF  - it does not provide link redundancy and does not provide higher BW (one VF is enough for that). I would suggest that SRIOV-DP to ensure that VFs are from different PF.  I agree  LACP enable/disable is another enhancement

pkdas: My suggestion will be not to expose "2" sriov_net_device to POD. But keep the resource request requirement of sriov_net_deivce to "1" but each sriov_net_device can be comprised with 2(or N) VFs from two (N) different PFs.

pkdas: This would require enhancement to SRIOV CNI which receives a list of PCIE-addresses (VF) for one sriov_net_device request.

zshi-redhat: @pkdas would you mind creating an issue in [bond-cni](https://github.com/intel/bond-cni/issues) so that we can discuss LACP support there?

docs/bond/README.md
killianmuldoon: Added

docs/bond/README.md
killianmuldoon: @pkdas Thanks for the input here. It would be good to move the discussion over to its own issue so we can properly dive in to how to improve this.

killianmuldoon: @zshi-redhat I've added a couple of lines about requiring VFs from different device pools for real world failover. We could also rework this guide to bring in different device pools, but that would need to change configs etc. from the basic guide set-up. I'm happy to do it, but I'd see this guide as introductory with more in depth information on the Bond repo. 

The same goes for the issues with fail-over-mac etc. I think we should put those on the main repo rather than in this guide. 

Let me know what you think.

zshi-redhat: @killianmuldoon it sounds good to me to document that in bond-cni, I think the key is that user can find it somewhere easily when they hit that issue.

docs/bond/README.md
```diff
@@ -0,0 +1,99 @@
+# Bond interfaces with SR-IOV
+
+The below shows a method to create a pod with 2 SR-IOV interfaces bonded into a single bond interface inside a container using [Bond CNI](https://github.com/intel/bond-cni)
+
+This style of configuration can be used to create workloads with high availability - i.e. with two VFs from different PFs connected to the pod. Note that Bond CNI will only work with interfaces utilising the kernel driver. Workloads that leverage userspace drivers on VFs - such as DPDK workloads - are not covered by this guide or by the Bond CNI.
+
+## Prerequisites
+This guide assumes the quick set up of the repo has been followed. That is that Multus, SRIOV CNI and SRIOV Device Plugin have been installed.
+
+Additionally at least two SRIOV Virtual Functions of the type 'intel_sriov-netdevice' should be available on a target node.
```
zshi-redhat: `intel_sriov-netdevice`, change hypen to underscore.

docs/bond/README.md
```diff
@@ -0,0 +1,99 @@
...
+        intel.com/intel_sriov_netdevice: '2'
+      limits:
+        intel.com/intel_sriov_netdevice: '2'
+```
+
+Multus will trigger the above CRDs, sriov-net and bond-net1 in the order in which they're listed. This will result in first the creation of a pod with two SRIOV VFs interfaces named net1 and net2 as specified in the networks annotation. It will then create a bond interface named bond0, using the interfaces named net1 and net2 as specified in the bond network attachment definition.
+
+The pod spec contains explicit naming of each of the interfaces in the networks annotation.
+
+In the above example both VFs used in the bond come from the same VF pool. This does not guarantee that the VFs are from different physical functions, or from different root network cards. In a real world scenario failover would not be guaranteed by the above configuration. Instead separate device pools with specific requests from each would need to be implemented to ensure bonding is performed on different network cards.
```
zshi-redhat: Thanks for adding this!

docs/bond/README.md
```diff
@@ -0,0 +1,99 @@
+# Bond interfaces with SR-IOV
+
+The below shows a method to create a pod with 2 SR-IOV interfaces bonded into a single bond interface inside a container using [Bond CNI](https://github.com/intel/bond-cni)
+
+This style of configuration can be used to create workloads with high availability - i.e. with two VFs from different PFs connected to the pod. Note that Bond CNI will only work with interfaces utilising the kernel driver. Workloads that leverage userspace drivers on VFs - such as DPDK workloads - are not covered by this guide or by the Bond CNI.
```
zshi-redhat: Thanks for adding the note on dpdk bonding!

PR #209, opened 2020-02-24
Replace '--resource-prefix' default value

adrianchiris:
This commit replaces the default value to be vendor agnostic
and updates relevant parts in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,7 +236,7 @@ Usage of ./sriovdp:
   -logtostderr
         log to standard error instead of files
   -resource-prefix string
-        resource name prefix used for K8s extended resource (default "intel.com")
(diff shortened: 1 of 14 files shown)


zshi-redhat:
Most of the example yaml files under `deployment` and `doc` that using default resource prefix may also need to be changed.

adrianchiris:
> Most of the example yaml files under deployment and doc that using default resource prefix may also need to be changed.

will update those as well

adrianchiris:
As discussed in the bi-weekly meeting, this shall be deferred.

Repropose if/when this moves to a vendor neutral repository, possibly with a different URI.

adrianchiris:
We should re-open this once the above occurs

PR #211, opened 2020-03-05
WIP: edited Dockerfiles and shellscript in line with static analysis

killianmuldoon:
Edited docker files and the entrypoint.sh shell script to meet static analysis directions from shellcheck(bash script) and Hadolint(Docker)

### DIFF
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -1,6 +1,6 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.10 AS builder
 
-ADD . /usr/src/sriov-network-device-plugin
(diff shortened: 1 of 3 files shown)


zshi-redhat:
@killianmuldoon if that's commonly static analysis, do you think we could also add entry in Makefile and travis for future changes?

killianmuldoon:
@zshi-redhat I WIPed this so we can have a discussion on what to include and what not in terms of the build pipeline - but I think it might be a good idea to have standardized linting for the additional scripts for sure

ahalim-intel:
@killianmuldoon is this ready to merge?

killianmuldoon:
Plan is now to rework this into two pull requests - one for Docker and one for shell so they can be considered separately.

Issue #212, opened 2020-03-11
Support pool-specific logic


amorenoz:
Currently, the SR-IOV Network Device Plugin works under the assumption that all of the `pciNetDevice`'s information can be obtained from its PCI address, independently from what `ResourcePool` it will end up being added to. Hence, no pool-dependent logic can be implemented.

This makes it difficult to use pool options to select what information from the device is given to the pod. As an example of this, `RdmaSpecs` are extracted (or at least tried) for all PciNetDevices but then only returned if the pool has been configured with `isRdma=true`. This is done by externally appending the RdmaSpecs in the only place where pool configuration is easily available, i.e: (rp *resourcePool) GetDeviceSpecs(deviceIDs []string).  This can be seen as suboptimal and a non-intuitive way of working around the above-mentioned assumption.

## What would you like to be added?

A clear way of adding pool-specific logic, either at runtime or at discovery-time.

### Option 1: Logic at runtime

If runtime logic was prefered, a new function (i.e: `Allocate()`) could be added to `PciNetDevice` and called (with the pool's configuration info) before retrieving the device's information. That way, we would have a unified way of modifying the device specs on a per-pool basis (instead of an ad hoc solution for rdma devices).

Another option would be to add the pool's configuration to the current 'get' functions (`GetDeviceSpecs()`, `GetMounts()`, etc).

### Option 2: Logic at discovery time

If runtime logic is to be avoided to minimize the potential performance impact, we could split the current PciNetDevice into two interfaces. PciNetDevice would continue to server low-level device data, and a new interface (e.g: PoolDevice) would serve the information that the ResourcePool really needs. 
For example:

```
// PciNetDevice provides an interface to get network device specific information. Used for filtering and pool assignment
type PciNetDevice interface {
	GetPFName() string
	GetPfPciAddr() string
	GetVendor() string
	GetDriver() string
	GetDeviceCode() string
	GetPciAddr() string
	GetNetName() string
	IsSriovPF() bool
	GetLinkSpeed() string
	GetLinkType() string
	GetSubClass() string
	GetRdmaSpec() RdmaSpec
	GetVFID() int
	GetDDPProfiles() string
}

// PoolDevice provides an interface to get the information that will be used by the ResourcePool
type PoolDevice interface {
	GetDeviceSpecs() []*pluginapi.DeviceSpec
	GetEnvVal() string
	GetMounts() []*pluginapi.Mount
	GetAPIDevice() *pluginapi.Device
}
```

[...]

PR #213, opened 2020-03-11
Add PoolInfo to implement discovery-time pool-specific logic

amorenoz:
This PR is intended to demonstrate #212 and propose a possible implementation (still work in progress).

Firstly, it splits the current PciNetDevice interface into two separate interfaces:
- PciNetDevice for low level device information (used for filtering)
- PoolDevice for providing the information that the ResourcePool needs (i.e: DeviceSpecs, Mounts, Envs, etc)

That way, InfoProviders could be selected based on ResourcePool information.

This is part of my efforts to cleanly implement vdpa support for the Device Plugin.

It can be observed how, with this preliminary rework, an experimental vdpa support can be implemented with little change in existing DP code:
https://github.com/amorenoz/sriov-network-device-plugin/tree/vdpa_pooldev

### DIFF
--- a/pkg/accelerator/accelPoolInfo.go
+++ b/pkg/accelerator/accelPoolInfo.go
@@ -1,0 +1,40 @@
+// Copyright 2020 Red Hat, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package accelerator
+
+import (
+	"github.com/golang/glog"
+	"github.com/intel/sriov-network-device-plugin/pkg/resources"
+	"github.com/intel/sriov-network-device-plugin/pkg/types"
+)
+
+// accelPoolInfo implements PoolDevice interface for Accelerator Devices
+type accelPoolInfo struct {
+	*resources.PoolInfoImpl
+}
+
+// newAccelPoolInfo create a accelPoolInfo
+func newAccelPoolInfo(accelDev types.AccelDevice, rc *types.ResourceConfig, rf types.ResourceFactory) (*accelPoolInfo, error) {
+
+	glog.Infof("Creating accelPoolInfo for AccelDevice: %+v\n", accelDev.GetPciAddr())
+	infoProvider := rf.GetInfoProvider(accelDev.GetDriver())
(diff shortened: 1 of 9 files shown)


amorenoz:
I'm waiting for some feedback regarding the high level design before starting to work on the tests. Could you ptal @ahalim-intel @zshi-redhat ?

zshi-redhat:
After adding poolInfo, we will have two places to get device related info, one is in discover phase where `ResourceConfig` is not passed in so we only do device filter based on device class, the other is when generating `ResourcePool` where `ResourceConfig` is passed in so advanced filtering is possible, for example, getting rdma spec only if `isRdma` is configured (see https://github.com/intel/sriov-network-device-plugin/pull/226).  

Do we have a clear definition in mind on when will each path be used? or shall we consider combining them to one, for example, use poolInfo for device discover?

amorenoz:
@zshi-redhat You're right, with this PR `isRdma` can be reinterpreted as a "Pool Flag" instead of a "`pciNetDevice` Filter", their correspondent meanings being:
- Pool Flag: If the flag is set, this pool will provide Rdma deviceSpecs. If any of the filtered devices does not have valid Rdma device specifications, the pool creation will fail.
- NetDevice Filter: RdmaSpecs will be retrieved from all NetDevices (it's a `DeviceType` property). If the flag is set, only devices with valid RdmaSpecs will be added to the pool.

In general, I think that all the properties that are not valid across all devices within a `DeviceType` could be interpreted as "Pool Flags" and only be retrieved if the pool flag is specified. With the future introduction of vDPA, both "isVdpa" and "isRdma" would fit in this definition since they are mutually exclusive. (Note: I don't mean to say isRdma devices cannot be exposed via vDPA, which they surely can. But the rdma devices will be handled by the vDPA Framework and only a vhost socket will be exposed via DevicePlugin API).

Regarding the other alternative you propose, IIUC it would entail:
- Having a DeviceProvider per `ResourceConfig` (instead of per `DeviceType`)
- Postpone the call to `DeviceProvider.AddTargetDevices()` to `ResourceManager.initServers()` so that the `ResourceConfig` is available.
- Perform device discovery  + filtering + pool information gathering with the `ResourceConfig` handy

I think your proposal would enable the usecase being tackled here (Same device info, different pool info). We would basically implement all the logic inside NewPciNetDevice(), that would now accept a `ResourceConfig` as parameter. We would still be combining both device-specific and pool-specific information in the same structure (hence I still think the API split proposed is technically valid) but it would make a smaller PR to enable this usecase.

zshi-redhat:
> I think your proposal would enable the usecase being tackled here (Same device info, different pool info). We would basically implement all the logic inside NewPciNetDevice(), that would now accept a `ResourceConfig` as parameter. We would still be combining both device-specific and pool-specific information in the same structure (hence I still think the API split proposed is technically valid) but it would make a smaller PR to enable this usecase.

makes sense to me, let's make the small change.

ahalim-intel:
@amorenoz please close this PR if #231 satisfies this usecase

amorenoz:
> @amorenoz please close this PR if #231 satisfies this usecase

Right. It does. Thanks

Issue #208, opened 2020-02-14
Add support for other SR-IOV devices


ahalim-intel:
# Add support for other SR-IOV devices

| Status        | Proposed                                             |
:-------------- |:---------------------------------------------------- |
| **Author(s)** | Abdul Halim (@ahalim-intel)                          |
| **Updated**   | 2020-02-14                                           |

## Objective

The objective of this proposal is to extend SR-IOV network device plugin to support other SR-IOV capable devices for example FPGA, Accelerators, SmartNICs etc.

## Motivation

The following PRs were added to support other devices with this plugin which is backed by user demands and requirements.

* https://github.com/intel/sriov-network-device-plugin/pull/196
* https://github.com/intel/sriov-network-device-plugin/pull/125

Since this plugin is already capable of creating multiple Kubelet device plugin endpoints and provides a many SR-IOV specific utility functions, it may also possible to extend it to support these devices without breaking its existing functionality.

The problem with the approach taken with the PRs above, the new devices are also treated same as network devices. This may make it difficult to maintain or introduce device specific changes in future for all devices.

## User Benefit

User could leverage the same device plugin to discover and register other SR-IOV Virtual Function(VF)s.

## Design Proposal

Current implementation of the device plugin focuses on physical SR-IOV networking devices in BareMetal only. The device discovery, resource configurations and selectors are very much network device centric.

This proposal suggests the following changes in current implementation so that handling other type of devices can be done by implementing some shared interfaces for generic SR-IOV device, not limited to network devices only.

1. Extract a generic `PciDevice` interface from `PciNetDevice`. Then each device type will extend this as shown in the example below for `PciNetDevice`.

Current `PciNetDevice` interface:

```go
type PciNetDevice interface {
	GetPFName() string
	GetPfPciAddr() string
	GetVendor() string
	GetDriver() string
	GetDeviceCode() string
	GetPciAddr() string
	GetNetName() string
	IsSriovPF() bool
	GetLinkSpeed() string
	GetLinkType() string
	GetSubClass() string
	GetDeviceSpecs() []*pluginapi.DeviceSpec
	GetEnvVal() string
	GetMounts() []*pluginapi.Mount
	GetAPIDevice() *pluginapi.Device
	GetRdmaSpec() RdmaSpec
	GetVFID() int
	GetDDPProfiles() string
}
```

Change this to: 

```go
// PciDevice provides an interface to get device specific information
type PciDevice interface {
	GetVendor() string
[...]

PR #214, opened 2020-03-13
extend plugin to support SR-IOV fpga devices

ahalim-intel:
This PR is the implementation as discussed in #208.
With this changes it will be easier to add support for other SR-IOV devices with this plugin and it would be possible to have device specific logics and configuration per resource pool etc.

In subsequent commit the documentation will be updated accordingly.


ahalim-intel:
cc @amorenoz @adrianchiris @damiankopyto @krishnajs @Levovar @pliurh @przemeklal 
Please review and leave comments. 
Thanks!

pkg/netdevice/netResourcePool.go
```diff
@@ -0,0 +1,70 @@
...
+// NewNetResourcePool returns an instance of resourcePool
+func NewNetResourcePool(rc *types.ResourceConfig, apiDevices map[string]*pluginapi.Device, devicePool map[string]types.PciDevice) types.ResourcePool {
+	rp := resources.NewResourcePool(rc, apiDevices, devicePool)
+	return &netResourcePool{
+		ResourcePoolImpl: rp,
+	}
+}
+
+// Overrides GetDeviceSpecs
+func (rp *netResourcePool) GetDeviceSpecs(deviceIDs []string) []*pluginapi.DeviceSpec {
```
amorenoz: This is a good way of also allowing Pool-Specific logic so it would help solve the problems described in #212 
However, only runtime logic is enabled. If discovery-time logic is desired, the logic would have to move to `NewNetResourcePool` and it would imply rewriting the already populated `deviceSpecs`, `mount` and `env`. So, I think that the separation of PoolDevice and PciDevice proposed in #213 would still be useful.

pkg/netdevice/netDeviceProvider.go
```diff
@@ -0,0 +1,121 @@
...
+func (np *netDeviceProvider) AddTargetDevices(devices []*ghw.PCIDevice, deviceCode int) error {
+
+	for _, device := range devices {
+		devClass, err := strconv.ParseInt(device.Class.ID, 16, 64)
+		if err != nil {
+			glog.Warningf("discoverDevices(): unable to parse device class for device %+v %q", device, err)
+			continue
+		}
+
+		if devClass == int64(deviceCode) {
```
adrianchiris: not sure if there is a use case for FPGA but filtering can be done according to PCI class and subclass.
as PCI spec allows further partitioning of the PCI class into subclasses.

see: https://pcisig.com/sites/default/files/files/PCI_Code-ID_r_1_11__v24_Jan_2019.pdf

e.g for network devices:
The PCI class is 0x02
A subclass of 0x00 represents Ethernet controller and 0x07 represents an Infiniband controllers

Anyway, I reckon that for the time being, using just the class is fine.

pkg/types/types.go
```diff
@@ -30,19 +33,60 @@ const (
...
-		PfNames     []string `json:"pfNames,omitempty"`
-		LinkTypes   []string `json:"linkTypes,omitempty"`
-		DDPProfiles []string `json:"ddpProfiles,omitempty"`
-	} `json:"selectors,omitempty"` // Whether devices have SRIOV virtual function capabilities or not
+	ResourcePrefix string           `json:"resourcePrefix,omitempty"` // optional resource prefix that will ovewrite global prefix specified in cli params
+	ResourceName   string           `json:"resourceName"`             // the resource name will be added with resource prefix in K8s api
+	IsRdma         bool             // the resource support rdma
+	DeviceType     DeviceType       `json:"deviceType,omitempty"`
+	Selectors      *json.RawMessage `json:"selectors,omitempty"`
+	DeviceFilter   DeviceFilter
```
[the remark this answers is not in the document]
ahalim-intel: I wanted to do the same but the problem with that is the `DeviceFilter` will vary per resource configuration but the `DeviceProvider` will differ per "deviceType" only. We may have more than one resource configuration for same `DeviceProvider` thus the actual data in `DeviceFilter` object could be different but the data in `DeviceProvider` remain the same.

Also, the initialization of `DeviceFilter` and `DeviceProvider` are happening at different stages. The `DeviceFilter` can only be initialized during unmarsalling user configs and `DeviceProvider` during device discovery. That's why I thought it's better to keep them in separate interfaces and implementations. But the implementation within the same package of a device type.

pkg/accelerator/accelResourcePool.go
```diff
@@ -0,0 +1,60 @@
...
+func (rp *accelResourcePool) GetDeviceSpecs(deviceIDs []string) []*pluginapi.DeviceSpec {
+	glog.Infof("GetDeviceSpecs(): for devices: %v", deviceIDs)
+	devSpecs := make([]*pluginapi.DeviceSpec, 0)
+
+	devicePool := rp.GetDevicePool()
+
+	// Add vfio group specific devices
+	for _, id := range deviceIDs {
+		if dev, ok := devicePool[id]; ok {
+			netDev := dev.(types.AccelDevice) // convert generic PciDevice to PciNetDevice
```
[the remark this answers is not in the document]
ahalim-intel: will fix this

pkg/types/types.go
```diff
@@ -84,38 +130,60 @@ type ResourcePool interface {
...
+	GetRdmaSpec() RdmaSpec
 	GetDDPProfiles() string
 }
 
+// AccelDevice extends PciDevice interface
+type AccelDevice interface {
+	PciDevice
+}
+
 // DeviceInfoProvider is an interface to get Device Plugin API specific device information
```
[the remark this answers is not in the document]
ahalim-intel: Naming things remain the hardest problems of all :) 
Thanks for the suggestions. I couldn't come up with anything better. I didn't want to rename existing interfaces. It's better to rename the new interface. I think `DeviceTypeProvider ` seems more appropriate.

pkg/accelerator/accelResourcePool.go
```diff
@@ -0,0 +1,60 @@
...
+// NewAccelResourcePool returns an instance of resourcePool
+func NewAccelResourcePool(rc *types.ResourceConfig, apiDevices map[string]*pluginapi.Device, devicePool map[string]types.PciDevice) types.ResourcePool {
+	rp := resources.NewResourcePool(rc, apiDevices, devicePool)
+	return &accelResourcePool{
+		ResourcePoolImpl: rp,
+	}
+}
+
+// Overrides GetDeviceSpecs
+func (rp *accelResourcePool) GetDeviceSpecs(deviceIDs []string) []*pluginapi.DeviceSpec {
```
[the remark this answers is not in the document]
ahalim-intel: Yes, we can. I have added this so that we know we can modify this in future if needed per deviceType.

pkg/types/types.go
```diff
@@ -84,38 +130,60 @@ type ResourcePool interface {
...
 	GetVendor() string
 	GetDriver() string
 	GetDeviceCode() string
 	GetPciAddr() string
-	GetNetName() string
 	IsSriovPF() bool
-	GetLinkSpeed() string
-	GetLinkType() string
 	GetSubClass() string
 	GetDeviceSpecs() []*pluginapi.DeviceSpec
```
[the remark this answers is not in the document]
ahalim-intel: We could tackle this with some additional comments instead of renaming. Renaming this will affect about 55 files.

pkg/resources/pciDevice.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	return ""
+}
+
+// NewPciDevice returns an instance of PciDevice interface
+func NewPciDevice(dev *ghw.PCIDevice, rFactory types.ResourceFactory) (types.PciDevice, error) {
+	//func NewPciDevice(pciDevice *ghw.PCIDevice, rFactory types.ResourceFactory) (types.PciNDevice, error) {
+	// populate all fields in pciNetDevice here
+
+	// 			1. get PF details, add PF info in its pciNetDevice struct
```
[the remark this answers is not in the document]
ahalim-intel: will update this comment.

pkg/netdevice/netResourcePool.go
```diff
@@ -0,0 +1,70 @@
...
+	glog.Infof("GetDeviceSpecs(): for devices: %v", deviceIDs)
+	devSpecs := make([]*pluginapi.DeviceSpec, 0)
+
+	devicePool := rp.GetDevicePool()
+	config := rp.GetConfig()
+
+	// Add vfio group specific devices
```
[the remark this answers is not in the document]
ahalim-intel: removing vfio from here as well and replacing with the following:
// Add device driver specific and rdma specific devices

pkg/types/types.go
```diff
@@ -30,19 +33,40 @@ const (
...
-		Vendors     []string `json:"vendors,omitempty"`
-		Devices     []string `json:"devices,omitempty"`
-		Drivers     []string `json:"drivers,omitempty"`
-		PfNames     []string `json:"pfNames,omitempty"`
-		LinkTypes   []string `json:"linkTypes,omitempty"`
-		DDPProfiles []string `json:"ddpProfiles,omitempty"`
-	} `json:"selectors,omitempty"` // Whether devices have SRIOV virtual function capabilities or not
+	ResourcePrefix string           `json:"resourcePrefix,omitempty"` // optional resource prefix that will ovewrite global prefix specified in cli params
+	ResourceName   string           `json:"resourceName"`             // the resource name will be added with resource prefix in K8s api
+	IsRdma         bool             // the resource support rdma
```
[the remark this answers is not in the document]
ahalim-intel: We could move `isRdma` into `NetDeviceSelectors` but that will result in compatibility issue with previous versions. Are we OK with this approach?

pkg/netdevice/netDeviceProvider.go
```diff
@@ -0,0 +1,121 @@
...
+
+			}
+		}
+	}
+	return nil
+}
+
+// hasDefaultRoute returns true if PCI network device is default route interface
+func hasDefaultRoute(pciAddr string) (bool, error) {
+
```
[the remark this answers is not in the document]
ahalim-intel: removed

pkg/netdevice/netDeviceProvider.go
```diff
@@ -0,0 +1,121 @@
...
+				if aPF && utils.SriovConfigured(device.Address) {
+					// do not add this device in net device list
+					continue
+				}
+
+				if newDevice, err := NewPciNetDevice(device, np.rFactory); err == nil {
+					np.deviceList = append(np.deviceList, newDevice)
+				} else {
+					glog.Errorf("discoverDevices() error adding new device: %q", err)
```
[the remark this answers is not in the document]
ahalim-intel: I thought I've updated this. let me check again.

pkg/types/types.go
```diff
@@ -30,19 +33,40 @@ const (
...
+	DeviceType     DeviceType       `json:"deviceType,omitempty"`
+	Selectors      *json.RawMessage `json:"selectors,omitempty"`
+	DeviceFilter   DeviceFilter
+}
+
+// NetDeviceSelectors contains network device related selectors fields
+type NetDeviceSelectors struct {
+	Vendors     []string `json:"vendors,omitempty"`
+	Devices     []string `json:"devices,omitempty"`
+	Drivers     []string `json:"drivers,omitempty"`
```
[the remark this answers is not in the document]
ahalim-intel: Very good suggestion. Let me try this and see if there is no issue during unmarshalling with embedded struct.

pkg/netdevice/netResourcePool.go
ahalim-intel: The discovery logic per device/drivers is handled with "DeviceInfoProvider" interface, that we implemented for uio and vfio devices. The same interface could be implemented for other device use cases as well (e.g. vDPA devices, which is still a netdevice but needs additional information to be sent). So, if we could determine whether from sysfs or from the driver that this device will support vDPA then we could have vDPA specifc "DeviceInfoProvider" that will handle adding additional device files for DeviceSpecs.

pkg/netdevice/netDeviceProvider.go
ahalim-intel: this can be handled with introducing new selector for device "subclass". But I think the existing "device" selector(which is for device hex code) can already target a very specific device. Not sure if this is needed.

pkg/types/types.go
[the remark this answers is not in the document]
adrianchiris: I think that in the issue discussion backward compatibility was desired.
it eventually depends on who deploys this plugin and how.

e.g for Openshift they would need to change their SR-IOV network Operator code and whether or not there is coupling between the Operator's version and the device plugin image version (i.e older Operator pull in a newer sriov device plugin image).

@zshi-redhat thoughts on the matter?

pkg/netdevice/netResourcePool.go
amorenoz: Unfortunately, AFAIK there is no way of getting that information from sysfs or the driver in a vendor-agnostic way for now.

pkg/types/types.go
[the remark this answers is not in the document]
zshi-redhat: The reason why`DeviceFilter` varies per resource configuration is because it contains unmarshalled selectors from resource config, the rest fields in `DeviceFilter` such as `rFactory` is already contained in `DeviceProvider` structure and `isRdma` is referenced in each `ResourcePool` via unmarshalled data in `rm.configList`. In my opinion, we don't have to create a separate structure only for saving the unmarshalled selectors which looks a little over engineered.  What I was thinking is to keep using the selector structure we have so that it can be referenced during initialization of resource pool.
`DeviceFilter`, without saving unmarshalled selector config, will not be resource specific, so its function `GetFilteredDevices` can be moved to `DeviceProvider` as it will only receive selector config as argument and each `DeviceProvider` applies same selector logic.

pkg/types/types.go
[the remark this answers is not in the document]
zshi-redhat: I agree it would touch many files. I think that's also the reason it may cause confusion when reading the code because there are so many files using this function that may come from different interfaces.

pkg/types/types.go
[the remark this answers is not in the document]
zshi-redhat: If we all agree that `isRdma` is a network specific flag for all rdma devices( including infiniband ),  we shall move it in network specific structure. I think this will provide an example of adding device type sepcific flag.

Re backward compatibility, I was thinking before merging this PR, we will need to create a new release (branch) so that there will be a pinned version for operator to integrate.

pkg/factory/factory.go
```diff
@@ -0,0 +1,174 @@
...
+		if len(filteredDevice) > 0 {
+			if _, ok := filteredDevice[0].(types.AccelDevice); ok {
+				rPool = accelerator.NewAccelResourcePool(rc, apiDevices, devicePool)
+			} else {
+				rPool = nil
+				err = fmt.Errorf("invalid device list for AcceleratorType")
+			}
+		}
+	default:
+		rPool = resources.NewResourcePool(rc, apiDevices, devicePool)
```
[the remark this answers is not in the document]
zshi-redhat: any thoughts?

pkg/types/types.go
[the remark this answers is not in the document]
ahalim-intel: I sorry, I am not quite following it. Can please provide some code examples of your suggested solution?

pkg/factory/factory.go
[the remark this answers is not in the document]
ahalim-intel: Sorry, I've missed this comment before.
The NewResourcePool will return a generic pool implementation, we expect not to reach here. I added this here just in case in future we have other devices that may work with a generic pool. 
For AcceleratorType, we could have used the generic pool but I think it's good idea to use it's own implementation for future extension.

We could change 'default' to return nil and error with not supporting device.

pkg/types/types.go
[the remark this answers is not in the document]
zshi-redhat: I wrote an example [here](https://github.com/zshi-redhat/sriov-network-device-plugin/commit/40f1cd31f09aba9b06f1ac149ad9cce905e2244c) and applied on top of this PR.

The idea is to move `GetFilteredDevices` in `DeviceProvider`:
```
type DeviceProvider interface {
	AddTargetDevices([]*ghw.PCIDevice, int) error
	GetDevices() []PciDevice
+	GetFilteredDevices(*ResourceConfig) []PciDevice
}
```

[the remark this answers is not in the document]
ahalim-intel: Thanks for showing the code examples. Now it's clear!
The whole idea of moving the `Selectors` struct out from the `ResourceConfig` so that we don't have to use (or limited by) same set of selectors for all device types. Having a static struct won't allow us to have different set of selectors from using different struct. Instead we have introduced an interface instead then unmarshal the selectors using customer struct in 2nd level unmarshalling.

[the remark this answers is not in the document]
zshi-redhat: I think It won't block deivce types from using device type specific selectors by keeping `Selectors` structure in `ResourceConfig` , instead it allows sharing the same selector for all types of device types. If a selector is not meaningful for a particular device type, then device provider won't handle this selector even it's configured. In enssence, one device type will only use one combination of various selectors, not two or more. The selectors are used to get filtered device only when creating resource pool during initialization, it's not necessary to carry them on to later phase where DP API is called.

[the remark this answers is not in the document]
ahalim-intel: We wanted to separate out selectors for each device types so that we don't necessarily modify structure meant for one device type for another(PfNmaes, isRdma would be some examples).
The resource config is part of resource pool object regardless we use different struct per device type or the same struct as it is now. But using the approach you suggested we will also have selectors fields that may not applicable to other pool at all.
In addition to that, we will also lose some configuration validation per device type during selector unmarshalling from user input.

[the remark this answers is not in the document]
zshi-redhat: > We wanted to separate out selectors for each device types so that we don't necessarily modify structure meant for one device type for another(PfNmaes, isRdma would be some examples).

We have a set of selectors defined in single structure, but it's not necessary one devcie type will use all of them, I think the device type field will be sufficient for device provider to intellegently choose which selectors to use and implememt its own selector logic as ultimately filtering device is a per device type attribute, not per resource pool. I think the discussion has drifted a little bit to focus on how selector is inputted. my original asking is more to move deviceFitler to deviceProvider. but I agree the selector structure will need further discussion and design.

> The resource config is part of resource pool object regardless we use different struct per device type or the same struct as it is now. But using the approach you suggested we will also have selectors fields that may not applicable to other pool at all.

Right, if resource pool already includes user input resourceConfig that has a full copy of selectors, it may not be needed to save another unmarshalled copy as the unmarshalling can be done at any time later. But if the goal of executing unmarshall is to examine and validate whether it has the correct format, then we can still keep the unmarshalling function, but not save the unmarshalled data.

> In addition to that, we will also lose some configuration validation per device type during selector unmarshalling from user input.

I don't see any particular validation in current code other than unmarshall raw data to a structure (if there are selectors in raw data that doesn't belong to a device type, the unmarshalling won't capture it, rather it's in device filtering function that the unexpected selector will be ignored )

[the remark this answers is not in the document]
ahalim-intel: > We have a set of selectors defined in single structure, but it's not necessary one devcie type will use all of them, I think the device type field will be sufficient for device provider to intellegently choose which selectors to use and implememt its own selector logic as ultimately filtering device is a per device type attribute, not per resource pool. I think the discussion has drifted a little bit to focus on how selector is inputted. my original asking is more to move deviceFitler to deviceProvider. but I agree the selector structure will need further discussion and design.
> 

The deviceFilter works with the selector struct and it has one to one relationship with a resource pool. Whereas deviceProvider has one to many relationship with resource pool. Now, it is possible for deviceProvider also implement the DeviceFilter interface iff we use same selectors struct for all devices. This is where I'm trying to getting at - We do not want to fall into same problems using same `Selectors` struct for all devices. We want to have different set of selectors(and possibly other fields as well) per device. And modifying a selector for one device should not affect another device. Yes, in the `GetFilteredDevices()` method we could pick and choose the only fields a device maybe interested in from a common selectors struct. But I still think it's a good idea to separate out the device specific selectors and use dependency using a interface(`DeviceFilter` interface) instead of a struct(`Selectors`).

> 
> Right, if resource pool already includes user input resourceConfig that has a full copy of selectors, it may not be needed to save another unmarshalled copy as the unmarshalling can be done at any time later. But if the goal of executing unmarshall is to examine and validate whether it has the correct format, then we can still keep the unmarshalling function, but not save the unmarshalled data.
> 

With new implementation the resourceConfig no longer holds the unmarshalled selectors but a reference to DeviceFilter interface that holds the actual selectors object. Only difference the new approach to the old is we one pointer to raw bytes of selectors as placeholder to for late unmarshalling. Wherein in existing approach the resourceConfig actually holds the full Selector struct.
 
> I don't see any particular validation in current code other than unmarshall raw data to a structure (if there are selectors in raw data that doesn't belong to a device type, the unmarshalling won't capture it, rather it's in device filtering function that the unexpected selector will be ignored )

It's not obvious now, but if we had field in selector that is required(without 'omitempty' json annotation) the unmarshalling could capture this.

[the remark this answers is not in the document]
ahalim-intel: Having said that, if you think it complicates things rather than adding any real benefits  then we can keep the `ResourceConfig` struct unchanged and use the common `Selectors` for all devices.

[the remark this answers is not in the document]
zshi-redhat: > With new implementation the resourceConfig no longer holds the unmarshalled selectors but a reference to DeviceFilter interface that holds the actual selectors object. Only difference the new approach to the old is we one pointer to raw bytes of selectors as placeholder to for late unmarshalling. Wherein in existing approach the resourceConfig actually holds the full Selector struct.

I agree that selectors may be different per device type. I guess the argument is do we need a separate `DeviceFilter` to hold the unmarshalled data, just for use when creating resource pool? If there is a case that we were to provide different selecting logics within one device type, I'd think it makes sense. But if we were to use one selecting logic for all devices of the same device type, it might not be necessary.

> 
> > I don't see any particular validation in current code other than unmarshall raw data to a structure (if there are selectors in raw data that doesn't belong to a device type, the unmarshalling won't capture it, rather it's in device filtering function that the unexpected selector will be ignored )
> 
> It's not obvious now, but if we had field in selector that is required(without 'omitempty' json annotation) the unmarshalling could capture this.

Do we have an example of such selector that can not be omitted?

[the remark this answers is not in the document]
ahalim-intel: > I agree that selectors may be different per device type. I guess the argument is do we need a separate `DeviceFilter` to hold the unmarshalled data, just for use when creating resource pool? If there is a case that we were to provide different selecting logics within one device type, I'd think it makes sense. But if we were to use one selecting logic for all devices of the same device type, it might not be necessary.
> 

I think yes, we need separate `DeviceFilter` to hold the unmarshalled data as the selector values will differ per config. And the selector logic also can be different per device type but not per config though. The DeviceProvider just holds a specific global device list based on types. It doesn't know anything about resource configurations, that is the reason it cannot implement `DeviceFilter` unless we provide config+selectors to it. The actual issue I see is the resource pool itself needs no know about the resource config and the selectors(maybe, only if we move isRdma inside a selector). 
 
> 
> Do we have an example of such selector that can not be omitted?

I don't have an explicit example in hand, but think about what if some device selector requires a flag that must be specified for that device type only but not for others? In that case a common selector won't facilitate this. 
In the original proposal the main idea was to separate out the selectors from main config so that we could have device specific selector struct.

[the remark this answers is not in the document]
ahalim-intel: @zshi-redhat PTAL the recent changes.
I have moved the GetFilteredDevices into DeviceProvider but retained the Device specific selectors structs. The GetFilteredDevices() now required to get the actual selectors for resource config which we can get from a type assertion. The same techniques is used to get selectors object in the resource Pool. So, we can now get "IsRdma" from the selectors instead of the main config.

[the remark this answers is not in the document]
zshi-redhat: @ahalim-intel thanks for the update! The change looks good to me. Now we allows adding device type specific selectors, but keep the same selecting logic per device type.

pkg/types/types.go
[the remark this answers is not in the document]
zshi-redhat: This was addressed in latest commits that moving `isRdma` flag to `NetDeviceSelectors`.

Note: the change break backward compability, doc update is needed.

README.md
```diff
@@ -184,11 +184,11 @@ This plugin creates device plugin endpoints based on the configurations given in
         {
             "resourceName": "mlnx_sriov_rdma",
             "resourcePrefix": "mellanox.com",
-            "isRdma": true,
             "selectors": {
                 "vendors": ["15b3"],
                 "devices": ["1018"],
-                "drivers": ["mlx5_ib"]
+                "drivers": ["mlx5_ib"],
+                "isRdma": true,
```
zshi-redhat: nit: comma is not needed, but it doesn't hurt.

README.md
```diff
@@ -197,25 +197,62 @@ This plugin creates device plugin endpoints based on the configurations given in
...
+|---------------|----------|----------------------------------------------------------------|---------------------------------------------------|--------------------------------------------------------------------------------------|
+| "pfNames"     | N        | VFs from PF matches list of PF names                           | `string` list Default: `null`                     | "pfNames": ["enp2s2f0"] (See follow-up sections for some advance usage of "pfNames") |
+| "linkTypes"   | N        | The link type of the net device associated with the PCI device | `string` list Default: `null`                     | "linkTypes": ["ether"]                                                               |
+| "ddpProfiles" | N        | A map of device selectors                                      | `string` list Default: `null`                     | "ddpProfiles": ["GTPv1-C/U IPv4/IPv6 payload"]                                       |
+| "isRdma"      | N        | Mount RDMA resources                                           | `bool`  values `true` or `false` Default: `false` | "isRdma": `true`                                                                     |
```
zshi-redhat: It might be good to also update example yaml files under /docs and /deployments that are still using `isRdma` as field of `ResourceConfig`.

```
./docs/rdma/configMap.yaml
./docs/dpdk/mlnx-dpdk/configMap.yaml
./docs/dpdk/mlnx-dpdk/configMap.yaml
./deployments/configMap.yaml
```

README.md
```diff
@@ -197,25 +197,62 @@ This plugin creates device plugin endpoints based on the configurations given in
             "selectors": {
                 "linkTypes": ["infiniband"]
             }
+        },
```
zshi-redhat: move `isRdma` inside `selectors` for `infiniband_rdma_netdevs` config.

README.md
ahalim-intel: I will fix this. Thanks

README.md
ahalim-intel: I will update them too. thanks

zshi-redhat:
@ahalim-intel the PR looks in good shape now, do you have plan to make a release/branch before merging it?

ahalim-intel:
> @ahalim-intel the PR looks in good shape now, do you have plan to make a release/branch before merging it?

Thanks @zshi-redhat for your valuation feedback, input and suggestions. Yes, lets discuss new release plans on our next meeting.

README.md
```diff
@@ -184,38 +184,75 @@ This plugin creates device plugin endpoints based on the configurations given in
...
+All device types support following common device selectors.
+
+|   Field   | Required |                Description                |         Type/Defaults          |   Example/Accepted values   |
+|-----------|----------|-------------------------------------------|--------------------------------|-----------------------------|
+| "vendors" | N        | Target device's vendor Hex code as string | `string` list Default: `null`  | "vendors": ["8086"]         |
+| "devices" | N        | Target Devices' device Hex code as string | `string` list Default: `null`  | "devices": ["154c", "10ed"] |
+| "drivers" | N        | Target device driver names as string      | `string` list Default: `null`  | "drivers": ["vfio-pci"]     |
+
+
+#### Extended NetDevice selectors
```
adrianchiris, line 241: nit:
#### Extended resource for Device Type `netDevice`

README.md
```diff
@@ -184,38 +184,75 @@ This plugin creates device plugin endpoints based on the configurations given in
...
+In addition to the common selectors from above table, "netDevice" also supports following selectors.
```
adrianchiris, line 242: nit:
.... from above table, the "netDevice" device type also supports.....

README.md
```diff
@@ -184,38 +184,75 @@ This plugin creates device plugin endpoints based on the configurations given in
...
+| "pfNames"     | N        | VFs from PF matches list of PF names                           | `string` list Default: `null`                     | "pfNames": ["enp2s2f0"] (See follow-up sections for some advance usage of "pfNames") |
+| "linkTypes"   | N        | The link type of the net device associated with the PCI device | `string` list Default: `null`                     | "linkTypes": ["ether"]                                                               |
+| "ddpProfiles" | N        | A map of device selectors                                      | `string` list Default: `null`                     | "ddpProfiles": ["GTPv1-C/U IPv4/IPv6 payload"]                                       |
+| "isRdma"      | N        | Mount RDMA resources                                           | `bool`  values `true` or `false` Default: `false` | "isRdma": `true`                                                                     |
+
+
+[//]: # (The tables above generated using: https://ozh.github.io/ascii-tables/)
+
+#### Extended Accelerator Selectors:
```
adrianchiris, line 254: suggestion:
#### Extended selectors for device type `accelerator`

README.md
```diff
@@ -184,38 +184,75 @@ This plugin creates device plugin endpoints based on the configurations given in
...
-| "selectors"      | N        | A map of device selectors              | Each selector is a map of string list.                                                             | "vendors": ["8086"],"devices": ["154c", "10ed"],"drivers": ["vfio-pci"],"pfNames": ["enp2s2f0"],"linkTypes": ["ether"] |
-| "isRdma"         | N        | Mount RDMA resources                   | `bool` - boolean value true or false                                                               | "isRdma": true                                                                                                         |
 
 Note: "resourceName" must be unique only in the scope of a given prefix, including the one specified globally in the CLI params, e.g. "example.com/10G", "acme.com/10G" and "acme.com/40G" are perfectly valid names.
 
-[//]: # (The table above generated using: https://ozh.github.io/ascii-tables/)
+#### Device selectors
```
adrianchiris, line 228: Can you also add a section describing what deviceType affects ?

pkg/resources/pciDevice.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	vfID, err := utils.GetVFID(pciAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	// 			3. Get Device file info (e.g., uio, vfio specific)
```
adrianchiris, line 64: nit: where is 1 and 2 :)

pkg/factory/factory.go
```diff
@@ -0,0 +1,174 @@
...
+	case types.AcceleratorType:
+		return accelerator.NewAccelDeviceProvider(rf)
+	default:
+		return nil
+	}
+}
+
+// GetDeviceFilter unmarshal the "selector" values from ResourceConfig and returns an instance of DeviceSelector based on
+// DeviceType in the ResourceConfig
+func (rf *resourceFactory) GetDeviceFilter(rc *types.ResourceConfig) (interface{}, error) {
```
adrianchiris, line 155: So providing an unsupported selector will fail silently, that's OK ?
you could add some logic to enforce that.

pkg/types/types.go
```diff
@@ -68,8 +115,11 @@ type ResourceFactory interface {
 	GetResourceServer(ResourcePool) (ResourceServer, error)
 	GetInfoProvider(string) DeviceInfoProvider
 	GetSelector(string, []string) (DeviceSelector, error)
-	GetResourcePool(rc *ResourceConfig, deviceList []PciNetDevice) (ResourcePool, error)
+	GetResourcePool(rc *ResourceConfig, deviceList []PciDevice) (ResourcePool, error)
 	GetRdmaSpec(string) RdmaSpec
+	GetDeviceProvider(DeviceType) DeviceProvider
+	// GetDeviceFilter(*ResourceConfig) (DeviceFilter, error)
+	GetDeviceFilter(*ResourceConfig) (interface{}, error)
```
adrianchiris, line 122: Seems like an interface definition is missing here. i dont understand from the definition here what should GetDeviceFilter return value represents

looking at the code i get that its an opaque to carry the parsed selector to the device provider.

pkg/accelerator/accelDeviceProvider.go
```diff
@@ -0,0 +1,117 @@
...
+	"strconv"
+
+	"github.com/golang/glog"
+	"github.com/intel/sriov-network-device-plugin/pkg/types"
+	"github.com/jaypipes/ghw"
+)
+
+type accelDeviceProvider struct {
+	deviceList []types.AccelDevice
+	rFactory   types.ResourceFactory
```
adrianchiris, line 28: not related to the commit, but the factory could really be a singleton instead of referencing to it all over the place as
it creates instances of interfaces used in all layers of the code, requiring to store it and pass it down.

pkg/accelerator/accelDeviceProvider.go
```diff
@@ -0,0 +1,117 @@
...
+func NewAccelDeviceProvider(rf types.ResourceFactory) types.DeviceProvider {
+	return &accelDeviceProvider{
+		rFactory:   rf,
+		deviceList: make([]types.AccelDevice, 0),
+	}
+}
+
+func (ap *accelDeviceProvider) GetDevices() []types.PciDevice {
+	newPciDevices := make([]types.PciDevice, len(ap.deviceList))
+	for i := range ap.deviceList {
```
adrianchiris, line 41: you could just:
```
return copy(newPciDevice, ap.deviceList)
```

pkg/types/types.go
```diff
@@ -30,19 +33,63 @@ const (
...
+02	FDDI network controller
+03	ATM network controller
+04	ISDN controller
+05	WorldFip controller
+06	PICMG controller
+07	Infiniband controller
+08	Fabric controller
+80	Network controller
+*/
+var SupportedDevices = map[DeviceType]int{
```
adrianchiris, line 60: I dont really like the coupling here between the device PCI class code and the device type.

what if in the future, we would like to add a device type which would handle the same PCI class but different attributes ?
Or a single device type (device Provider) handling several PCI classes ?

pkg/types/types.go
```diff
@@ -84,28 +134,46 @@ type ResourcePool interface {
 	GetMounts(deviceIDs []string) []*pluginapi.Mount
 }
 
-// PciNetDevice provides an interface to get device specific information
-type PciNetDevice interface {
-	GetPFName() string
+// DeviceProvider provides interface for device discovery
+type DeviceProvider interface {
```
adrianchiris, line 138: A device provider could report the device type it handles instead of having the device type  to provider knowledge in sriovdp manager

pkg/accelerator/accelDeviceProvider.go
```diff
@@ -0,0 +1,117 @@
...
+		newPciDevices[i] = ap.deviceList[i]
+	}
+	return newPciDevices
+}
+
+func (ap *accelDeviceProvider) AddTargetDevices(devices []*ghw.PCIDevice, deviceCode int) error {
```
adrianchiris, line 47: deviceCode name is a bit vague, can you elaborate what it represents ? (perhaps add some documentation in the corresponding interface)

pkg/factory/factory.go
```diff
@@ -0,0 +1,174 @@
...
+
+	var rPool types.ResourcePool
+	var err error
+	switch rc.DeviceType {
+	case types.NetDeviceType:
+		if len(filteredDevice) > 0 {
+			if _, ok := filteredDevice[0].(types.PciNetDevice); ok {
+				rPool = netdevice.NewNetResourcePool(rc, apiDevices, devicePool)
+			} else {
+				rPool = nil
```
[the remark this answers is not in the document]
adrianchiris: yes, explicit initialization not needed as an interface zero value is nil

pkg/factory/factory.go
```diff
@@ -0,0 +1,174 @@
...
+				err = fmt.Errorf("invalid device list for NetDeviceType")
+			}
+		}
+	case types.AcceleratorType:
+		if len(filteredDevice) > 0 {
+			if _, ok := filteredDevice[0].(types.AccelDevice); ok {
+				rPool = accelerator.NewAccelResourcePool(rc, apiDevices, devicePool)
+			} else {
+				rPool = nil
```
adrianchiris, line 127: same here

pkg/netdevice/netDeviceProvider.go
```diff
@@ -0,0 +1,194 @@
...
+type netDeviceProvider struct {
+	deviceList []types.PciNetDevice
+	rFactory   types.ResourceFactory
+}
+
+// NewNetDeviceProvider DeviceProvider implementation from netDeviceProvider instance
+func NewNetDeviceProvider(rf types.ResourceFactory) types.DeviceProvider {
+	return &netDeviceProvider{
+		rFactory:   rf,
+		deviceList: make([]types.PciNetDevice, 0),
```
adrianchiris, line 37: The zero len is redundant imo as no capacity is provided

pkg/netdevice/netDeviceProvider.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+}
+
+func (np *netDeviceProvider) GetDevices() []types.PciDevice {
+	newPciDevices := make([]types.PciDevice, len(np.deviceList))
+	for i := range np.deviceList {
```
adrianchiris, line 43: same as my comment in in the accl, you could just use `copy()` builtin function

README.md
ahalim-intel: There is a description of deviceType in the table. What would you like to add?

pkg/resources/pciDevice.go
ahalim-intel: Old comments, will fix this.

pkg/factory/factory.go
ahalim-intel: For invalid deviceType it will not fail silently. As you can see, errors are propagated from here, manager will not to save this config for creating resource pool and will be logged.

pkg/types/types.go
ahalim-intel: Well, interface{} is like a void pointer in C to which you can return any struct. This makes it generic method to return different selector struct. I will remove the commented line above(old definition)

pkg/accelerator/accelDeviceProvider.go
ahalim-intel: factory is already implemented as singleton. It doesn't make any difference really whether we use the constructor everywhere to get the instance or pass it as an argument. Its one single instance everywhere.

pkg/accelerator/accelDeviceProvider.go
ahalim-intel: You cannot use copy with slice of different types.

pkg/accelerator/accelDeviceProvider.go
adrianchiris: yah, it just needlessly holds an extra attribute ...

pkg/types/types.go
adrianchiris: Perhaps explain as well what GetDeviceFilter is expected to do/return.

pkg/factory/factory.go
adrianchiris: so it silently ignores unsupported selectors of a certain *valid* device type.
i dont understand your comment

pkg/types/types.go
ahalim-intel: Well for now it make sense to use the device class code to differentiate different device types. Same class different attributes - the we that device specific selectors to handle additional logic etc.
I don't know what better way to determine different deviceType. You have any recommendation?

README.md
adrianchiris: essentially what deviceType affects e.g  the types of selectors which are used.
there is some jump in logic imo straight to selectors in the README

pkg/accelerator/accelDeviceProvider.go
adrianchiris: ah missed that 👍

pkg/types/types.go
ahalim-intel: I'm not following you on this. Can you please explain little bit more?

pkg/accelerator/accelDeviceProvider.go
ahalim-intel: comment added in the interface.

pkg/factory/factory.go
ahalim-intel: agree, will remove this.

pkg/netdevice/netDeviceProvider.go
ahalim-intel: Cannot initialize a slice with make without giving its length.
Try here: https://play.golang.org/p/ae-ktK6cOLK

pkg/accelerator/accelDeviceProvider.go
ahalim-intel: I don't understand it. Even if we create an instance of the factory here(instead of getting this as an argument in its constructor NewAccelDeviceProvider() ) we still need an attribute for it's reference, no?

pkg/factory/factory.go
ahalim-intel: for valid device type, this method will try to unmarshal selector configuration. Any error occurred during this will also be returned and subsequently be logged by the manager. I don't see anything silently ignored, am I missing something?

pkg/netdevice/netDeviceProvider.go
adrianchiris: My bad then, no idea why i remembered differently

pkg/accelerator/accelDeviceProvider.go
adrianchiris: What i mean is that every component that requires to call factory will do (for example):
`factory.Factory.<factory method>`
or
`factory.Get().<factory method>`

It would require to modify `ResourceFactory.GetResourceServer` to get the endpoint prefix/suffix and pluginWatch  (so we dont retain parameters in `resourceFactory` struct)

Then what you get in return is that you dont need to store a reference to it (e.g in netDeviceProvider, accelDeviceProvider)  and you don't need to pass it in their New methods, it just simplifies the code imo.

In regards to functionality it would be the same.

adrianchiris: To be clear, thats not something i request to be changed here, its just an observation.
ResourceFactory existed before this change.

README.md
ahalim-intel: I've updated this section with additional explanation. PTAL

pkg/accelerator/accelDeviceProvider.go
ahalim-intel: Ok, maybe some improvement for future.

pkg/factory/factory.go
adrianchiris: So im most likely missing something, what will happen if i provide the following resource config ?
```
 {
            "resourceName": "intel_fpga",
            "deviceType": "accelerator",
            "selectors": {
                    "vendors": ["8086"],
                    "devices": ["0d90"],
                    "pfNames": ["enp0s0f0","enp2s2f1"]
            }
        }
```

will it fail with something resembling "pfNames selector is invalid for deviceType accelerator" ?

ahalim-intel: Ok, with this particular example it will not fail because json unmarshal will correctly fill in vendors and devices fields in the AccelDeviceSelectors struct but will ignore pfNames as there is no matching field for it there. But if we had a type mismatch for example int in "vendor" will result in error and will be captured during unmarshalling. Also, if there were some mandatory fields for AccelDeviceSelectors (without 'omitempty' json tag) and that was not given in the selectors that would be captured as well during unmarshalling.

adrianchiris: Indeed, its provided by unmarshall :)

so going back to my initial comment: 
> So providing an unsupported selector will fail silently, that's OK ?

clarifying unsupported as unsupported by the device type.

you could first unmarshal it into a map[string]interface{} check the keys then unmarshal to the final struct

README.md
adrianchiris: Thanks!

pkg/factory/factory.go
ahalim-intel: I still don't get the "silently failing". Giving arbitrary fields in json will always be ignored if that's what you call silently failing.

adrianchiris: edited, my comment above, i hope its clearer.

ahalim-intel: > Indeed, its provided by unmarshall :)
> 
> Clarifying my initial comment:
> 
> So providing an unsupported selector by the device type will ignore it silently, that's OK ?

I think this is how generally json objects are handle - CNI would be a perfect example. Only cares about the fields that a plugin actually need.

> 
> you could first unmarshal it into a map[string]interface{} check the keys then unmarshal to the final struct if silently failing is not desired

This would be a little overkill and we still will be left with same situation. Say for example, we first unmarshall the selectors into the map[string]interface{} and which will give us something like this: "vendors": interface{} and it's still unknown what is the actual value given for the vendors.  So, if there was some `int` value given instead of `[]string` we still could not capture this error here.

adrianchiris: i see, OK.

ahalim-intel:
Thanks everyone for your reviews, comments and feedback. I think I have addressed pretty much everything. We will go ahead and merge this PR now as some other PRs may depend on this.

amorenoz:
@moshe010 maybe you accidentally edited your comment but I find the original interesting:
> it seem you change the configuration for isRDMA and move into the selectors. What was that reason for that?
I guess the reason for that is that it's DeviceType specific and it got split into the `Selectors` structure to limit the DeviceType-specific unmarshalling to one structure.
@ahalim-intel: do you think it's worth differentiating between "configuration" and "selectors" in the DeviceType-specific struct?

ahalim-intel:
> it seem you change the configuration for isRDMA and move into the selectors. What was that reason for that?

@moshe010 This change was was suggested and then discussed [here](https://github.com/intel/sriov-network-device-plugin/pull/214#discussion_r392893343).

> do you think it's worth differentiating between "configuration" and "selectors" in the DeviceType-specific struct?

@amorenoz I am not sure if this is needed, but worth trying. Then again, imo it's not a good idea to move/change user config parameters very often without any strong reason.  The device specific `Selectors` struct allows us to pass "selectors" + additional configuration too when needed.

PR #218, opened 2020-03-31
fix travis deployment script errors

ahalim-intel:
The deployment scripts weren't parsing properly and resulted in
errors.

This change also uses the REGISTRY_USER variable for image tagging
so that deployement can be tested on local forks with their own
REGISTRY_USER value.

Change-Id: Idf03b5c8c256f63ce01beff5c316cb8c5a786db3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ script:
   - make test
   - make build
   - make
-  - make image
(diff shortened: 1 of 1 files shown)


ahalim-intel:
The deployment was successful on my forked repo: https://travis-ci.org/github/ahalim-intel/sriov-network-device-plugin/builds/669252140

PR #221, opened 2020-04-07
Update mellanox dpdk example files

Mmduh-483:


docs/dpdk/mlnx-dpdk/crd-dpdk-connectx4.yaml
```diff
@@ -6,6 +6,16 @@ metadata:
     k8s.v1.cni.cncf.io/resourceName: mellanox.com/mlnx_connectx4
 spec:
   config: '{
+    "cniVersion": "0.3.1",
     "type": "sriov",
-    "name": "sriov-mlnx_connectx4"
+    "name": "sriov-mlnx-connectx4",
+    "link_state": "enable",
```
moshe010, line 12: I think you can remove the link state

docs/dpdk/mlnx-dpdk/mlnx-pod_testpmd_connectx4.yaml
```diff
@@ -8,20 +8,20 @@ spec:
   containers:
   - name: testpmd
     image: <DPDK testpmd image>
-    imagePullPolicy: Never
+    imagePullPolicy: IfNotPresent
     securityContext:
      capabilities:
-        add: ["IPC_LOCK"]
+        add: ["IPC_LOCK", "NET_RAW", "NET_ADMIN", "SYS_ADMIN"]
```
moshe010, line 14: Remove the sysadmin is for pa. Also the net admin.

moshe010:
@Mmduh-483, can you fix also https://github.com/intel/sriov-network-device-plugin/blob/master/docs/rdma/configMap.yaml to have the prefix.

docs/rdma/configMap.yaml
```diff
@@ -9,6 +9,7 @@ data:
         "resourceList": [
             {
                 "resourceName": "mlnx_rdma",
+                "resourcePrefix": "mellanox.com",
```
moshe010, line 12: remove the mlx5_ib

Mmduh-483:
@ahalim-intel @zshi-redhat Can you please review?

docs/dpdk/mlnx-dpdk/mlnx-pod_testpmd_connectx4.yaml
```diff
@@ -8,20 +8,20 @@ spec:
   containers:
   - name: testpmd
     image: <DPDK testpmd image>
-    imagePullPolicy: Never
+    imagePullPolicy: IfNotPresent
     securityContext:
      capabilities:
-        add: ["IPC_LOCK"]
+        add: ["IPC_LOCK", "NET_RAW"]
```
zshi-redhat, line 14: Why is `NET_RAW` added?

moshe010: Mellanox dpdk driver is beforcated so we need additional capabilities from the kernel see https://github.com/intel/sriov-network-device-plugin/issues/123

zshi-redhat: With crio, `NET_RAW` is one of the default capabilities passed to each container/pod. User doesn't need to explicitly specify it in pod spec.

zshi-redhat: With crio, `NET_RAW` is one of the default capabilities assigned to each container/pod. User doesn't need to explicitly specify it in pod spec.

moshe010: but with docker we need this. do you want us to add comment that is not required for crio?

zshi-redhat: From this doc, it looks like docker keeps `net_raw` as default capability: https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities
Is the doc outdated?

moshe010: I see I remember that it wasn't but I see I am mistaken. @Mmduh-483 please remove the NET_RAW as it is the default.

docs/dpdk/mlnx-dpdk/mlnx-pod_testpmd_connectx5.yaml
```diff
@@ -8,7 +8,7 @@ spec:
   containers:
   - name: testpmd
     image: <DPDK testpmd image>
-    imagePullPolicy: Never
+    imagePullPolicy: IfNotPresent
     securityContext:
      capabilities:
         add: ["IPC_LOCK"]
```
adrianchiris, line 14: Dont we need CAP_NET_ADMIN (e.g for mtu changes) ?
Also i think we should explicitly state here that CAP_NET_RAW is desired if only as declarative statement in case the low-level runtime enables it by default.

moshe010: Currently I didn't see any cni that allow to configure MTU and link state in the container. for example OVN  configure mtu with  the cni. In sriov cni the link state is configurable by the cni. 
I do  think we need to add MTU support to sriov-cni.

adrianchiris: i see, so your take is to go with the minimal set of capabilities for DPDK container.

Mmduh-483:
@zshi-redhat  Can you please re-review?

adrianchiris:
Overall i think this looks Good.

PR #224, opened 2020-04-19
Add small comment for the device type list

SchSeba:

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -45,6 +45,10 @@ const (
 
 // SupportedDevices is map of 'device identifier as string' to 'device class hexcode as int'
 /*
+Supported PCI Device Classes. ref: https://pci-ids.ucw.cz/read/PD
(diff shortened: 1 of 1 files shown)


pkg/types/types.go
```diff
@@ -56,6 +56,7 @@ Network controller subclasses. ref: https://pci-ids.ucw.cz/read/PD/02
 07	Infiniband controller
 08	Fabric controller
 80	Network controller
+12	Processing Accelerator
```
zshi-redhat, line 59: This list above is about network subclasses, to make it cosistent, maybe we should either of below:
1) list subclasses for `Processing Accelerator`
2) list the only `PCI Device Classes` for `Network controller` and `Processing Accelerator`

SchSeba: Can you please take another look please?

zshi-redhat:
/lgtm
@ahalim-intel PTAL

PR #225, opened 2020-04-19
Switch to go mod

SchSeba:
This commmit also update the kubernetes version.


ahalim-intel:
Hi @SchSeba thanks for this PR.,
I have following two questions, maybe we need input from other contributors.

1. Are we keeping the vendor/ and glide files for older version of Go? If so, then we may need some documentation for developer(in separate PR) when updating dependencies, we need to update both(for now, until we stop supporting vendoring altogether)

2. What are the implication of setting device plugin API to "k8s.io/kubelet v0.18.1"? I'm concerned of compatibility with older version of K8s(v1.13+)

SchSeba:
Hi @ahalim-intel thanks for the comment here is my replay.

> Are we keeping the vendor/ and glide files for older version of Go? If so, then we may need some documentation for developer(in separate PR) when updating dependencies, we need to update both(for now, until we stop supporting vendoring altogether)

I remove the glide files but keep the vendor folder.
I also add a new entry in the makefile `deps-update` that will run `go mod tidy` and also `go mod vendor` to update the vendor folder.
Not sure if this required any documentation change I search for glide in the documentation and wasn't able to find anything.

> What are the implication of setting device plugin API to "k8s.io/kubelet v0.18.1"? I'm concerned of compatibility with older version of K8s(v1.13+)

This will continue to support the same versions as before because we are still using the `v1beta1` API version.

SchSeba:
@ahalim-intel can you please take another look on this PR?

Makefile
```diff
@@ -140,14 +139,9 @@ fmt: ; $(info  running gofmt...) @ ## Run gofmt on all source files
...
-
-glide.lock: glide.yaml | $(BASE) ; $(info  updating dependencies...)
-	$Q cd $(BASE) && $(GLIDE) update -v
-	@touch $@
-vendor: glide.lock | $(BASE) ; $(info  retrieving dependencies...)
-	$Q cd $(BASE) && $(GLIDE) --quiet install -v
-	@ln -nsf . vendor/src
-	@touch $@
+.PHONY: deps-update
+deps-update: $(info  updating dependencies...)
```
zshi-redhat, line 143: Missing comma after `deps-update: `. It makes `updating dependencies` info always be printed when `make` is executed even if it's not updating dependency.

SchSeba: This for the comment!

I fix it :)

zshi-redhat:
@SchSeba when generating go.mod, did you manually remove those indirect dependencies or it is not generated at all?  In my local run, there are three indrect dependencies and I was using go1.12.17 with GO111MODULE set to auto.

SchSeba:
> @SchSeba when generating go.mod, did you manually remove those indirect dependencies or it is not generated at all? In my local run, there are three indrect dependencies and I was using go1.12.17 with GO111MODULE set to auto.

@zshi-redhat I did not remove anything from the go.mod file manually.

I am using 1.13 with GO111MODULE  set to on

PR #226, opened 2020-04-20
Add more logs for rdma selector

Mmduh-483:


Mmduh-483:
@moshe010

pkg/factory/factory.go
```diff
@@ -114,6 +115,13 @@ func (rf *resourceFactory) GetResourcePool(rc *types.ResourceConfig, filteredDev
 		if len(filteredDevice) > 0 {
 			if _, ok := filteredDevice[0].(types.PciNetDevice); ok {
 				rPool = netdevice.NewNetResourcePool(rc, apiDevices, devicePool)
+				// check /dev/infiniband exist if rdma is required
+				selector, _ := rc.SelectorObj.(*types.NetDeviceSelectors)
+				if selector.IsRdma {
+					if _, err := os.Stat("/dev/infiniband"); err != nil {
```
zshi-redhat, line 121: If `/dev/infiniband` doesn't exist and `isRdma` is true, the `filteredDevice` would be zero.
I think this code line won't be reached.

zshi-redhat: Maybe add the rdma check in `GetFilteredDevices` when filtering devices or simply in `validConfigs`.

Mmduh-483: moved to `GetFilteredDevices`

pkg/factory/factory.go
```diff
@@ -158,13 +158,17 @@ func (rf *resourceFactory) GetDeviceFilter(rc *types.ResourceConfig) (interface{
 		if err := json.Unmarshal(*rc.Selectors, netDeviceSelector); err != nil {
 			return nil, fmt.Errorf("error unmarshalling NetDevice selector bytes %v", err)
 		}
+
+		glog.Infof("net device selector for resource %s is %+v", rc.ResourceName, netDeviceSelector)
```
adrianchiris, line 162: perhaps put this as debug ?

pkg/factory/factory.go
```diff
@@ -158,13 +158,17 @@ func (rf *resourceFactory) GetDeviceFilter(rc *types.ResourceConfig) (interface{
...
 		return netDeviceSelector, nil
 	case types.AcceleratorType:
 		accelDeviceSelector := &types.AccelDeviceSelectors{}
 
 		if err := json.Unmarshal(*rc.Selectors, accelDeviceSelector); err != nil {
 			return nil, fmt.Errorf("error unmarshalling Accelerator selector bytes %v", err)
 		}
+
+		glog.Infof("accelerator device selector for resource %s is %+v", rc.ResourceName, accelDeviceSelector)
```
adrianchiris, line 171: perhaps put this as debug ?

pkg/netdevice/netDeviceProvider.go
```diff
@@ -174,9 +175,12 @@ func (np *netDeviceProvider) GetFilteredDevices(devices []types.PciDevice, rc *t
 
 	// filter for rdma devices
 	if nf.IsRdma {
+		if _, err := os.Stat("/dev/infiniband"); err != nil {
```
adrianchiris, line 178: I think the netDeviceProvider should not be aware of RDMA spec internal details. (i,e the knowledge of `/dev/infiniband`)

i think that a better place to log the absence of the RDMA char devsis in `rdma.go:NewRdmaSpec()` where if we get no char devices we log a warning.

also perhaps in the log add some hint towards what might cause it and solution

pkg/factory/factory.go
Mmduh-483: There is no glog.Debug, also glog.V() is not used in the project

pkg/factory/factory.go
Mmduh-483: Same as above

pkg/netdevice/netDeviceProvider.go
Mmduh-483: moved to rdma.go

pkg/factory/factory.go
adrianchiris: I see.
As a general note perhaps it would be good to start using leveled logging to avoid clutter when not needed.

adrianchiris: we could start with the logs here :) 

anyway, im fine with it as is.

pkg/netdevice/rdma.go
```diff
@@ -16,6 +19,11 @@ type rdmaSpec struct {
 func NewRdmaSpec(pciAddrs string) types.RdmaSpec {
 	deviceSpec := make([]*pluginapi.DeviceSpec, 0)
 	isSupportRdma := false
+
+	if _, err := os.Stat("/dev/infiniband"); err != nil {
+		glog.Warning("dir \"/dev/infiniband\" is not found")
```
adrianchiris, line 24: what may cause this issue ? i think its when rdma modules are not loaded right ?

perhaps the warning could be: "dir /dev/infiniband is not found.  are RDMA modules loaded?" or something similar.
what do you think ?

it should be noted that the _"infiniband"_ in `/dev/infiniband` is just legacy name and also refers to chardevs created for RDMA over other fabrics as well e.g ethernet  (RoCE)

Mmduh-483: updated the warning message

pkg/netdevice/netDeviceProvider.go
zshi-redhat: @adrianchiris If moving `/dev/infiniband` check in rdma.go, then evey device in a non-rdma environment will hit this warning message, which might be confusing even if rdma is not configured. Since `NewRdmaSpec` is called in device discover phase (where resourceConfig is not passed in), we cannot detect whether `isRdma` is configured or not by user, so it's not easy to add condition for warning message.

I suggested to add it in `GetFilteredDevices` because we already have a condition there for `isRdma`.

pkg/factory/factory.go
zshi-redhat: This log here is duplicate with https://github.com/intel/sriov-network-device-plugin/blob/master/cmd/sriovdp/manager.go#L86.

pkg/netdevice/netDeviceProvider.go
adrianchiris: The code proposes to perform the check at the moment the RDMA Spec is requested.
without pushing this logic to upper layers.

Its true it has the drawback of reporting on every call. (which i originally weighted against having the logic in GetFilteredDevices)

its possible to run the check once or move the logic back to upper layer if thats acceptable.

zshi-redhat: When looking at this [proposal](https://github.com/intel/sriov-network-device-plugin/pull/213), it seems to me that rdma spec could also be retrieved when generating resource pool (similar to getting device spec), where a conditional check on `isRdma` can be applied.

adrianchiris: lets move this to its original location where you proposed.
if/once pool specific logic is introduced, this can then be moved there.

Mmduh-483:
@zshi-redhat Any comments on the PR?

zshi-redhat:
> @zshi-redhat Any comments on the PR?

@Mmduh-483 when #231 https://github.com/intel/sriov-network-device-plugin/pull/231 get merged, it will be possible to prompt device specific warning message for rdma spec by checking `resourceConfig.SelectorObj.IsRdma` && `rdmaSpec.isSupportRdma`  in `NewPciNetDevice`.

adrianchiris:
@Mmduh-483 seeing PR #231 got merged, this can now be reworked

Mmduh-483:
@zshi-redhat @adrianchiris  Updated the PR

Mmduh-483:
@zshi-redhat @ahalim-intel Can you provide review

PR #231, opened 2020-05-20
Create devices on Filter, not on Discovery

### LINKED ISSUE #212: Support pool-specific logic

amorenoz:
Currently the Device Provider performs two operations.

AddTargetDevices() discovers the devices only filtering by deviceType
and creates PciDevices.

GetFilteredDevices() filters the devices based on the ResourceConfig and
returns a subset of them.

Postpone the PciDevice creation to GetFilteredDevices() so that we can
consider the ResourceConfig on their creation

Currently, the ResouceConfig is ignored by NewPciNetDevice(). However, in future patches we can make use of it. Some examples would be:
Only discovering Rdma information if "isRdma" is present. (See [comment](https://github.com/intel/sriov-network-device-plugin/pull/226#discussion_r422451957) on #226)
Devices with Same PCI information but different ways to be exposed: See #212 

Fixes #212 


amorenoz:
@zshi-redhat , is this what you had in mind on [this discussion](https://github.com/intel/sriov-network-device-plugin/pull/213#issuecomment-626469053)?

amorenoz:
@ahalim-intel what do you think of this?

ahalim-intel:
Hi @amorenoz 
I have looked at these changes and I'm okay with it as long as this is needed. 
I'm just wondering how are you planning to integrate this with #213

amorenoz:
@ahalim-intel. This is a smaller PR that enables the same usecase as #213, i.e: pools-specific logic.

I still think the `PoolInfo` abstraction makes sense. If you agree, I'll add it on top by taking some logic out of NewPciNetDevice and moving it into a "newNetPoolInfo"

zshi-redhat:
> @ahalim-intel. This is a smaller PR that enables the same usecase as #213, i.e: pools-specific logic.
> 
> I still think the `PoolInfo` abstraction makes sense. If you agree, I'll add it on top by taking some logic out of `NewPciNetDevice` and moving it into `newNetPoolInfo`

I think this PR is mainly to address the issue:
Device type specific flags such as `isRdma` or `vDPA` don't apply to every device (e.g non-rdma network device or non-vDPA network device), but DP is discovering these capabilities for all network devices in `discoverHostDevices` without knowing if it is requested or not.  This is probably not desired as when adding warning messages for lacking of these capabilities, it prints out message for every device.

pkg/accelerator/accelDeviceProvider.go
```diff
@@ -25,22 +25,30 @@ import (
...
-	for i := range ap.deviceList {
-		newPciDevices[i] = ap.deviceList[i]
+func (ap *accelDeviceProvider) GetDiscoveredDevices() []*ghw.PCIDevice {
+	return ap.deviceList
+}
+
+func (ap *accelDeviceProvider) GetDevices(rc *types.ResourceConfig) []types.PciDevice {
+	newPciDevices := make([]types.PciDevice, 0)
+	for _, device := range ap.deviceList {
+		if newDevice, err := NewAccelDevice(device, ap.rFactory); err == nil {
```
zshi-redhat, line 47: This doesn't add ResourceConfig for NewAccelDevice. Is it because accelerator device doesn't have special flag or it is missed?

amorenoz: It's because it doesn't have any special flag. Since there is no interface that the device "constructors" have to follow, it can be `deviceType`-dependent.

amorenoz:
> > @ahalim-intel. This is a smaller PR that enables the same usecase as #213, i.e: pools-specific logic.
> > I still think the `PoolInfo` abstraction makes sense. If you agree, I'll add it on top by taking some logic out of `NewPciNetDevice` and moving it into `newNetPoolInfo`
> 
> I think this PR is mainly to address the issue:
> Device type specific flags such as `isRdma` or `vDPA` don't apply to every device (e.g non-rdma network device or non-vDPA network device), but DP is discovering these capabilities for all network devices in `discoverHostDevices` without knowing if it is requested or not. This is probably not desired as when adding warning messages for lacking of these capabilities, it prints out message for every device.

Exactly.
I was tempted to add the cleaning of the `isRdma` flag in this PR but decided that it was better for me to do that after this is merged.

ahalim-intel:
> @ahalim-intel. This is a smaller PR that enables the same usecase as #213, i.e: pools-specific logic.
> 
> I still think the `PoolInfo` abstraction makes sense. If you agree, I'll add it on top by taking some logic out of `NewPciNetDevice` and moving it into `newNetPoolInfo`

So, are you saying that if this is merged, we won't need tne `PoolInfo` abstraction for your usecase?

amorenoz:
> > @ahalim-intel. This is a smaller PR that enables the same usecase as #213, i.e: pools-specific logic.
> > I still think the `PoolInfo` abstraction makes sense. If you agree, I'll add it on top by taking some logic out of `NewPciNetDevice` and moving it into `newNetPoolInfo`
> 
> So, are you saying that if this is merged, we won't need tne `PoolInfo` abstraction for your usecase?

Right.
No, we don't strictly need it because the `ResouceConfig` can now be used [when the PciNetDevice is created](https://github.com/intel/sriov-network-device-plugin/pull/231/commits/6797173dded646b84eb6d098b2f24848b6c9f052#diff-d1eb2b748fa5b126329c3f0432f237e5)

PR #233, opened 2020-05-21
Use "go clean" to clean modules

### LINKED ISSUE #232: make clean fails: rm: cannot remove ... Permission denied
### What happened?

`make clean` no longer works. A long list of logs like this is shown:
`rm: cannot remove '.../sriov-network-device-plugin/.gopath/pkg/mod/golang.org/x/net@v0.0.0-20200324143707-d3edc9973b7e/route/interface_multicast.go': Permission denied`  

Probably due to the recent switch to "go mod" (#225 )

`make clean` should work

`make; make clean`

### Component Versions
Please fill in the below table with the version numbers of components used.

Component | Version|
------------------------------|--------------------|
|SR-IOV Network Device Plugin |master|

amorenoz:
Files downloaded by 'go mod' are not writable (removable).

Use the recommended way of cleaning modules [1]

[1] https://github.com/golang/go/issues/27161#issuecomment-415213240

Fixes #232 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,6 +153,7 @@ image: | $(BASE) ; $(info Building Docker image...)
 
 .PHONY: clean
 clean: ; $(info  Cleaning...)	@ ## Cleanup everything
+	@go clean --modcache
 	@rm -rf $(GOPATH)
 	@rm -rf $(BUILDDIR)/$(BINARY_NAME)
 	@rm -rf test/tests.* test/coverage.*


zshi-redhat:
This PR works for me.

One additional comment (for discussion) is after switching to go mod, `make` will download modules when GO111MODULE is on which is time-consuming and not necessary as we have local vendor pkgs up to date. Should we disable it in Makefile for regular build and only enable it when running deps-update?

ahalim-intel:
@zshi-redhat Sorry, just noticed your last comment here now. Let's review the behavior and discuss proper steps here. I think it's important. Right now, `make` doesn't update modules with `GO111MODULE=on` unless I clean the module cache with `make clean` beforehand. I'm on go version go1.13.8

PR #235, opened 2020-05-25
fix pfNames selector to skip device with no PF name

ahalim-intel:
Current behaviour of pfName selector adds devices that do not have
PF names available. For instance, if there is a PF attached to
igb_uio driver and VFs are created under that PF, those VFs will
have no PF name information. If a resource config prvides the
"pfName" selector any VFs under that PF(igb_uio driver) will
wrongly be added to that pool. This happens because this selector
matches PF name with string.HasPrefix() method and providing ""
for PF name results in true.

Fix: skip devices that do not have PF names when "pfNames" selector
is used.

Change-Id: I1ddd49743ef2e8f25718285e80c16dfcef4c90f5
Signed-off-by: Abdul Halim <abdul.halim@intel.com>


ahalim-intel:
fixed typos in commit message

PR #239, opened 2020-06-08
Update rhel7 builder image go version

SchSeba:


SchSeba:
@zshi-redhat PTAL

zshi-redhat:
/lgtm

This is an error introduced when adding `-modcache` in make clean entry that:
make clean fails with golang-1.10: "flag provided but not defined: -modcache"

The proposed change is to update base image to use golang-1.13.

@ahalim-intel PTAL

PR #240, opened 2020-06-08
added VF set up guide to docs

killianmuldoon:
Added updated version of VF creation guide from SR-IOV CNI docs to the device plugin docs folder as per https://github.com/intel/sriov-cni/issues/74

Input welcome to update the info on VF creation here. 

Once this is merged it can be linked from the new SR-IOV CNI docs contained in https://github.com/intel/sriov-cni/pull/127

### DIFF
--- a/docs/vf-setup.md
+++ b/docs/vf-setup.md
@@ -1,0 +1,152 @@
+# Setting up Virtual Functions
+
+The SR-IOV Network Device Plugin requires SR-IOV virtual functions (VFs) to be created on the host ahead of startup. The device plugin does not manage the creation of virtual functions, and it does not automatically update when there is a change in the virtual functions. Each time the number of VFs or the driver used by the VF change the SR-IOV Network device plugin needs to be restarted. 
+This guide gives an overview for setting up SR-IOV virtual functions on a linux host for various NICs. 
+
+  * [Intel](#intel)
+  * [Mellanox](#mellanox)
+
+## Intel
+The below works on Intel SR-IOV enabled cards including those from the Gigabit Ethernet, X710 and X722 families of Network Interface Cards. Note not all Gigabit Ethernet NICs are enabled with SR-IOV.
+
+Cards from the following families can be enabled through the sysfs endpoint as described below:
+* **Intel i40e Driver**
+	* Intel® Ethernet Controller X722 
+	* Intel® Ethernet Controller X710
+* **Intel ixgbe Driver** 
+	* Intel® Ethernet Controller 82599
+	* Intel® Ethernet Controller X520
+	* Intel® Ethernet Controller X540
+	* Intel® Ethernet Controller x550
+	* Intel® Ethernet Controller X552
+	* Intel® Ethernet Controller X553
+
+### Creating VFs with sysfs
+First select a compatible NIC on which to create VFs and record its name (shown as PF_NAME below). 
+
+To create 8 virtual functions run:
+```
+echo 8 > /sys/class/net/${PF_NAME}/device/sriov_numvfs
+```
+To check that the VFs have been successfully created run:
+
+```
+lspci | grep "Virtual Function"
+``` 
+
+This method requires the creation of VFs each time the node resets. This can be handled automatically by placing the above command in a script that is run on startup such as `/etc/rc.local`.
+
+### Common issues
+
+**Cannot allocate memory**
+
+Some cards using the `ixgbe` may require additional configuration. This is likely the case if the message `write error: Cannot allocate memory` is returned when creating VFs using the above method. This issue has been observed on the X552 NIC. 
+
+To resolve this issue, try setting the maximum allowed VFs in the driver config:
+
+``` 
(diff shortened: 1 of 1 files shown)


adrianchiris:
LGTM

docs/vf-setup.md
```diff
@@ -0,0 +1,126 @@
...
+Cards from the following families can be enabled through the sysfs endpoint as described below:
+* **Intel i40e Driver**
+	* Intel® Ethernet Controller X722 
+	* Intel® Ethernet Controller X710
+* **Intel ixgbe Driver** 
+	* Intel® Ethernet Controller 82599
+	* Intel® Ethernet Controller X520
+	* Intel® Ethernet Controller X540
+	* Intel® Ethernet Controller x550
+	* Intel® Ethernet Controller X552
```
zshi-redhat, line 21: I have seen one issue when provisioning VFs with intel x552 ixgbe card, that it requires two steps:
1) modprobe -r ixgbe; modprobe ixgbe max_vfs=3
2) echo 3 > /sys/class/net/${PF_NAME}/device/sriov_numvfs

Without step 1), step 2) fails with `write error: Cannot allocate memory`. Is this expected behavior?

killianmuldoon: The driver doc says 

> NOTE: This parameter is only used on kernel 3.7.x and below. On kernel 3.8.x
and above, use sysfs to enable VFs. Also, for Red Hat distributions, this
parameter is only used on version 6.6 and older. For version 6.7 and newer, use
sysfs.

Are you on an older system like that? I think we should be able to figure out the cutoff version for this behaviour.

zshi-redhat: @killianmuldoon we were using centos 7/8, it has newer kernel versions than 6.7.
@yrobla ^^

yrobla: The cannot allocate memory problem was actually a missing setting on my bios. However, i am experimenting errors on my x552 card. As soon as i do something like "echo 3 > /sys/class/net/eno3/device/sriov_numvfs" , the system reboots and seems unable to persistently keep this value.

zshi-redhat: note: this doesn't necessary block the merging of this PR, it is just to check if we have ever hit the same issue.

killianmuldoon: @zshi-redhat I've not had a change to figure out the issue here or where the cutoff might be for the cards/OS. I've put in a troubleshooting block with your instructions and specifically called out the error message and the 552 NIC - hopefully that should help people who run into this issue.

docs/vf-setup.md
```diff
@@ -0,0 +1,141 @@
...
+To resolve this issue, try setting the maximum allowed VFs in the driver config:
+
+``` 
+modprobe -r ixgbe; modprobe ixgbe max_vfs=3
+
+```
+Next set the VFs as above
+
+```
+echo 8 >> /sys/class/net/${PF_NAME}/device/sriov_numvfs
```
ahalim-intel, line 51: I think this is a type here? This should be `>` instead of `>>`

killianmuldoon: Thanks for catching that - fixed

docs/vf-setup.md
```diff
@@ -0,0 +1,141 @@
+# Setting up Virtual Functions
+
+The SR-IOV Network Device Plugin requires SR-IOV virtual functions (VFs) to be created on the host ahead of startup. The device plugin does not manage the creation of virtual functions, and it does not automatically update when there is a change in the virtual functions. Each time the number of VFs or the driver used by the VF change the SR-IOV Network device plugin needs to be restarted. 
+This guide gives an overview for setting up SR-IOV virtual functions on a linux host for various NICs. 
+
+  * [Intel](#intel)
+  * [Mellanox](#mellanox)
+
+## Intel
+The below works on intel SR-IOV enabled cards including those from the Gigabit Ethernet, X710 and X722 families of Network Interface Cards. Note not all Gigabit Ethernet NICs are enabled with SR-IOV.
```
martinkennelly, line 10: > The below works on intel SR-IOV enabled ..

Could you reword this
>intel

Intel*

docs/vf-setup.md
```diff
@@ -0,0 +1,141 @@
...
+	* Intel® Ethernet Controller x550
+	* Intel® Ethernet Controller X552
+	* Intel® Ethernet Controller X553
+
+### Creating VFs with sysfs
+First select a compatible NIC on which to create VFs and record its name (shown as PF_NAME below). 
+
+To create 8 virtual functions run:
+```
+echo 8 > /sys/class/net/${PF_NAME}/device/sriov_numvfs
```
martinkennelly, line 29: Should we reset the number of VFs here before asking for 8 in case it was configured previously and will throw an error if its not reset first before asking for a new set of VFs?

killianmuldoon: Added a section under "common problems" to pick up this error if someone wants to reset the number of VFs

PR #243, opened 2020-06-10
Change daemonset imagePullPolicy from Never to IfNotPreset

Mmduh-483:
With image imagePullPolicy set to IfNotPreset the user doesn't need to pull the image manually


Mmduh-483:
@zshi-redhat @ahalim-intel Can you please review the PR?

PR #250, opened 2020-06-23
added known good ddptool source to Dockerfile

killianmuldoon:
Edited the source of the ddp tool to a known good mirror. This is a quick bridging fix to get builds working.


zshi-redhat:
Both old and new links work for me, it's better if someone who fails on old link can verify the new one. 
@SchSeba any chance to give it a try?

PR #254, opened 2020-07-14
Fix CVE-2020-14040

zshi-redhat:
golang/x/text UTF-16 decoder behaves incorrectly on single-byte input


adrianchiris:
@Mmduh-483 

@zshi-redhat have you checked which package/s depend on it and where it interfaces with our project ?
fixing CVEs is important so im LGTM :)

zshi-redhat:
> @Mmduh-483
> 
> @zshi-redhat have you checked which package/s depend on it and where it interfaces with our project ?
> Im all for fixing CVEs so im LGTM :)

I didn't check which exact function is called directly by device plugin that will cause issue.
There was a RHBZ that tracks this security issue(which is not intended to be public), the fix is in golang/x/text/unicode v0.3.3 version. some more info can be found: https://groups.google.com/g/golang-announce/c/bXVeAmGOqz0?pli=1

PR #257, opened 2020-07-28
Use latest image version v3.2 in daemon templates

### LINKED ISSUE #256: images in yaml file are not updated when releases are tagged
## What issue would you like to bring attention to?
Currently the tags that you do to release a new version do not include the charts being updated, this makes it harder for people to install a specific version. If you look at plugins such as multus and flannel you will see that they also update the yaml files when tagging.

## What is the impact of this issue?
It makes providing instructions on how to install specific versions harder and increases the risk that people will have mismatches between the version of the yaml file and the version of the software if they manually change the yaml file.

## Do you have a proposed response or remediation for the issue?
As part of the release process you could add updating the image tags in the yaml file to reflect the new tag.

zshi-redhat:
Fixes #256

### DIFF
--- a/deployments/k8s-v1.10-v1.15/sriovdp-daemonset.yaml
+++ b/deployments/k8s-v1.10-v1.15/sriovdp-daemonset.yaml
@@ -32,7 +32,7 @@ spec:
       serviceAccountName: sriov-device-plugin
       containers:
       - name: kube-sriovdp
-        image: nfvpe/sriov-device-plugin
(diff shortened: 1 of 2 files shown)


adrianchiris:
So its true just for `amd64` arch right ?

zshi-redhat:
> So its true just for `amd64` arch right ?

Yes, for now it's only amd64.
I think we don't have builds for other arch yet.

> 
> i think we should have images per arch with release tags, then the rest can be updated (preferably all in the same repo...)
> 
> LGTM

PR #263, opened 2020-08-19
pfName selector: match exact pf name

zshi-redhat:


zshi-redhat:
pfNames may contain names with overlapped prefix such as ens3f0, ens3;
this results in VFs from ens3 be wrongly exposed as extended resource
when pfNames only contain ens3f0.

PR #272, opened 2020-10-07
Add support for running inside a VM with vfio-noiommu

atyronesmith:
This PR is required for https://github.com/openshift/sriov-network-operator/pull/369.

This PR enables the mapping of a no-iommu vfio device into a container.  

Associated system configuration enables no-iommu for the vfio-pci driver.

This PR is required to allow Kubernetes to run in a virtual environment that does not provide a virtualized iommu.  The initial target environment is OpenStack.  OpenStack Nova does not currently provide a virtualized iommu. Support for vfio-noiommu was added to linux a few years ago with the caveats outlined below...

`There is really no way to safely give a user full access to a DMA
capable device without an IOMMU to protect the host system.  There is
also no way to provide DMA translation, for use cases such as device
assignment to virtual machines.  However, there are still those users
that want userspace drivers even under those conditions.  The UIO
driver exists for this use case, but does not provide the degree of
device access and programming that VFIO has.  In an effort to avoid
code duplication, this introduces a No-IOMMU mode for VFIO.

This mode requires building VFIO with CONFIG_VFIO_NOIOMMU and enabling
the "enable_unsafe_noiommu_mode" option on the vfio driver.  This
should make it very clear that this mode is not safe.  Additionally,
CAP_SYS_RAWIO privileges are necessary to work with groups and
containers using this mode.  Groups making use of this support are
named /dev/vfio/noiommu-$GROUP and can only make use of the special
VFIO_NOIOMMU_IOMMU for the container.  Use of this mode, specifically
binding a device without a native IOMMU group to a VFIO bus driver
will taint the kernel and should therefore not be considered
supported.  This patch includes no-iommu support for the vfio-pci bus
driver only.`

In the future, the current QEMU machine type in use for OpenStack is i440fx.  The
machine type should be switched to Q35 in the future.  The Q35 machine provides
a virtualized iommu implementation and will obsolete this patch.  

In summary, the use of noiommu introduces security concerns, but the performance
boost provided by vfio-pci can be justified in many cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ The SR-IOV network device plugin is Kubernetes device plugin for discovering and
 - Detects Kubelet restarts and auto-re-register
 - Detects Link status (for Linux network devices) and updates associated VFs health accordingly
 - Extensible to support new device types with minimal effort if not already supported
+- Works within virtual deployments of Kubernetes that do not have virtualized-iommu support (VFIO No-IOMMU support)
 
 To deploy workloads with SR-IOV VF this plugin needs to work together with the following two CNI components:
 
@@ -366,6 +367,36 @@ $ kubectl get node node1 -o json | jq '.status.allocatable'
 }
 
 ```
+
+### Virtual environments with no iommu
+
+In virtual deployments of Kubernetes where the underlying virtualization platform does not support a virtualized iommu, the VFIO driver needs to be loaded with a special 
+flag.  The file **/etc/modprobe.d/vfio-noiommu.conf** must be created with the contents:
+
+````
+# cat /etc/modprobe.d/vfio-noiommu.conf
+options vfio enable_unsafe_noiommu_mode=1
+````
+
+With the above option, vfio devices will be created with the form on the virtual host (VM):
+
+````
+/dev/vfio/noiommu-0
+/dev/vfio/noiommu-1
+...
+````
+
+The presence of noiommu-* devices will automatically be detected by the sriov-device-plugin.  The noiommu-N devices will be mounted **inside** the pod in their expected/normal location;
+
+````
+/dev/vfio/0
+/dev/vfio/1
+...
+````
+It should be noted that with no IOMMU, there is no way to ensure safe use of DMA.  When *enable_unsafe_noiommu_mode* is used, CAP_SYS_RAWIO privileges are necessary to work with groups and
+containers using this mode.  Use of this mode, specifically
+binding a device without a native IOMMU group to a VFIO bus driver will taint the kernel.  Only no-iommu support for the vfio-pci bus is provided.  However, there are still those users
+that want userspace drivers even under those conditions.
 
 ## Example deployments
 
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@
   - [Command line arguments](#command-line-arguments)
   - [Assumptions](#assumptions)
   - [Workflow](#workflow)
+- [Virtual Deployments](#virtual-deployments)
 - [Example deployments](#example-deployments)
     - [Deploy the Device Plugin](#deploy-the-device-plugin)
     - [Deploy SR-IOV workloads when Multus is used](#deploy-sr-iov-workloads-when-multus-is-used)
@@ -367,45 +368,18 @@ $ kubectl get node node1 -o json | jq '.status.allocatable'
 }
 
 ```
+## Virtual Deployments
+
+SR-IOV network device plugin supports running in a virtualized environment.  However, not all device selectors are 
+applicable as the VFs are passthrough to the VM without any association to their respective PF, hence any device 
+selector that relies on the association between a VF and its PF will not work and therefore the _pfNames_ and 
+_rootDevices_ extended selectors will not work in a virtual deployment.  The common selector _pciAddress_ can be 
+used to select the virtual device.
 
 ### Virtual environments with no iommu
 
-In virtual deployments of Kubernetes where the underlying virtualization platform does not support a virtualized iommu, the VFIO driver needs to be loaded with a special 
-flag.  The file **/etc/modprobe.d/vfio-noiommu.conf** must be created with the contents:
-
-````
-# cat /etc/modprobe.d/vfio-noiommu.conf
-options vfio enable_unsafe_noiommu_mode=1
-````
-
-With the above option, vfio devices will be created with the form on the virtual host (VM):
-
-````
-/dev/vfio/noiommu-0
-/dev/vfio/noiommu-1
-...
-````
-
-The presence of noiommu-* devices will automatically be detected by the sriov-device-plugin.  The noiommu-N devices will be mounted **inside** the pod in their expected/normal location;
-
-````
-/dev/vfio/0
-/dev/vfio/1
-...
-````
-It should be noted that with no IOMMU, there is no way to ensure safe use of DMA.  When *enable_unsafe_noiommu_mode* is used, CAP_SYS_RAWIO privileges are necessary to work with groups and
-containers using this mode.  Use of this mode, specifically
-binding a device without a native IOMMU group to a VFIO bus driver will taint the kernel.  Only no-iommu support for the vfio-pci bus is provided.  However, there are still those users
-that want userspace drivers even under those conditions.
-
-Unfortunately, the SR-IOV CNI is not compatible with the _noiommu_ feature.  To use the _noiommu_ feature, remove the 
-networks annotation from the container spec.  To foster compatiblity between virtual and baremetal deployments creating the associated SrioveNetwork CR is recommended.
-
-````yaml
-  annotations:
-#    k8s.v1.cni.cncf.io/networks: sriov-net1
-  ...
-````
+SR-IOV network device plugin supports allocating VFIO devices in a virtualized environment without a virtualized iommu.
+For more information refer to [this](./docs/dpdk/README-virt.md).
 
 ## Example deployments
 
--- a/README.md
+++ b/README.md
@@ -398,6 +398,15 @@ containers using this mode.  Use of this mode, specifically
 binding a device without a native IOMMU group to a VFIO bus driver will taint the kernel.  Only no-iommu support for the vfio-pci bus is provided.  However, there are still those users
 that want userspace drivers even under those conditions.
 
+Unfortunately, the SR-IOV CNI is not compatible with the _noiommu_ feature.  To use the _noiommu_ feature, remove the 
+networks annotation from the container spec.  To foster compatiblity between virtual and baremetal deployments creating the associated SrioveNetwork CR is recommended.
+
+````yaml
+  annotations:
+#    k8s.v1.cni.cncf.io/networks: sriov-net1
+  ...
+````
+
 ## Example deployments
 
 We assume that you have working K8s cluster configured with one of the supported meta plugins for multi-network support. Please see [Features](#features) and [Quick Start](#quick-start) sections for more information on required CNI plugins.
--- a/docs/dpdk/README-virt.md
+++ b/docs/dpdk/README-virt.md
@@ -1,0 +1,102 @@
+# Running DPDK applications in a Kubernetes virtual environment
+
+## Pre-requisites
+
+### Hugepages
+DPDK applications require Hugepages memory. Please refer to the [Hugepages section](http://doc.dpdk.org/guides/linux_gsg/sys_reqs.html#use-of-hugepages-in-the-linux-environment) in DPDK getting started guide on hugespages in DPDK.
+
+Make sure that the virtual environment is enabled for creating VMs with hugepage support.  
+
+Kubernetes nodes can only advertise a single size pre-allocated hugepages. Which means even though one can have both 2M and 1G hugepages in a system, Kubernetes will only recognize the default hugepages as schedulable resources. Workload can request for hugepages using resource requests and limits specifying `hugepage-2Mi` or `hugepage-1Gi` resource references.
+
+> One important thing to note here is that when requesting hugepage resources, either memory or CPU resource requests need to be specified.
+
+For more information on hugepage support in Kubernetes please see [here](https://kubernetes.io/docs/tasks/manage-hugepages/scheduling-hugepages/).
+
+
+### VF drivers
+DPDK applications require devices to be attached with supported dpdk backend driver.
+* For Intel® x700 series NICs `vfio-pci` is required.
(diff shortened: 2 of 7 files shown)


pkg/utils/utils.go
```diff
@@ -291,6 +291,19 @@ func GetVFIODeviceFile(dev string) (devFile string, err error) {
...
+	namePath := filepath.Join(linkName, "name")
+	// Read the iommu group name
+	// The name file will not exist on baremetal
+	vfioName, errName := ioutil.ReadFile(namePath)
+	if errName == nil {
+		vName := strings.TrimSpace(string(vfioName))
+
+		// if the iommu group name == vfio-noiommu then we are in a VM, adjust path to vfio device
+		if vName == "vfio-noiommu" {
+			linkName = filepath.Join(filepath.Dir(linkName), "noiommu-"+filepath.Base(linkName))
```
zshi-redhat, line 304: so the symbol link of iommuDir for no-iommu vfio device is the same as iommu vfio device, that's why we need to add `noiommu-` prefix here. Just curious the symbol link of iommuDir doesn't contain the actual device spec.

pkg/utils/utils.go
```diff
@@ -291,6 +291,19 @@ func GetVFIODeviceFile(dev string) (devFile string, err error) {
 		return
 	}
 
+	// Get a file path to the iommu group name
+	namePath := filepath.Join(linkName, "name")
```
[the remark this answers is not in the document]
atyronesmith: Inside the /sys/bus/pci/devices/xxxx:xx:xx:x.x directory is a symlink call iommu_groups which is a 
link such as: iommu_group -> ../../../kernel/iommu_groups/1.  When evaluated, this symlink looks like:
/sys/kernel/iommu_groups/1.  

So at this point in the code, linkName == /sys/kernel/iommu_groups/0

pkg/utils/utils.go
atyronesmith: The directory does contain a device symlink.  i.e.

sh-4.2# ls -l /sys/kernel/iommu_groups/0                        
total 0
drwxr-xr-x. 2 root root    0 Oct 13 14:14 devices
-r--r--r--. 1 root root 4096 Oct 12 17:21 name
-r--r--r--. 1 root root 4096 Oct 13 14:14 reserved_regions
-r--r--r--. 1 root root 4096 Oct 13 14:14 type

devices points to the device dir...

on virtual, the type and the name are present.

As a shortcut, I decided to key off of the presence and value of the name file.
The name file is present for the noiommu case and not present for the baremetal
case (in my testing).  In addition, the name file == vfio-noiommu.  I use this fact
as postive confirmation that we are running virtual.  If we are running on a virtual
platform with virtualized iommu, the name field should not be present and/or the
name will not be vfio-noiommu from looking at the kernel code.

The name file is references [here](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-kernel-iommu_groups).

zshi-redhat: Aaron, thanks for the detailed explanation!

pkg/utils/utils.go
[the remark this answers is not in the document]
zshi-redhat: ack+

pkg/utils/utils.go
ahalim-intel: I'm not very clear on this. What would be the deviceSpec for noiommu vfio devices?
Which device files do we need to mount in the container?
is this going to be always:
`/dev/vfio/vfio` and 
`/dev/vfio/0`
?

atyronesmith: There is no additional deviceSpec required for noiommu.  Part of additional code determines if the noiommu device exists.  If the device, noiommu, exists, then it is used.  The  /dev/vfio/noiommu-0 device in the VM is mounted as /dev/vfio/0 inside the container.

ahalim-intel: So, if the /dev/vfio/0 in host with noiommu will be mounted as /dev/vfio/0 in container why do we need this change here?

atyronesmith: Sorry, I was unclear/wrong about deviceSpec...   

in the VM case in the following section of code (vfrioResource.go,59), (HostPath) vfioDev == /dev/vfio/noiommu-0.  
The containerPath in this case need to be /dev/vfio/0, hence the replace.

		devSpecs = append(devSpecs, &pluginapi.DeviceSpec{
			HostPath:      vfioDev,
			ContainerPath: strings.Replace(vfioDev, "noiommu-", "", 1),
			Permissions:   "mrw",
		})

ahalim-intel: Isn't this being replaced because it was concatenated in L304? If we didn't add that in there there'd be no need to do the string replace here. I mean without adding "noiommu-" prefix to that iommu_group '0' we would still get the usual `/dev/vfio/0` as we would get in normal vfio case. 

What I am trying to understand is what do we need to do differently in "noiommu" cases?
Am I missing something? I don't have a VM instance to verify this properly.

atyronesmith: Inside the VM, when you load and bind vfio-pci, the device shows up as  /dev/vfio/noiommu-N. where N is 0,1,2, etc...   
Not /dev/vfio/0.  The current sriov-network-device-plugin code looks for /dev/vfio/0 on the host (which in this case is a VM).  /dev/vfio/0 does not exists.   The new code checks for the VM case by checking an optional *name* file that exists before the device has a special name instead of the default name.  If the *name* file is found and the name is vfio-noiommu, then the hostPath has to be /dev/vfio/noiommu-0, but the containerPath must remain /dev/vfio/0 to make things compatible across platforms.

ahalim-intel: > Inside the VM, when you load and bind vfio-pci, the device shows up as /dev/vfio/noiommu-N. where N is 0,1,2, etc...

^This is what I wasn't aware of :) Thank you!

adrianchiris (approved):

adrianchiris (approved):
Overall code looks OK to me.
Small comment posted.

pkg/resources/vfioResource.go
```diff
@@ -52,11 +54,11 @@ func (rp *vfioResource) GetDeviceSpecs(pciAddr string) []*pluginapi.DeviceSpec {
 
 	vfioDev, err := utils.GetVFIODeviceFile(pciAddr)
 	if err != nil {
-		glog.Errorf("GetDeviceSpecs(): error getting vfio device file for device: %s", pciAddr)
+		glog.Errorf("GetDeviceSpecs(): error getting vfio device file for device: %s, %s", pciAddr, err.Error())
 	} else {
 		devSpecs = append(devSpecs, &pluginapi.DeviceSpec{
 			HostPath:      vfioDev,
-			ContainerPath: vfioDev,
+			ContainerPath: strings.Replace(vfioDev, "noiommu-", "", 1),
```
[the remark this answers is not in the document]
atyronesmith: This is indeed clumsy code.  I wanted to minimize changes to the current code as much as possible.  I could change the return signature of the function call to return a hostpath and a container path.  

The reason for this Replace is that vfioDev might contain. noiommu-0(1,2,etc) in the VM use case.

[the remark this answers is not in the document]
adrianchiris: > I could change the return signature of the function call to return a hostpath and a container path
this will work as well, then all logic is rolled into `GetVFIODeviceFile(...)`

moshe010 (commented):
1. Can update the readme for the VM with vfio-noiommu? 
2. According to your PR description the pod should have SYS_RAWIO in the capabilities or to be privileged maybe you can have examples like we have for dpdk or rdma https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/tree/master/docs/dpdk.
3. It is not explicitly stated in the PR, but my understanding is that the use-case is for running dpdk POD in the VM. Am I right? if that the case I think we need to  update plugin/tree/master/docs/dpdk/README file with that use-case.

martinkennelly:
Second the three points made by @moshe010 especially emphasizing the addition of documentation. PR looks good. I am currently testing it locally. Thank you for this.

atyronesmith:
@moshe010 @martinkennelly Docs pushed

zshi-redhat (approved):

docs/dpdk/README-virt.md
```diff
@@ -0,0 +1,102 @@
...
+Certain privileges are required for dpdk application to function properly in Kubernetes Pod. The level of privileges depend on the application and the host device driver attached (as mentioned above).  When running in an environment without a fully virtualized IOMMU, the *enable_unsafe_noiommu_mode* of vfio must be used by creating a modprobe.d file.
+
+````
+# cat /etc/modprobe.d/vfio-noiommu.conf
+options vfio enable_unsafe_noiommu_mode=1
+````
+
+With `vfio-pci` an application must run privilege Pod with  **IPC_LOCK** and **CAP_SYS_RAWIO** capability.
+
+# Example deployment
```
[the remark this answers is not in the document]
martinkennelly: I tried with host-device, but also failed. 
I am getting: 
>failed to find host device: no net directory under pci device 0000:00:06.0

Is there a patch outstanding on a CNI to fix this for vfio-pci devices?

martinkennelly:
LGTM overall besides nits / CNI Question.

atyronesmith:
@martinkennelly Aplogies, forgot to document the SR-IOV CNI situation. I will include an example that I have been using for testing.  Basically, just remove the annotations...

````yaml
  annotations:
    k8s.v1.cni.cncf.io/networks: sriov-net1
````

and keep the resources...
````yaml
    resources:
      requests:
        intel.com/intel_sriov_netdevice: '1'
      limits:
        intel.com/intel_sriov_netdevice: '1'
```
unchanged.

docs/dpdk/README-virt.md
```diff
@@ -0,0 +1,114 @@
...
+# Performance
+It is worth mentioning that to achieve maximum performance from a dpdk application the followings are required:
+
+1. Application process needs to be pinned to some dedicated isolated CPUs. Detailing how to achieve this is out of scope of this document. You can refer to [CPU Manager for Kubernetes](https://github.com/intel/CPU-Manager-for-Kubernetes) that provides such functionality in Kubernetes.  In the virtualized case, cpu pinning and isolation must be considered at the phyiscal layer as well as the virtual layer.
+
+2. All application resources(CPUs, devices and memory) are from same NUMA locality.  In the virtualized case, NUMA locality is controlled by the underlying virtualized platform for the VM.
+
+# Pod Usage
+
+Unfortunately, the SR-IOV CNI is not compatible with the noiommu feature.  To use the noiommu feature, remove the
```
adrianchiris, line 105: This section kinda hints that SR-IOV CNI should support this, which i am not sure is the case.

suggestion:
just have the pod example, then a note:

> _*Note*_: When consuming a VFIO device in a virtual environment secondary network is not required as network configuration for the underlying VF should be performed at the hypervisor level.

docs/dpdk/pod_testpmd_virt.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: testpmd
+#  annotations:
```
adrianchiris, line 5: no need for the commented annotation IMO

there are usecases where device plugin is used without a CNI so while sriov cni needs device plugin to operate in a k8s environment the other way is not true.

docs/dpdk/README-virt.md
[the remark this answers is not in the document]
adrianchiris: so at this point no CNI is expected to be used if i understood correctly

README.md
```diff
@@ -367,6 +368,45 @@ $ kubectl get node node1 -o json | jq '.status.allocatable'
 
 ```
 
+### Virtual environments with no iommu
```
adrianchiris, line 371: This seems to me  a rather specialized usecase, IMO we move this section under docs/dpdk/README-virt

we can have a general description here and point to docs/dpdk/README-virt
we should also explain that not all selectors are applicable.

e.g

### Virtual environments

SR-IOV network device plugin supports running in a virtualized environment, however, not all device selectors are applicable as the VFs are passthrough to the VM without any association to their respective PF, hence any device selector that relies on the association between a VF and its PF will not work.

The following selectors will not work in a virtualized environment: `pfNames`, `rootDevices`

#### Virtual environments with no iommu
SR-IOV network device plugin supports allocating VFIO devices in a virtualized environment without a virtualized iommu
for more information refer to *{add link}*

adrianchiris: P.S
I realize we are missing some documentation on consuming sr-iov resources in virtualized environment in general, however thats not related to this PR.

adrianchiris (changes requested):

docs/dpdk/pod_testpmd_virt.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: testpmd
+spec:
+  containers:
+  - name: testpmd
+    image: <DPDK testpmd image>
+    securityContext:
+        privileged: true
```
adrianchiris, line 10: so in the example pod should it be privileged ? or will it generally work with CAP_IPC_LOCK and CAP_SYS_RAWIO ?
id prefer to list the minimum capabilities.

docs/dpdk/README-virt.md
```diff
@@ -0,0 +1,137 @@
...
+# Performance
+It is worth mentioning that to achieve maximum performance from a dpdk application the followings are required:
+
+1. Application process needs to be pinned to some dedicated isolated CPUs. Detailing how to achieve this is out of scope of this document. You can refer to [CPU Manager for Kubernetes](https://github.com/intel/CPU-Manager-for-Kubernetes) that provides such functionality in Kubernetes.  In the virtualized case, cpu pinning and isolation must be considered at the phyiscal layer as well as the virtual layer.
+
+2. All application resources(CPUs, devices and memory) are from same NUMA locality.  In the virtualized case, NUMA locality is controlled by the underlying virtualized platform for the VM.
+
+# Usage
+
+_When consuming a VFIO device in a virtual environment, a secondary network is not required as network configuration for the underlying VF should be performed at the hypervisor level._
```
adrianchiris, line 133: nit: formatting her should be with a `>` to emphasize its a note/commet similarly as done in other part of the docs e.g L40 in this file

> When consuming a VFIO device in a virtual environment, a secondary network is not required as network configuration for the underlying VF should be performed at the hypervisor level

zshi-redhat:
@adrianchiris could you take a look at the new doc update?

adrianchiris (approved):
LGTM, Thank you !

martinkennelly (approved):
LGTM! Thanks!

PR #279, opened 2020-11-03
add --config-file in entrypoint.sh

zshi-redhat:
config-file cmd option is supported in sriovdp binary but not in image entrypoint script.
With this commit, config-file option is added in the entrypoint.sh which can be used to
specify individual configuration for device plugin daemonSet pod for a certain node.
Compared with global device plugin configMap, it allows more granular control over
which resource shall be exposed on a specific node.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -6,3 +6,4 @@ This page contains supplimentary documention that users may find useful for vari
 * [Using UDEV Rules for changing NIC's network device name](udev/)
 * [Running RDMA application in Kubernetes](rdma/)
 * [SR-IOV network device plugin with DDP](ddp/)
+* [Using node specific config file for running device plugin DaemonSet](config-file)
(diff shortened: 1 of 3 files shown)


moshe010 (approved):
LGTM

adrianchiris:
nit: some more info in the commit message on why is this required would be helpful for posterity

LGTM

amorenoz:
Do we need to update the README explaining this new way of configuring the DP?

zshi-redhat:
> nit: some more info in the commit message on why is this required would be helpful for posterity
> 
> LGTM

Updated commit message.

adrianchiris:
> Updated commit message.

Thanks !

zshi-redhat:
> Do we need to update the README explaining this new way of configuring the DP?

@amorenoz The `Command line arguments` describes all the options including `config-file` for sriovdp binary. 
I think this one is more of using it in the device plugin daemon which we don't have a doc section for it.
Do you think I shall add an example in the daemonSet.yaml template about how to specify this config per node?

amorenoz:
@zshi-redhat  Maybe an example deployment (or a comment in the existing one) would be nice

martinkennelly:
LGTM.

martinkennelly (approved):

zshi-redhat:
> @zshi-redhat Maybe an example deployment (or a comment in the existing one) would be nice

added doc for node specific config file under `docs/nodeconfig`

adrianchiris (approved):
Added documentation is good IMO

martinkennelly:
Added documentation is great! We have passed merge criteria and therefore this is merged. Thanks Zenghui.

PR #280, opened 2020-11-03
update code and doc with NPWG git repos URLs

mcoquelin:
The README.md was still pointing to Intel's repos for
both srio-network-device-plugin and srio-cni. This patch
changes to to NPWG's repos.

Signed-off-by: Maxime Coquelin <maxime.coquelin@redhat.com>

PR #281, opened 2020-11-05
Use DeviceInfoProviders for all DeviceSpec generation

amorenoz:
Currently, the generation of DeviceSpecs is done in two different places.

First, the PciNetDevice is created, and it may create alternative deviceSpecs objects (e.g: `RdmaSpecs`) that are stored within `PciNetDevice`. Then the underlying PciDevice is created and, based on the driver that is bound to the device, an implementation of the  `DeviceInfoProvider` interface is created, which provides more DeviceSpecs.

After, when Allocate() is called,  the `NetResourcePool` has some logic to append these deviceSpecs.

This imposes a limitation on how flexibly we can choose what DeviceSpecs are exposed (must be determined by the driver) and fosters ad-hoc DeviceSpec-appending logic (see how vhostNet DeviceSpec was added).

In order to improve this, I'd like to propose the generalization of DeviceInfoProviders for deviceSpec selection. This PR first allows the PciDevice to be built based on a list of DeviceInfoProviders. The appending logic is moved to the pciDevice getters (e.g: `pciDevice.GetDeviceSpec()`) and the selection of what DeviceInfoProviders to pass is DeviceType specific.

Finally, Rdma and VhostNet deviceSpec appending are implemented using DeviceInfoProviders.
Note that the `DeviceInfoProvider.getDeviceSpecs()` is called after the device have been filtered. Therefore, it can generate errors if it fails to generate the DeviceSpecs correctly.

### DIFF
--- a/pkg/netdevice/netInfoProviders.go
+++ b/pkg/netdevice/netInfoProviders.go
@@ -56,3 +56,33 @@ func (rip *rdmaInfoProvider) GetEnvVal(pciAddr string) string {
 func (rip *rdmaInfoProvider) GetMounts(pciAddr string) []*pluginapi.Mount {
 	return nil
 }
+
+/*
+   VhostNetInfoProvider wraps any DeviceInfoProvider and adds a vhost-net device
+*/
+type vhostNetInfoProvider struct {
+}
+
+// NewVhostNetInfoProvider returns a new Vhost Information Provider
+func NewVhostNetInfoProvider() types.DeviceInfoProvider {
+	return &vhostNetInfoProvider{}
+}
+
+// *****************************************************************
+/* DeviceInfoProvider Interface */
+
+func (rip *vhostNetInfoProvider) GetDeviceSpecs(pciAddr string) []*pluginapi.DeviceSpec {
+	if !VhostNetDeviceExist() {
+		glog.Errorf("GetDeviceSpecs(): /dev/vhost-net doesn't exist")
+		return nil
+	}
+	return GetVhostNetDeviceSpec()
+}
+
+func (rip *vhostNetInfoProvider) GetEnvVal(pciAddr string) string {
+	return pciAddr
+}
+
+func (rip *vhostNetInfoProvider) GetMounts(pciAddr string) []*pluginapi.Mount {
+	return nil
+}
--- a/pkg/netdevice/netInfoProviders.go
+++ b/pkg/netdevice/netInfoProviders.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2020 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package netdevice
+
+import (
+	"github.com/golang/glog"
+
+	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
+
+	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
+)
+
+/*
+   rdmaInfoProvider provides the RDMA information
+*/
+type rdmaInfoProvider struct {
+	rdmaSpec types.RdmaSpec
+}
+
+// NewRdmaInfoProvider returns a new Rdma Information Provider
+func NewRdmaInfoProvider(rdmaSpec types.RdmaSpec) types.DeviceInfoProvider {
+	return &rdmaInfoProvider{
+		rdmaSpec: rdmaSpec,
+	}
+}
+
+// *****************************************************************
+/* DeviceInfoProvider Interface */
+
+func (rip *rdmaInfoProvider) GetDeviceSpecs(pciAddr string) []*pluginapi.DeviceSpec {
+	if !rip.rdmaSpec.IsRdma() {
+		glog.Errorf("GetDeviceSpecs(): rdma is required in the configuration but the device %s is not rdma device", pciAddr)
+		return nil
+	}
+	return rip.rdmaSpec.GetRdmaDeviceSpec()
+}
+
+func (rip *rdmaInfoProvider) GetEnvVal(pciAddr string) string {
+	return pciAddr
+}
+
+func (rip *rdmaInfoProvider) GetMounts(pciAddr string) []*pluginapi.Mount {
+	return nil
+}
--- a/pkg/netdevice/netInfoProviders.go
+++ b/pkg/netdevice/netInfoProviders.go
@@ -41,19 +41,19 @@ func NewRdmaInfoProvider(rdmaSpec types.RdmaSpec) types.DeviceInfoProvider {
 // *****************************************************************
 /* DeviceInfoProvider Interface */
 
-func (rip *rdmaInfoProvider) GetDeviceSpecs(pciAddr string) []*pluginapi.DeviceSpec {
+func (rip *rdmaInfoProvider) GetDeviceSpecs() []*pluginapi.DeviceSpec {
 	if !rip.rdmaSpec.IsRdma() {
-		glog.Errorf("GetDeviceSpecs(): rdma is required in the configuration but the device %s is not rdma device", pciAddr)
+		glog.Errorf("GetDeviceSpecs(): rdma is required in the configuration but the device is not rdma device")
 		return nil
 	}
 	return rip.rdmaSpec.GetRdmaDeviceSpec()
 }
 
-func (rip *rdmaInfoProvider) GetEnvVal(pciAddr string) string {
-	return pciAddr
+func (rip *rdmaInfoProvider) GetEnvVal() string {
+	return ""
 }
 
-func (rip *rdmaInfoProvider) GetMounts(pciAddr string) []*pluginapi.Mount {
+func (rip *rdmaInfoProvider) GetMounts() []*pluginapi.Mount {
 	return nil
 }
 
@@ -71,7 +71,7 @@ func NewVhostNetInfoProvider() types.DeviceInfoProvider {
 // *****************************************************************
 /* DeviceInfoProvider Interface */
 
-func (rip *vhostNetInfoProvider) GetDeviceSpecs(pciAddr string) []*pluginapi.DeviceSpec {
+func (rip *vhostNetInfoProvider) GetDeviceSpecs() []*pluginapi.DeviceSpec {
 	if !VhostNetDeviceExist() {
 		glog.Errorf("GetDeviceSpecs(): /dev/vhost-net doesn't exist")
 		return nil
@@ -79,10 +79,10 @@ func (rip *vhostNetInfoProvider) GetDeviceSpecs(pciAddr string) []*pluginapi.Dev
 	return GetVhostNetDeviceSpec()
 }
 
-func (rip *vhostNetInfoProvider) GetEnvVal(pciAddr string) string {
-	return pciAddr
+func (rip *vhostNetInfoProvider) GetEnvVal() string {
+	return ""
 }
 
-func (rip *vhostNetInfoProvider) GetMounts(pciAddr string) []*pluginapi.Mount {
+func (rip *vhostNetInfoProvider) GetMounts() []*pluginapi.Mount {
 	return nil
 }
--- a/pkg/netdevice/netResourcePool_test.go
+++ b/pkg/netdevice/netResourcePool_test.go
@@ -48,7 +48,7 @@ var _ = Describe("NetResourcePool", func() {
 		})
 	})
 	Describe("getting DeviceSpecs", func() {
-		Context("for non-RDMA devices", func() {
+		Context("for multiple devices", func() {
 			rf := factory.NewResourceFactory("fake", "fake", true)
 			nadutils := rf.GetNadUtils()
 			rc := &types.ResourceConfig{
@@ -60,29 +60,25 @@ var _ = Describe("NetResourcePool", func() {
 			}
 			devs := map[string]*v1beta1.Device{}
 
-			rdmaNo := &mocks.RdmaSpec{}
-			rdmaNo.On("IsRdma").Return(false)
-
 			// fake1 will have 2 device specs
 			fake1 := &mocks.PciNetDevice{}
 			fake1ds := []*pluginapi.DeviceSpec{
 				&pluginapi.DeviceSpec{ContainerPath: "/fake/path", HostPath: "/dev/fake1a"},
 				&pluginapi.DeviceSpec{ContainerPath: "/fake/path", HostPath: "/dev/fake1b"},
 			}
-			fake1.On("GetDeviceSpecs").Return(fake1ds).On("GetRdmaSpec").Return(rdmaNo)
+			fake1.On("GetDeviceSpecs").Return(fake1ds)
 
 			// fake2 will have 1 device spec
 			fake2 := &mocks.PciNetDevice{}
 			fake2ds := []*pluginapi.DeviceSpec{
 				&pluginapi.DeviceSpec{ContainerPath: "/fake/path", HostPath: "/dev/fake2"},
 			}
-			fake2.On("GetDeviceSpecs").Return(fake2ds).
-				On("GetRdmaSpec").Return(rdmaNo)
+			fake2.On("GetDeviceSpecs").Return(fake2ds)
 
 			// fake3 will have 0 device specs
 			fake3 := &mocks.PciNetDevice{}
 			fake3ds := []*pluginapi.DeviceSpec{}
-			fake2.On("GetDeviceSpecs").Return(fake3ds).On("GetRdmaSpec").Return(rdmaNo)
+			fake3.On("GetDeviceSpecs").Return(fake3ds)
 
 			pcis := map[string]types.PciDevice{"fake1": fake1, "fake2": fake2, "fake3": fake3}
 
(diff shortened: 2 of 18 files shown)


pkg/netdevice/netInfoProviders.go
```diff
@@ -0,0 +1,58 @@
...
+*/
+
+package netdevice
+
+import (
+	"github.com/golang/glog"
+
+	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
+
+	"github.com/intel/sriov-network-device-plugin/pkg/types"
```
mcoquelin, line 24: Depending on whether your series or mine (https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/280) is merged first. I will either have to change this import to NPWG's repo, or you'll have to change to NPWG's.

amorenoz: This one will likely take longer to review, so no worries. Thanks for the heads up

amorenoz:
This is a preliminary rework to make it easier to implement vdpa support, as it makes sense to implement it using a `DeviceInfoProvider`
@zshi-redhat @adrianchiris @martinkennelly PTAL

zshi-redhat:
> This is a preliminary rework to make it easier to implement vdpa support, as it makes sense to implement it using a `DeviceInfoProvider`
> @zshi-redhat @adrianchiris @martinkennelly PTAL

@amorenoz sorry for getting to this late.

About adding the rdmaInfoProvider, would it make sense to remove current pkg/netdevice/rdma.go and re-use the rdma info provider file? I see there are two reasons:
1) Both files contain similar data such as rdma device spec. 
2) For other info providers (vfio, uio etc), devSpec is retrieved in GetDeviceSpec function while rdma device spec is retrieved in the NewRdmaSpec funciton.

This may require a change in pciNetDevice struct to replace rdmaSpec wth DeviceInfoProvider.

amorenoz:
> About adding the rdmaInfoProvider, would it make sense to remove current pkg/netdevice/rdma.go and re-use the rdma info provider file? I see there are two reasons:
> 
>     1. Both files contain similar data such as rdma device spec.
> 

Clearly there is a bit of a name clash with the Rdma bits.
You have a very good point. I thought about this as well and reached the conclusion that having structs within the pciNetDevice that can contain more information than what the DeviceInfoProvider "provides" might be useful.
For instance, you might want to store information such as the rdma device name that will end up in the DeviceInfo File as per the recently approved Spec. If we store only the DeviceSpec, we would have to convert it back and might loose some info in the way.
Another example is vDPA + virtio_vdpa. The DeviceSpec is actually empty (the virtio netdevice will be moved to the Pod's namespace), however, vdpa-related information has to end up in the DeviceInfoFile and reach the CNI.

>     2. For other info providers (vfio, uio etc), devSpec is retrieved in GetDeviceSpec function while rdma device spec is retrieved in the NewRdmaSpec funciton.
> 

Another argument is that having the possibility of retrieving complex information in discovery phase (as opposed to GetDeviceSpec() which is "runtime" phase) might also be useful.
It's true that the rdma spec information retrieval is simple enough to be moved to runtime, but what if it get's more complicated? For example, the current proposal to implement vDPA - DPDK is to have the DevicePlugin make a json-rpc call to retrieve the vhost-user socket information. I think we should avoid doing that at runtime if possible.

> 
> This may require a change in pciNetDevice struct to replace rdmaSpec wth DeviceInfoProvider.

A list of DeviceInfoProvider is already in pciDevice struct, so it's all about whether we need the extra info or not.

adrianchiris:
As a general comment, today rdma device spec is created when PciNetDevice is allocated and statically retrieved on `netResourcePool.GetDeviceSpecs()` so there was no runtime operation to retrieve RDMA mounts before this PR.

Currently `pciNetDevice.rdmaSpec` is used for:

1. filter devices if `isRdma` is set to true in resource config
2. provide device spec on `netResourcePool.GetDeviceSpecs()` call if needed  (if `isRdma` is set to true in resource config)

this PR affects 2. as it moves the responsibility to provide this information by `rdmaInfoProvider`

perhaps for 1. we should just keep a boolean flag in PCINetDevice so we know this device supports RDMA ? 
or a list of "DeviceAttributes" which would just be a list of consts if we think it may have additional attributes to be queried (e.g vhostnet)

`DeviceInfoProvider` interface can later be extended to provide additional information for DeviceInfo File.

amorenoz:
> As a general comment, today rdma device spec is created when PciNetDevice is allocated and statically retrieved on `netResourcePool.GetDeviceSpecs()` so there was no runtime operation to retrieve RDMA mounts before this PR.
> 
> Currently `pciNetDevice.rdmaSpec` is used for:
> 
>     1. filter devices if `isRdma` is set to true in resource config
> 
>     2. provide device spec on `netResourcePool.GetDeviceSpecs()` call if needed  (if `isRdma` is set to true in resource config)
> 
> 
> this PR affects 2. as it moves the responsibility to provide this information by `rdmaInfoProvider`
> 
> perhaps for 1. we should just keep a boolean flag in PCINetDevice so we know this device supports RDMA ?
> or a list of "DeviceAttributes" which would just be a list of consts if we think it may have additional attributes to be queried (e.g vhostnet)
> `DeviceInfoProvider` interface can later be extended to provide additional information for DeviceInfo File.

IIUC what you're suggesting is to only store the Rdma DeviceSpec in the DeviceInfoProvider and the flag (isRdma) in the pciNetDevice. For vDPA this would mean store what vdpa-driver is used in the pciNetDevice.
Having these pieces of information split between InfoProvider and pciNetDevice would make obtaining a coherent DeviceInfo quite fuzzy don't you think? Besides, DeviceSpecs can be appended but DeviceInfo(s) cannot.

If your concern is the duplicated information, how about the following:

- We store all the information in pciNetDevice (for rdma, isRdma and the output of `rdmamap.GetRdmaDevicesForPcidev(pciAddrs)`)
- All the information is ready to be used for filtering
- Instead of having an instance of rdmaInfoProvider per device, we make it static and accept a pciNetDevice as parameter. Then each InfoProvider has access to the full object. E.g:

```
type pciNetDevice struct {
	types.PciDevice
	ifName    string
	pfName    string
	linkSpeed string
	rdmaInfo  RdmaInfo
	linkType  string
}
type RdmaInfo struct {
	IsRdma      bool
	RdmaDevices []string
}
func (nd *pciNetDevice) GetRdma() RdmaInfo {
	return nd.rmaInfo
}
//...//
func (rip *rdmaInfoProvider) GetDeviceSpecs(dev types.PciDevice) []*pluginapi.DeviceSpec {
	deviceSpec := make([]*pluginapi.DeviceSpec, 0)
	netdev, _ := dev.(types.PciNetDevice)

	for _, rdmaDevice := range netdev.GetRdma().RdmaDevices {
		deviceSpec = append(deviceSpec, &pluginapi.DeviceSpec{
			HostPath:      rdmaDevice,
			ContainerPath: rdmaDevice,
			Permissions:   "rwm",
		})
	}
	return deviceSpec
}
```

zshi-redhat:
> > `DeviceInfoProvider` interface can later be extended to provide additional information for DeviceInfo File.

I think we need to define whether InfoProvider should be used only for getting Device Plugin API info such as Env/Mount/devSpec. I thought it was but we didn't call it out explicitly, no?

> 
>     * Instead of having an instance of rdmaInfoProvider per device, we make it static and accept a pciNetDevice as parameter. Then each InfoProvider has access to the full object. E.g:
> 
This looks clear that infoProvider is somehow separated from pciDev or pciNetdev.
How should we handle the case where one dev requires multiple infoProviders?
> 
> 
> That way, the InfoProvider will have the full picture of the device it would be in a better position to create the DeviceInfo. This approach does not play nice with appending InfoProviders though.

Yeah, that's a question for making above changes.

amorenoz:
> > > `DeviceInfoProvider` interface can later be extended to provide additional information for DeviceInfo File.
> 
> I think we need to define whether InfoProvider should be used only for getting Device Plugin API info such as Env/Mount/devSpec. I thought it was but we didn't call it out explicitly, no?
The way I see it, InfoProviders should only be used for Env/Mount/devSpecs

IMHO, DeviceInfo files should be populated using the whole pciNetDevice.

That's why I originally did not go ahead with your first proposal of removing the rdma information from pciNetDevice and moving it to the InfoProvider

adrianchiris:
> IIUC what you're suggesting is to only store the Rdma DeviceSpec in the DeviceInfoProvider and the flag (isRdma) in the pciNetDevice. For vDPA this would mean store what vdpa-driver is used in the pciNetDevice.
Having these pieces of information split between InfoProvider and pciNetDevice would make obtaining a coherent DeviceInfo quite fuzzy don't you think?

im not sure about vDPA as we still dont support it in device plugin, what im saying is that pciNetDevice should only store information thats being used (the source of this information can be common ofc, e.g for RDMA, rdmaSpec). About coherency, I think it is the responsibility of the creator of the pciNetDevice object to create it in a coherent manner so that later a DeviceInfo file may be composed.

> Besides, DeviceSpecs can be appended but DeviceInfo(s) cannot. 

Correct , but it can be composed by merging (i.e merge of maps).

> If your concern is the duplicated information,

thats not my concern, and the original approach in this PR is OK IMO, my concern is: "pciNetDevice should only store information that's being used (the source of this information can be common ofc, e.g for RDMA, rdmaSpec)."

> think we need to define whether InfoProvider should be used only for getting Device Plugin API info such as Env/Mount/devSpec. I thought it was but we didn't call it out explicitly, no?

My original though was to extend the interface to provide information for Device Information file creation.
Thinking on this more i agree it should be separate. `DeviceInfoProvider` should be limited to _kubelet_ related Device plugin APIs.

> IMHO, DeviceInfo files should be populated using the whole pciNetDevice.

Agree.


We should really focus on the changes in this PR and not how DeviceInfo file would be created, it was not my intention to trigger a discussion on it but i understand that this drives some of the design decisions here.


To summarize, from my POV in regards to this PR:
Im OK with keeping `pciNetDevice.rdmaSpec`

amorenoz:
> 
> To summarize, from my POV in regards to this PR:
> 
>     * Im OK with keeping `pciNetDevice.rdmaSpec`
> 
OK. @zshi-redhat, WDYT about this?

>     * InfoProvider should be aggregated as a list and be and member of PciDevice so they may be used in a generic manner to provide information for k8s device plugin API

@adrianchiris: I believe I have implemented it as you suggest:
https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/281/commits/62d7bcd19c199c4f93cdd16ab7a46b714165e3fb#diff-64650e8db2ae468b3bde641746773b58d3a9bd369269af83ee16dc65207640b8

zshi-redhat:
> > To summarize, from my POV in regards to this PR:
> > ```
> > * Im OK with keeping `pciNetDevice.rdmaSpec`
> > ```
> 
> OK. @zshi-redhat, WDYT about this?
> 

I'm fine with this move.

nit: do we want to rename the default info providers (vfioResource, uioResource) to use consistent names such as `uioInfoProvider` and `vfioInfoProvider` ? or maybe even add `API` explicitly in the name (`uioApiInfoProvider`)

adrianchiris:
> nit: do we want to rename the default info providers (vfioResource, uioResource) to use consistent names such as uioInfoProvider and vfioInfoProvider

+1 although no directly related to this PR but i wouldn't mind :)  also while we are at it also rename `genericResource` to `GenericInfoProvider` as they all implement the same interface.

pkg/netdevice/netInfoProviders.go
amorenoz: Rebased

pkg/netdevice/pciNetDevice_test.go
```diff
@@ -18,10 +18,12 @@ import (
 	"testing"
 
 	"github.com/jaypipes/ghw"
+	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
 
 	"github.com/intel/sriov-network-device-plugin/pkg/factory"
 	"github.com/intel/sriov-network-device-plugin/pkg/netdevice"
 	"github.com/intel/sriov-network-device-plugin/pkg/types"
+	"github.com/intel/sriov-network-device-plugin/pkg/types/mocks"
```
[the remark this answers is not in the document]
amorenoz: Done. Thanks

amorenoz:
> > nit: do we want to rename the default info providers (vfioResource, uioResource) to use consistent names such as uioInfoProvider and vfioInfoProvider
> 
> +1 although no directly related to this PR but i wouldn't mind :) also while we are at it also rename `genericResource` to `GenericInfoProvider` as they all implement the same interface.

Thanks @martinkennelly, @zshi-redhat and @adrianchiris for your comments. Updated PR with GenericResource->GenericInfoProvider renaming

adrianchiris (commented):
I still need to have a deeper look at the unit-tests.

Overall looks good,
One quirk that caught my eye is that we now have DeviceInfoProvider implementations that dont really use the pciAddr for `GetDeviceSpecs`, `GetEnvVal`, `GetMounts` (i.e vhost, rdma).

Also DeviceInfoProviders are created per pciDevice so i wonder if it would be more correct to roll the `pciAddr` parameter (and in the future maybe other parameters) into the impl via `New<providerName>` method and save it as a struct attribute when needed, then we would have:

```
type DeviceInfoProvider interface {
	GetEnvVal() string
}
```

amorenoz:
> I still need to have a deeper look at the unit-tests.
> 
> Overall looks good,
> One quirk that caught my eye is that we now have DeviceInfoProvider implementations that dont really use the pciAddr for `GetDeviceSpecs`, `GetEnvVal`, `GetMounts` (i.e vhost, rdma).
> 
> Also DeviceInfoProviders are created per pciDevice so i wonder if it would be more correct to roll the `pciAddr` parameter (and in the future maybe other parameters) into the impl via `New<providerName>` method and save it as a struct attribute when needed, then we would have:
> 
> ```
> type DeviceInfoProvider interface {
> 	GetDeviceSpecs() []*pluginapi.DeviceSpec
> 	GetEnvVal() string
> 	GetMounts() []*pluginapi.Mount
> }
> ```

Yep, that looks cleaner, will update the PR. Thanks

adrianchiris:
Thanks you @amorenoz  for addressing my comments,
added commit looks good IMO.
Just left with reviewing unit tests i will do my best to get to it early next week !

if you could rebase the PR as there are conflicts, would be great!

pkg/netdevice/pciNetDevice_test.go
```diff
@@ -121,6 +124,133 @@ var _ = Describe("PciNetDevice", func() {
...
+			rdma1.On("IsRdma").Return(true).On("GetRdmaDeviceSpec").Return(fake1ds)
+			// fake2 will have 0 rdma device specs to trigger error msg
+			rdma2 := &mocks.RdmaSpec{}
+			fake2ds := []*pluginapi.DeviceSpec{
+				&pluginapi.DeviceSpec{ContainerPath: "/fake/path", HostPath: "/dev/fake2"},
+			}
+			rdma2.On("IsRdma").Return(false).On("GetRdmaDeviceSpec").Return(fake2ds)
+
+			f := &mocks.ResourceFactory{}
+			f.On("GetDefaultInfoProvider", AnythingOfType("string"), "vfio-pci").Return(func(addr string, driver string) types.DeviceInfoProvider {
```
adrianchiris, line 172: I think it would be cleaner to create a mock.DeviceInfoProvider{} and use it here instead of using `vfio-pci`
as we dont really care in this test (and below test) for the default info provider right ?

you can then remove vfio-pci sysfs related stuff and simplify (abit) the test. WDYT ?

pkg/netdevice/pciNetDevice_test.go
```diff
@@ -121,6 +124,133 @@ var _ = Describe("PciNetDevice", func() {
...
+				&pluginapi.DeviceSpec{ContainerPath: "/fake/path", HostPath: "/dev/fake1a"},
+				&pluginapi.DeviceSpec{ContainerPath: "/fake/path", HostPath: "/dev/fake1b"},
+			}
+			rdma1.On("IsRdma").Return(true).On("GetRdmaDeviceSpec").Return(fake1ds)
+			// fake2 will have 0 rdma device specs to trigger error msg
+			rdma2 := &mocks.RdmaSpec{}
+			fake2ds := []*pluginapi.DeviceSpec{
+				&pluginapi.DeviceSpec{ContainerPath: "/fake/path", HostPath: "/dev/fake2"},
+			}
+			rdma2.On("IsRdma").Return(false).On("GetRdmaDeviceSpec").Return(fake2ds)
```
adrianchiris, line 169: if `IsRdma` is false , it should not have RDMA deviceSpecs IMO.

pkg/netdevice/pciNetDevice_test.go
```diff
@@ -121,6 +124,133 @@ var _ = Describe("PciNetDevice", func() {
...
+			in1 := &ghw.PCIDevice{Address: "0000:00:00.1"}
+			in2 := &ghw.PCIDevice{Address: "0000:00:00.2"}
+
+			It("should populate Rdma device specs if isRdma", func() {
+				defer fs.Use()()
+				defer utils.UseFakeLinks()()
+				dev, err := netdevice.NewPciNetDevice(in1, f, rc)
+
+				Expect(dev.GetDriver()).To(Equal("vfio-pci"))
+				Expect(dev.GetNetName()).To(Equal("eth0"))
```
adrianchiris, line 188: nit: if the driver is vfio-pci we would not have a netName nor RDMA resources.
thats just semantics but i would prefer if you changed the driver name to reflect a real-life scenario.

pkg/netdevice/pciNetDevice_test.go
```diff
@@ -121,6 +124,133 @@ var _ = Describe("PciNetDevice", func() {
...
+			f := factory.NewResourceFactory("fake", "fake", true)
+			in := &ghw.PCIDevice{Address: "0000:00:00.1"}
+			It("should add the vhost-net deviceSpec", func() {
+				defer fs.Use()()
+				defer utils.UseFakeLinks()()
+
+				dev, err := netdevice.NewPciNetDevice(in, f, rc)
+
+				Expect(dev.GetDriver()).To(Equal("vfio-pci"))
+				Expect(dev.GetNetName()).To(Equal("eth0"))
```
adrianchiris, line 243: if driver is vfio-pci we would not have a netdev name right ?

pkg/netdevice/pciNetDevice_test.go
amorenoz: Done

pkg/netdevice/pciNetDevice_test.go
amorenoz: Agree. Better show real cases

pkg/netdevice/netResourcePool_test.go
```diff
@@ -92,49 +92,5 @@ var _ = Describe("NetResourcePool", func() {
 				Expect(actual).To(ContainElement(fake2ds[0]))
 			})
 		})
-		Context("for RDMA devices", func() {
```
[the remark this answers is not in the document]
amorenoz: Done

pkg/netdevice/pciNetDevice_test.go
amorenoz: Done

pkg/netdevice/pciNetDevice_test.go
amorenoz: I agree, we don't care of what they return. We do, however, care that they are called. Right?

amorenoz: Updated: Used a mock and checked it's being called
WDYT?

adrianchiris: Perfect !

adrianchiris (approved):
Thank you for this contribution @amorenoz  !
LGTM

martinkennelly (approved):
/LGTM!

PR #285, opened 2020-11-18
Remove hostPID cap from deployment examples

adrianchiris:
This commit removes the HostPID cap from deployment
examples as it is not needed for proper operation
of device plugin.

Signed-off-by: Adrian Chiris <adrianc@nvidia.com>


zshi-redhat (approved):

martinkennelly (approved):
LGTM

Issue #286, opened 2020-11-19
Add Device Info Spec support


amorenoz:
## What would you like to be added?
The Network Plumbing Working Group approved an extension spec called [Device Information Specification](https://github.com/k8snetworkplumbingwg/device-info-spec/) that defines how device information can be shared between Device Plugins, CNI Plugins and Pods.

The part of the spec that affects the SR-IOV Network Device Plugin is fairly simple: the Device Plugin should store the device information in a file on a specified path and take care of removing such files on termination.
Luckily, a good part of the spec details have already been implemented in the [network-attachment-definition-client library](https://github.com/k8snetworkplumbingwg/network-attachment-definition-client).

## What is the use case for this feature / enhancement?
All the [usecases that the Device Information Spec enables](https://github.com/k8snetworkplumbingwg/device-info-spec/#use-cases)

PR #287, opened 2020-11-19
Add support for Device Information Spec

amorenoz:
As per the device-info specification, save the device information in a file using [github.com/k8snetworkplumbingwg/network-attachment-definition-client
](https://github.com/k8snetworkplumbingwg/network-attachment-definition-client) utility functions.

As per the design decisions made: Added the create/clean functions to the ResourcePool Interface and implement them in the netResourcePool so that the entire device information can be accessed

Error handling:
- Startup: Do not stop the server creation if the DeviceInfo file creation fails, but log it at Error level
- Cleanup: Do not stop cleaning if one DeviceInfo file deletion fails, try to clean them all and aggregate errors. 

For now, only add the mandatory fields.

Fixes: #286


amorenoz:
cc @Billy99

amorenoz:
With this change, having the following pools:

```
$ kubectl get node $HOSTNAME -o json | jq .status.allocatable
{
  "cpu": "8",
  "ephemeral-storage": "3570039515370",
  "hugepages-1Gi": "0",
  "hugepages-2Mi": "16Gi",
  "intel.com/intelVFs": "3",
  "intel.com/mlxPFs": "1",
  "intel.com/mlxVFs": "4",
  "memory": "15908252Ki",
  "pods": "110"
}
```

I get the following files:
```
$ tree /var/run/k8s.cni.cncf.io/devinfo/dp
/var/run/k8s.cni.cncf.io/devinfo/dp
├── intel.com-intelVFs-0000:05:00.2-device.json
├── intel.com-intelVFs-0000:05:00.3-device.json
├── intel.com-intelVFs-0000:05:00.4-device.json
├── intel.com-mlxPFs-0000:04:00.1-device.json
├── intel.com-mlxVFs-0000:04:00.3-device.json
├── intel.com-mlxVFs-0000:04:00.4-device.json
├── intel.com-mlxVFs-0000:04:00.5-device.json
└── intel.com-mlxVFs-0000:04:00.6-device.json
$ cat /var/run/k8s.cni.cncf.io/devinfo/dp/intel.com-mlxVFs-0000\:04\:00.3-device.json 
{"type":"pci","version":"0.1.0","pci":{"pci-address":"0000:04:00.3"}}
$ cat /var/run/k8s.cni.cncf.io/devinfo/dp/intel.com-intelVFs-0000\:05\:00.3-device.json 
{"type":"pci","version":"0.1.0","pci":{"pci-address":"0000:05:00.3"}}
```

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
 	github.com/jaypipes/ghw v0.6.0
 	github.com/jaypipes/pcidb v0.5.0
+	github.com/k8snetworkplumbingwg/network-attachment-definition-client v1.1.1-0.20201110193115-68e49246082c
```
Billy99, line 10: I know this is still WIP, but just letting you know that I push a few PRs since this. Just added some constant strings to the types.go. But may be worth moving so all the repos are on the same version.

pkg/netdevice/netResourcePool.go
```diff
@@ -15,8 +15,12 @@
 package netdevice
 
 import (
+	"fmt"
 	"github.com/golang/glog"
+	"strings"
 
+	nadtypes "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+	nadutils "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/utils"
```
Billy99, line 23: I used nettypes and nadutils based in multus and app-netutil (based on your POC) in case you want to be consistent across repos (I tried to be consistent between the two I modified).

pkg/netdevice/netResourcePool.go
```diff
@@ -80,3 +84,42 @@ func (rp *netResourcePool) GetDeviceSpecs(deviceIDs []string) []*pluginapi.Devic
...
+//  k8snetworkplumbingwg/device-info-spec
+func (rp *netResourcePool) StoreDeviceInfoFile(resourceNamePrefix string) error {
+	for id, dev := range rp.GetDevicePool() {
+		netDev, ok := dev.(types.PciNetDevice)
+		if !ok {
+			return fmt.Errorf("StoreDeviceInfoFile: Only pciNetDevices are supported")
+		}
+		devInfo := nadtypes.DeviceInfo{
+			Type:    "pci",
+			Version: "0.1.0",
```
Billy99, line 98: I added a `nettypes.DeviceInfoVersion` so that it always updates the version based on what version of network-attachment-definition-client the code is compiling against.

Billy99 (approved):
/lgtm

None of the comments are real issues, just nits. I tested with the version just before the PR was created and worked fine with pending device-info Multus PR. The PR comment indicated that only the mandatory fields were filled in. Is that the plan for the final PR or just because this is still [WIP] (just wondering, not pressuring to fill in the rest)?

pkg/netdevice/netResourcePool.go
zshi-redhat: +1 on using defined const.

go.mod
amorenoz: Done

pkg/netdevice/netResourcePool.go
amorenoz: Done

pkg/netdevice/netResourcePool.go
amorenoz: Done

amorenoz:
> /lgtm
> 
> None of the comments are real issues, just nits. I tested with the version just before the PR was created and worked fine with pending device-info Multus PR. The PR comment indicated that only the mandatory fields were filled in. Is that the plan for the final PR or just because this is still [WIP] (just wondering, not pressuring to fill in the rest)?

I was thinking in pushing the mandatory fields first (pci-address) mainly because how the rest of the device info is handled is being discussed in #281

pkg/netdevice/netResourcePool.go
```diff
@@ -80,3 +84,42 @@ func (rp *netResourcePool) GetDeviceSpecs(deviceIDs []string) []*pluginapi.Devic
...
+// StoreDeviceInfoFile stores the Device Info files according to the
+//  k8snetworkplumbingwg/device-info-spec
+func (rp *netResourcePool) StoreDeviceInfoFile(resourceNamePrefix string) error {
+	for id, dev := range rp.GetDevicePool() {
+		netDev, ok := dev.(types.PciNetDevice)
+		if !ok {
+			return fmt.Errorf("StoreDeviceInfoFile: Only pciNetDevices are supported")
+		}
+		devInfo := nettypes.DeviceInfo{
+			Type:    "pci",
```
zshi-redhat, line 97: should this also use the const defined in net-attach-def-client for "pci" Type?

zshi-redhat (commented):
looks like CI failed, is it because due to not using 1.15 golang version?

amorenoz:
> looks like CI failed, is it because due to not using 1.15 golang version?

I'm not sure why it's failing. I'm looking into it but it's taking some time since I get errors in Travis that I cannot reproduce locally.

pkg/netdevice/netResourcePool.go
amorenoz: Done. Thanks

amorenoz:
> > looks like CI failed, is it because due to not using 1.15 golang version?
> 
> I'm not sure why it's failing. I'm looking into it but it's taking some time since I get errors in Travis that I cannot reproduce locally.

@zshi-redhat, the CI instability issues I was facing are now solved. PTAL

martinkennelly (approved):
/LGTM. Performed as expected.

deployments/k8s-v1.16/sriovdp-daemonset.yaml
```diff
@@ -111,13 +116,18 @@ spec:
...
+          mountPath: /var/run/k8s.cni.cncf.io/devinfo/dp
       volumes:
         - name: devicesock
           hostPath:
             path: /var/lib/kubelet/
         - name: log
           hostPath:
             path: /var/log
+        - name: device-info
+          hostPath:
```
adrianchiris, line 129: do you need to specify
`DirectoryOrCreate` type ?

what if the path does not exist on the host ? will it be created ?

https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

amorenoz: I did not see any error when I tested it. The nad library [would take care of creating the directory](https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/blob/9d213757d22d5d06f1b8ff289ce16d0cf9aabc6c/pkg/utils/cniconfig.go#L163) so from a DevicePlugin perspective, it should not care. However, it's not a bad idea to add that type flag.
Thanks

adrianchiris (commented):
IMO changes are good. I also tested the change locally to get a better feel for the new behaviour.

Are you planning to add unit test for {Store/Clean}DeviceInfoFile` ?

`nadutils` can be put behind an interface to allow mocking

p.s
regarding my last comment indeed it seems to work when `DirectoryOrCreate` is not specified.
but no harm in keeping it IMO.

amorenoz:
> IMO changes are good. I also tested the change locally to get a better feel for the new behaviour.
> 
> Are you planning to add unit test for {Store/Clean}DeviceInfoFile` ?
> 
The current unit test verifies that Store/CleanDeviceInfoFile are called but not what it does internally

> `nadutils` can be put behind an interface to allow mocking
> 
Good idea, I did not think of that option. So we would have to:
- Add an interface in types.go with the functions from 'nadutils' we use
- Implement it (for example) in pkg/utils/utils.go
- Add a function to the factory to create the implementation
- Make the `resourceManager` pass the factory to `server.Start()` and `server.Stop()` (boy, we need to make this factory a proper singleton!)

Are we good with this approach?

amorenoz:
@martinkennelly, @adrianchiris, Talking to @zshi-redhat, we considered an alternative approach in order to make the unit tests cleaner: make `resourceFactory` a global singleton. It's anyway referenced all over the place and it makes little sense to carry it's pointer around.
However, that is a big change that would likely miss the upcoming release.
So the strategy would be to merge this PR as is, make the release, and then I'll work on making the `resourceFactory` global. Once that is done, I'll add Unit tests for the `Store/CleanDeviceInfoFile`

What do you think?

adrianchiris:
>Good idea, I did not think of that option. So we would have to:
>
>- Add an interface in types.go with the functions from 'nadutils' we use
>- Implement it (for example) in pkg/utils/utils.go
>- Add a function to the factory to create the implementation
>- Make the resourceManager pass the factory to server.Start() and server.Stop() (boy, we need to make this factory a proper >singleton!)

`NetResourcePool` would either need the factory or an implementation to the "`nadutils`" interface. i dont think you need to pass it to server.Start(), server.Stop().

>Are we good with this approach?

Generally Yes from my POV, See my comments below on the proper singleton bit.

>@martinkennelly, @adrianchiris, Talking to @zshi-redhat, we considered an alternative approach in order to make the unit tests cleaner: make resourceFactory a global singleton. It's anyway referenced all over the place and it makes little sense to carry it's pointer around.
However, that is a big change that would likely miss the upcoming release.
So the strategy would be to merge this PR as is, make the release, and then I'll work on making the resourceFactory global. Once that is done, I'll add Unit tests for the Store/CleanDeviceInfoFile

resourceFactory is a signleton today, its `New` method returns the same instance.
having it "globally" available would require alot of changes in unit-test to swap its global implementation with a mocked one.
as well as code refactoring of the various packages. The benefits of doing so is not apparent to me (at least not in a substantial manner). 

I think having resourceFactory instance as a member of other constructs follows the pattern of [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection), to help with unit tests and ease refactoring/replacing existing implementations of different interfaces. what you propose kinda deviates from that path.

amorenoz:
> > Good idea, I did not think of that option. So we would have to:
> > 
> > * Add an interface in types.go with the functions from 'nadutils' we use
> > * Implement it (for example) in pkg/utils/utils.go
> > * Add a function to the factory to create the implementation
> > * Make the resourceManager pass the factory to server.Start() and server.Stop() (boy, we need to make this factory a proper >singleton!)
> 
> `NetResourcePool` would either need the factory or an implementation to the "`nadutils`" interface. i dont think you need to pass it to server.Start(), server.Stop().
> 
> > Are we good with this approach?
> 
> Generally Yes from my POV, See my comments below on the proper singleton bit.
> 
> > @martinkennelly, @adrianchiris, Talking to @zshi-redhat, we considered an alternative approach in order to make the unit tests cleaner: make resourceFactory a global singleton. It's anyway referenced all over the place and it makes little sense to carry it's pointer around.
> > However, that is a big change that would likely miss the upcoming release.
> > So the strategy would be to merge this PR as is, make the release, and then I'll work on making the resourceFactory global. Once that is done, I'll add Unit tests for the Store/CleanDeviceInfoFile
> 
> resourceFactory is a signleton today, its `New` method returns the same instance.

I don't think that's really the case (although the intention does seem that one):
https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/89b9bc9adbea4df09cf6455e62bfa00ab5ed26c0/pkg/factory/factory.go#L39-L49


> having it "globally" available would require alot of changes in unit-test to swap its global implementation with a mocked one.
> as well as code refactoring of the various packages. The benefits of doing so is not apparent to me (at least not in a substantial manner).
>
> I think having resourceFactory instance as a member of other constructs follows the pattern of [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection), to help with unit tests and ease refactoring/replacing existing implementations of different interfaces. what you propose kinda deviates from that path.
> 
> I think the refactor for `resourceFactory` should be further discussed to reach an agreement, if you could open an issue on it to trigger discussion it would be great ! :)
>
#300 created

> If you think adding unit-test in the current state of code requires alot of code changes im fine with having it as a separate PR.
> (and an issue to track that ofc)

I don't think it's a a lot of code, it just requires modifying the `resourcePool` struct. If that's acceptable I'd go ahead with it.

adrianchiris:
> I don't think that's really the case (although the intention does seem that one):

you are right ! there is no assignment to `instance` although i expected to be one.


> #300 created

Thanks!

> I don't think it's a a lot of code, it just requires modifying the resourcePool struct. If that's acceptable I'd go ahead with it.

IMO its a reasonable change to facilitate UT of the added methods

amorenoz:
@adrianchiris PTAL at the last 3 commits (the rest of the PR unchanged)

zshi-redhat:
CI failed because `pkg/types/mocks/NadUtils.go` file is not generated.

pkg/netdevice/netResourcePool_test.go
```diff
@@ -137,4 +147,55 @@ var _ = Describe("NetResourcePool", func() {
...
+			err_store := rp.StoreDeviceInfoFile("fakeOrg.io")
+			err_clean := rp.CleanDeviceInfoFile("fakeOrg.io")
+
+			It("should call nadutils to create a well formatted DeviceInfo object", func() {
+				nadutils.AssertCalled(GinkgoT(), "SaveDeviceInfoFile", "fakeOrg.io/fakeResource", "fake1", Anything)
+				nadutils.AssertCalled(GinkgoT(), "SaveDeviceInfoFile", "fakeOrg.io/fakeResource", "fake2", Anything)
+				Expect(err_store).ToNot(HaveOccurred())
+			})
+			It("should call nadutils to clean the DeviceInfo objects", func() {
+				fmt.Printf("%+v", nadutils.Calls)
```
adrianchiris, line 194: is this needed in the test ?

pkg/netdevice/netResourcePool_test.go
```diff
@@ -137,4 +147,55 @@ var _ = Describe("NetResourcePool", func() {
...
+					return nil
+				})
+			nadutils.On("SaveDeviceInfoFile", "fakeOrg.io/fakeResource", "fake2", Anything).
+				Return(func(rName, id string, devInfo *nettypes.DeviceInfo) error {
+					if devInfo.Type != nettypes.DeviceInfoTypePCI || devInfo.Pci == nil || devInfo.Pci.PciAddress != "0000:01:00.2" {
+						return fmt.Errorf("wrong device info")
+					}
+					return nil
+				})
+			nadutils.On("CleanDeviceInfoFile", "fakeOrg.io/fakeResource", "fake1").Return(nil)
```
adrianchiris, line 174: nit:  I would move nadutil mock initialization down to the `It` block and only mock whats needed per the specific test.
then you can use : nadutils.AssertExpectations(GinkgoT())

also perhaps add:
`t := GinkgoT()` in `Describe` block so `GinkgoT()` is not being called each time and use `t` where needed

pkg/netdevice/nadutils.go
```diff
@@ -0,0 +1,27 @@
+package netdevice
+
+import (
+	nettypes "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+	nadutils "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/utils"
+
+	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
+)
+
+// nadutils implements types.NADUtils interface
```
adrianchiris, line 10: nit: `types.NadUtils` ?

amorenoz: Done

pkg/netdevice/netResourcePool_test.go
amorenoz: Done

pkg/netdevice/netResourcePool_test.go
amorenoz: Clearly it's a debug leftover. Done

pkg/netdevice/netResourcePool_test.go
amorenoz: Done

adrianchiris (approved):
Thanks @amorenoz  !!

LGTM

martinkennelly (approved):
LGTM. Thanks for the rework / noticing the singleton issue.

pkg/types/types.go
```diff
@@ -213,3 +215,9 @@ type RdmaSpec interface {
 	IsRdma() bool
 	GetRdmaDeviceSpec() []*pluginapi.DeviceSpec
 }
+
+// NadUtils is an interface for Netowork-Attachment-Definition utilities
```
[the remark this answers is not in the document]
zshi-redhat: @amorenoz I think this PR is very close to merge, do you want to fix this minor issue or change it along with future PRs.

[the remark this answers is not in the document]
amorenoz: Sorry, I missed this comment yesterday. Done

zshi-redhat (approved):

Issue #294, opened 2020-11-26
.PHONY targets could all be defined in one line


mrymsza:
## What issue would you like to bring attention to?
.PHONY targets are not defined in one line at the top.

## What is the impact of this issue?
Defining .PHONY targets all over the file makes the Makefile not easy to read.

## Do you have a proposed response or remediation for the issue?
It's probably better practice to keep all .PHONY commands together at the top.

martinkennelly:
Closing issue. Will not implement.

PR #295, opened 2020-11-26
Keep all .PHONY options at the top to increase readability

mrymsza:
Fixes https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/issues/294

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,9 +45,6 @@ TIMEOUT = 15
 V = 0
 Q = $(if $(filter 1,$V),,@)
 
-.PHONY: all
-all: fmt lint build
-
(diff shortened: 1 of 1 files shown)


adrianchiris (commented):
Seems like adding a phony target to `.PHONY` before declaing it is considered best-practice (by some?)

see: https://clarkgrubb.com/makefile-style-guide#phony-targets

im not sure having them all bunched in one place is better, but im not an expert here :)

adrianchiris:
Closing this PR for the same reason as https://github.com/k8snetworkplumbingwg/sriov-cni/pull/158

Issue #291, opened 2020-11-26
makefile GODOC variable is not used


mrymsza:
## What issue would you like to bring attention to?
Makefile contains some unused and unnecessary variables and comments. The layout of the file is also inconsistent (some sections are named others aren't).

## What is the impact of this issue?
Makefile is no longer clear and easy to read.

## Do you have a proposed response or remediation for the issue?
All of the variables should be kept at the top of the file, the comments should be either removed or placed before every section.

PR #298, opened 2020-11-26
Clean up Makefile variables and comments

mrymsza:
Keep all variables at the top of the file
Remove unused and/or unnecessary variables
Delete unnecessary comments

Fixes https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/issues/291

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,14 +11,19 @@ GOPATH=$(CURDIR)/.gopath
 GOBIN=$(CURDIR)/bin
 BUILDDIR=$(CURDIR)/build
 BASE=$(GOPATH)/src/$(REPO_PATH)
-PKGS     = $(or $(PKG),$(shell cd $(BASE) && env GOPATH=$(GOPATH) $(GO) list ./... | grep -v "^$(PACKAGE)/vendor/"))
+PKGS = $(or $(PKG),$(shell cd $(BASE) && env GOPATH=$(GOPATH) go list ./... | grep -v "^$(PACKAGE)/vendor/"))
 GOFILES = $(shell find . -name *.go | grep -vE "(\/vendor\/)|(_test.go)")
-TESTPKGS = $(shell env GOPATH=$(GOPATH) $(GO) list -f '{{ if or .TestGoFiles .XTestGoFiles }}{{ .ImportPath }}{{ end }}' $(PKGS))
-
-export GOPATH
-export GOBIN
-
-# Docker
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -11,18 +11,22 @@ GOPATH=$(CURDIR)/.gopath
...
 BASE=$(GOPATH)/src/$(REPO_PATH)
-PKGS     = $(or $(PKG),$(shell cd $(BASE) && env GOPATH=$(GOPATH) $(GO) list ./... | grep -v "^$(PACKAGE)/vendor/"))
+PKGS = $(or $(PKG),$(shell cd $(BASE) && env GOPATH=$(GOPATH) go list ./... | grep -v "^$(PACKAGE)/vendor/"))
 GOFILES = $(shell find . -name *.go | grep -vE "(\/vendor\/)|(_test.go)")
-TESTPKGS = $(shell env GOPATH=$(GOPATH) $(GO) list -f '{{ if or .TestGoFiles .XTestGoFiles }}{{ .ImportPath }}{{ end }}' $(PKGS))
-
-export GOPATH
-export GOBIN
-
-# Docker
```
adrianchiris, line 21: even after the refactoring i would prefer if we could group the variables according to their related usage, e.g have a comment stating that the following variables are docker related so they are not bunched in a single long list of variables.

Makefile
```diff
@@ -37,11 +41,9 @@ ifdef STATIC
 	LDFLAGS=-a -ldflags '-extldflags \"-static\"'
 endif
 
-# Go tools
-GO      = go
```
adrianchiris, line 41: While this is best practice in C/C++ where different compilers are used, i think we only have one compiler in go right ?

So im fine with not having this as a parameter (i also see that other parts of this makefile directly reference go compiler directly)

Makefile
```diff
@@ -134,23 +122,18 @@ lint: | $(BASE) $(GOLINT) ; $(info  running golint...) @ ## Run golint
...
 	 done ; exit $$ret
 
-# Dependency management
 .PHONY: deps-update
 deps-update: ; $(info  updating dependencies...)
-	go mod tidy && go mod vendor
+	@go mod tidy && go mod vendor
 
-# Docker image
-# To pass proxy for Docker invoke it as 'make image HTTP_POXY=http://192.168.0.1:8080'
```
adrianchiris, line 147: nit: i think its not half bad to keep this comment.

adrianchiris (approved):
LGTM Thanks!

martinkennelly (approved):
Thanks for the changes

Issue #288, opened 2020-11-23
Resource names containing hyphens will fail


martinkennelly:
### What happened?
Including a hyphen in a resource name causes SRIOV DP failure at startup.

### What did you expect to happen?
Hyphens are ok in resource names

### What are the minimal steps needed to reproduce the bug?
Include hyphen in resource name
### Anything else we need to know?
NA
### Component Versions
All versions

I checked K8 object API naming convention and hyphens are ok. I checked  Is there a reason why we do not allow it in SRIOV DP?

Regex of resource name:
https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/4563418370e11a92c0c1349b65d32e79a3449fc6/pkg/utils/utils.go#L257

I tried looking for K8 object naming convention: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names

Is this behavior ok with the community?

zshi-redhat:
I forgot why we disable that, but there was a reason. @ahalim-intel do you remember?

adrianchiris:
As i see, the resource name is injected as environment variable into pod in the format `<PCIDEVICE>_<ResourceNameUpperCase>`

e.g:
`PCIDEVICE_MY_PCI_RESOURCE=0000:06:02.7`

environment variables cannot have hyphens so they would need to be translated to another char (not underscore...)

martinkennelly:
@adrianchiris Thank you. 
I will add short PR to add to docs.

PR #299, opened 2020-11-27
Expand resourceName description to include hyphen constraint

martinkennelly:
Due to not being able to set environment vars with hyphens,
we are limiting their use in resourceNames.

#288 
Signed-off-by: Martin Kennelly <martin.kennelly@intel.com>


zshi-redhat:
/lgtm

Another option is to replace hypen in `getEnvs` function and allow use of hypen in resourcePrefix/resourceName.

zshi-redhat (approved):

adrianchiris (approved):

PR #308, opened 2020-12-21
Increment docker image version to v3.3

martinkennelly:
Before I can create a new release which would increment the tag ref, I would like to update the daemonset images versions. I understand this would be invalid until a new release would be created but I would rather this than creating a release which referenced old docker image version. 

Opinions?


zshi-redhat:
> Before I can create a new release which would increment the tag ref, I would like to update the daemonset images versions. I understand this would be invalid until a new release would be created but I would rather this than creating a release which referenced old docker image version.
> 
> Opinions?

+1 on updating the tag info **before** creating the release.
This would allow use of new release by simply downloading the tarball or clone the new tag without having to update the daemonset.yaml or cherry-pick another tag changing commit.

zshi-redhat (approved):

PR #313, opened 2021-01-18
Add pci.ids file in test filesystem

zshi-redhat:
This temporarily fixes a unit test error in travis
where pci.ids doesn't exist in the fake filesystem
and ghw lib tries to download it from internet, but
fails to unzip the downloaded package header.
    
Signed-off-by: Zenghui Shi <zshi@redhat.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ matrix:
 
 before_install:
   - sudo apt-get update -qq
+  - sudo apt-get install hwdata -yq
   - sudo systemctl restart docker
 
 script:
--- a/pkg/utils/testing.go
+++ b/pkg/utils/testing.go
@@ -37,6 +37,23 @@ func (fs *FakeFilesystem) Use() func() {
 			panic(fmt.Errorf("error creating fake file: %s", err.Error()))
 		}
 	}
+	err = os.MkdirAll(path.Join(fs.RootDir, "usr/share/hwdata"), 0755)
+	if err != nil {
+		panic(fmt.Errorf("error creating fake directory: %s", err.Error()))
+	}
+
+	// TODO: Remove writing pci.ids file once ghw is mocked
+	// This is to fix the CI failure where ghw lib fails to
+	// unzip pci.ids file downloaded from internet.
+	pciData, err := ioutil.ReadFile("/usr/share/hwdata/pci.ids")
+	if err != nil {
+		panic(fmt.Errorf("error reading file: %s", err.Error()))
+	}
+	err = ioutil.WriteFile(path.Join(fs.RootDir, "usr/share/hwdata/pci.ids"), pciData, 0644)
+	if err != nil {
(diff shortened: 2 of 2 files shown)


adrianchiris:
ghw may depend on other files later on,
a more robust approach would be to create an interface (and mock) for ghw/pcidb libs and use dependency injection for them.
then in tests it would not depend on the filesystem.

This can be a future improvement ofc.

.travis.yml
```diff
@@ -16,6 +16,7 @@ matrix:
 
 before_install:
   - sudo apt-get update -qq
+  - sudo apt-get install hwdata -y
```
adrianchiris, line 19: nit: similar to the above apt-get update command, id use -q flag as then apt-get output is more suitable for logging

i.e replace `-y` with `-yq`

```
-q, --quiet
Quiet. Produces output suitable for logging, omitting progress indicators. More q's will produce more quiet up to a maximum of two. You can also use -q=# to set the quiet level, overriding the configuration file. Note that quiet level 2 implies -y, you should never use -qq without a no-action modifier such as -d, --print-uris or -s as APT may decided to do something you did not expect.
```

adrianchiris (changes requested):
Thanks for working on it Zenghui !

More context in commit message on why this change is needed for posterity is appreciated

zshi-redhat:
> Thanks for working on it Zenghui !
> 
> More context in commit message on why this change is needed for posterity is appreciated

added.

adrianchiris:
golang:alpine was updated yesterday

https://hub.docker.com/layers/golang/library/golang/alpine/images/sha256-e1b06b72500a80383af2a426db0e29d7b65211b297ffd47a0939266cded3fd6a?context=explore

This causes ddp tool to fail linking.

zshi-redhat:
> golang:alpine was updated yesterday
> 
> https://hub.docker.com/layers/golang/library/golang/alpine/images/sha256-e1b06b72500a80383af2a426db0e29d7b65211b297ffd47a0939266cded3fd6a?context=explore
> 
> This causes ddp tool to fail linking during make image in travis.
> 
> option 1: look into possible fix to ddp tool to work with the installed toolchain in the newer image
> option 2: explicitly specify tag for an older base container image (then look into option 1)

option 2.

Is it an option to update `images/ddptool-1.0.0.0.tar.gz`?

martinkennelly:
I am looking into resolving this DDPTool build error with Alpine 3.13. I think sticking with a previous alpine build version is desired until then to unblock us.

I am also seeing an issue with Alpine 3.13 internally here in Intel behind proxies and we see the same error that started off this PR during building of DP. The alpine community are fixing it.

martinkennelly (approved):

zshi-redhat:
Rebased and include fix for another CI failure: https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/317

PR #316, opened 2021-01-19
fix hadolint errors

### LINKED ISSUE #315: Dockerfile needs fixing
### What happened?
Scanning Dockerfile with hadolint tool returns quite a few errors.
Hadolint scan should result in clean output.
Install hadolint
run 'hadolint images/Dockerfile'

mrymsza:
Fixed errors:
DL3020 Use COPY instead of ADD for files and folders
DL3003 Use WORKDIR to switch to a directory
DL3019 Use the `--no-cache` switch to avoid the need to use `--update` and remove `/var/cache/apk/*` when done installing packages
Remaining errors:
DL3018 Pin versions in apk add. Instead of `apk add <package>` use `apk add <package>=<version>`
(It's probably safer to not pin versions in case they aren't available in the future, which might cause the build to fail.)
DL3006 Always tag the version of an image explicitly (community discussion is required to establish which version of alpine should be used). 

Fixes #315

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -1,24 +1,29 @@
 FROM golang:alpine3.12 as builder
 
-ADD . /usr/src/sriov-network-device-plugin
+COPY . /usr/src/sriov-network-device-plugin
 ADD images/ddptool-1.0.0.0.tar.gz /tmp/ddptool/
 
 ENV HTTP_PROXY $http_proxy
 ENV HTTPS_PROXY $https_proxy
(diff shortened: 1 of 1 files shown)


martinkennelly (approved):
LGTM

adrianchiris (approved):
Thanks !

martinkennelly:
@mrymsza Could you rebase this please? I want the CI to pass before merging.

PR #318, opened 2021-01-25
Fix alpine 3.13 APK proxy build issue

### LINKED ISSUE #315: Dockerfile needs fixing

martinkennelly:
Alpine 3.13 set APK repositories to https instead of http.
APK does not work with https proxy because of libfetch issue.
Set alpine to 3.12 until issue is resolved.

See: 
https://github.com/alpinelinux/docker-alpine/issues/136
https://github.com/alpinelinux/docker-alpine/issues/98
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=220468

Partially fixes #315

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -11,7 +11,7 @@ RUN apk add --update --virtual build-dependencies build-base linux-headers && \
     make build && \
     cd /tmp/ddptool && make
 
-FROM alpine
+FROM alpine:3.12
 RUN apk add hwdata-pci
 COPY --from=builder /usr/src/sriov-network-device-plugin/build/sriovdp /usr/bin/
 COPY --from=builder /tmp/ddptool/ddptool /usr/bin/


adrianchiris (approved):
This also aligns build image with final image.

once the issue is resolved we can update both hopefully

martinkennelly:
> This also aligns build image with final image.
> 
> once the issue is resolved we can update both hopefully

Yes, it was a mistake not setting the images to a specific version. We should only upgrade once images are tested.
I think a patch version increment is needed once this is merged - so v3.3.1. WDYT?

adrianchiris:
essentially building from tag v3.3 is broken right ? so yes i think its a good idea,

martinkennelly:
> essentially building from tag v3.3 is broken right ? yes i think its a good idea,

yes - within proxy environments. Alpines APK repos were changed from http to https. This is a good thing but there is a bug in libfetch with respecting proxy settings it seems which APK depends on. So upgrading libfetch is a must and they are working on it.

Issue #327, opened 2021-03-05
Build, test and static analysis for each PR/push and time period using GitHub actions


martinkennelly:
Hey folks,
I think the experiment for NRI seems to work fine. I propose adding build, test and static analysis GitHub actions for each PR / push and also build and unit test at a specific time interval.
I will add the PRs for consideration which will give us a sample output from the GAs.

adrianchiris:
I think we are now missing this bit and we can mark this as done.

> also build and unit test at a specific time interval.

PR #328, opened 2021-03-05
Add Build/Test/Lint/go mod/vendor consistency workflow

martinkennelly:

#327


adrianchiris:
With golangci, i think its better to maintain a `.golangci.yaml` file in the project, then golang-ci uses it when invoked.
(that way you can also invoke it on your machine before submitting PR)
example for .golangci.yaml we use in one of our projects: https://github.com/Mellanox/network-operator/blob/master/.golangci.yml

Also i think it might be better to have this as another job within the test workflow that way we consolidate tests to a single file WDYT?

It would be great if there was also a makefile target for it so its easier for developers to test it before submitting a PR
maybe by replacing the existing make lint target as it covers golint as well.

example of Makefile target:
install golangci: https://github.com/Mellanox/network-operator/blob/eb50b92f07d54f8fd2bc878ea8b1662cbce198a4/Makefile#L114
run lint: https://github.com/Mellanox/network-operator/blob/eb50b92f07d54f8fd2bc878ea8b1662cbce198a4/Makefile#L135

Invoking it locally via Makefile can be done at a later stage though.

martinkennelly:
@adrianchiris Good point. Ill rework it to what you suggested.

martinkennelly:
Will add the makefile updates in a following PR. Thanks for the great input @adrianchiris. Combining build, test, lint it into one workflow is more ergonomic and gives you an overview on a single 'webpage'.

adrianchiris:
@martinkennelly will you be able to add another commit to fix the issues discovered in the tests you added ?

martinkennelly:
@adrianchiris Force push was re-basing against master.
I've reduced 95% of the issues in the static scan outputs. Further reduction will need separate PRs and discussion. If you notice anymore low hanging fruit, let me know.

In the commit "Partial fix for static analysis errors", I needed to add "context" to the go mod file to fix an issue. I then executed go mod tidy in that same commit.
 
In the next commit, I then added the changes from go mod vendor to a separate commit which may aid your review of all the static scan fixes. Enjoy the review :)...

pkg/netdevice/netDeviceProvider.go
```diff
@@ -104,16 +106,16 @@ func hasDefaultRoute(pciAddr string) (bool, error) {
 	}
 
 	if len(ifNames) > 0 { // there's at least one interface name found
-		for _, ifName := range ifNames {
-			link, err := netlink.LinkByName(ifName)
+		for iIfName := range ifNames {
```
adrianchiris, line 109: not really a fan of this, do you recall which CI check complains for the string copy here ?
i dont think these memory optimizations should be a strict requirement.

pkg/types/types.go
```diff
@@ -71,13 +71,14 @@ var SupportedDevices = map[DeviceType]int{
...
 
-// ResourceConfig contains configuration paremeters for a resource pool
+// ResourceConfig contains configuration parameters for a resource pool
 type ResourceConfig struct {
-	ResourcePrefix string           `json:"resourcePrefix,omitempty"` // optional resource prefix that will ovewrite global prefix specified in cli params
-	ResourceName   string           `json:"resourceName"`             // the resource name will be added with resource prefix in K8s api
-	DeviceType     DeviceType       `json:"deviceType,omitempty"`
-	Selectors      *json.RawMessage `json:"selectors,omitempty"`
-	SelectorObj    interface{}
+	ResourcePrefix string `json:"resourcePrefix,omitempty"` // optional resource prefix that will overwrite
```
adrianchiris, line 76: nit: you can put long comments above the variable, I think its clearer that way, especially for multiline comments

.github/workflows/build-test-lint.yml
```diff
@@ -0,0 +1,94 @@
...
+      uses: ludeeus/action-shellcheck@master
+      with:
+        ignore: vendor
+
+  hadolint:
+    runs-on: ubuntu-latest
+    name: Hadolint
+    steps:
+    - uses: actions/checkout@v2
+    - uses: brpaz/hadolint-action@v1.2.1
```
adrianchiris, line 70: should we just ignore `DL3018` for now ?

adrianchiris:
looking at ShellCheck issues, they seem relatively straight forward.

most remaining golang-ci issues dont seem to be too complex, but are not instant fixes.
(some may need a localy ignore in code IMO). Im fine with following up on remaining issue.

pkg/netdevice/netDeviceProvider.go
martinkennelly: Linter message:
```hugeParam: X is heavy (C bytes); consider passing it by pointer (gocritic)``` 

yeah - I did it just to silence the CI check output. Definitely agree it shouldn't be a strict requirement since this software is just a control plane component and it's dealing with objects in memory that are relatively small mem footprint (~300bytes) and are few in number.

Do you not like it because it reduces readability? I think it's fine but happy to revert if you want.

File: .github/workflows/build-test-lint.yml (hunk 3 above)
Comment (martinkennelly) [reply .github/workflows/build-test-lint.yml:70]: I don't think we shouldn't ignore it because it will reduce the need for a discussion and outcome. I will create an issue following this PR to get consensus for this issue. We may just want to ignore it..

**Pros for setting it to a version:**
We get a version we know that works - more deterministic.
**Cons for settings it to a version:**
We need to update the version before its no longer available in the package managers repo.

Comment (adrianchiris) [reply .github/workflows/build-test-lint.yml:72]: yes +1 on opening an issue and meanwhile ignore so check passes with current state of things.

File: pkg/netdevice/netDeviceProvider.go (hunk 1 above)
Comment (adrianchiris) [reply pkg/netdevice/netDeviceProvider.go:109]: > Do you not like it because it reduces readability?

yes exactly. so essentially i prefer to ignore this check or make it less restrictive so it would work in most cases.

LMK if you have other thoughts on this, its just my opinion here :) and ill be fine to keep it as is if you think its important to have this check enforced.


perhaps just set a bigger threshold for this.

from gocritic docs:
> `@hugeParam.sizeThreshold` size in bytes that makes the warning trigger (default 80)

so maybe set this to a higher number ? (512?)

Comment (martinkennelly) [reply pkg/netdevice/netDeviceProvider.go]: Done

File: pkg/types/types.go (hunk 2 above)
Comment (martinkennelly) [reply pkg/types/types.go]: done

File: .github/workflows/build-test-lint.yml (hunk 3 above)
Comment (martinkennelly) [reply .github/workflows/build-test-lint.yml]: Done. Ignored following community response on the issue.

**Comment by martinkennelly** [comment]
@adrianchiris I needed to bump hadolint version to get ignore option and then it showed up issues in rhel7 dockerfile. Corrected them minor issue and needed to add an ignore tag for yum to hadolint. I think we are good for merge. Thanks for all your support here.

**Comment by adrianchiris** [comment]
Thanks for working on this Martin !
I will sort out remaining lint issues once we merge this.

I see Mellanox CI is failing on this PR with error deploying device plugin.
@martinkennelly  did you encounter any issue running device plugin from your PR ?

```
flag provided but not defined: -v 10 --log_dir /var/log/sriovdp --alsologtostderr
Usage of /usr/bin//sriovdp:
  -alsologtostderr
    	log to standard error as well as files
  -config-file string
    	JSON device pool config file location (default "/etc/pcidp/config.json")
  -log_backtrace_at value
    	when logging hits line file:N, emit a stack trace
  -log_dir string
    	If non-empty, write log files in this directory
  -logtostderr
    	log to standard error instead of files
  -resource-prefix string
    	resource name prefix used for K8s extended resource (default "intel.com")
  -stderrthreshold value
    	logs at or above this threshold go to stderr
  -v value
    	log level for V logs
  -vmodule value
    	comma-separated list of pattern=N settings for file-filtered logging
```

adrianchiris:
Thanks Martin !

LGTM, I will followup with a fix for remaining golangci-lint issues.

martinkennelly:
> @martinkennelly did you encounter any issue running device plugin from your PR ?
@adrianchiris Embarrassingly, I did not. Lesson learned.

PR #331, opened 2021-03-05
Update go mod / sum to fix Go 1.16 build

martinkennelly:
For Go 1.16, No automatic changes to go.mod and go.sum [1].
Previously, if Go found a problem in go.mod or go.sum,
it would attempt to fix however this feature is not
available for 1.16.

[1] https://blog.golang.org/go116-module-changes

Signed-off-by: Kennelly, Martin <martin.kennelly@intel.com>

#330

PR #332, opened 2021-03-06
Add test GitHub action

martinkennelly:

GA will fail until #331 is merged.

Issue: #327

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,25 @@
+name: Test
+on:
+    push:
+    pull_request:
+jobs:
+  build:
+    strategy:
+      matrix:
+        go-version: [1.13.x, 1.14.x, 1.15.x, 1.16.x]
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v1
+        with:
+          go-version: ${{ matrix.go-version }}
+
+      - name: Check out code into the Go module directory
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,25 @@
...
+    pull_request:
+jobs:
+  build:
+    strategy:
+      matrix:
+        go-version: [1.13.x, 1.14.x, 1.15.x, 1.16.x]
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Set up Go 1.13
```
manuelbuil, line 13: The go version moves, right? So you should just add Set up go and remove the "1.13"

martinkennelly: Corrected.

.github/workflows/test.yml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          go-version: ${{ matrix.go-version }}
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: Install hwdata
+        run: sudo apt-get install hwdata -y
+
+      - name: Go test
```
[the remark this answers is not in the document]
martinkennelly: The reason why I did it was to ensure ```make test``` is ok for each version of golang. There is a small chance here to capture bugs for sure however it is free to execute this on github, and we just have to worry about log clutter only. I think that is an ok cost. Its easy to see at a glance if pass or failed.

martinkennelly:
Combined into PR #328

PR #333, opened 2021-03-15
Fix DDPTool build with gcc 10

martinkennelly:
Add upgraded DDPTool source code [1] with fix for compiling
with GCC 10. This issue was due to GCC 10 linker not
tolerating redeclartion of global variable.

[1] https://github.com/intel/ddp-tool


zshi-redhat:
@martinkennelly  out of curious, do you know there is any work related to using devlink to query DDP info?

martinkennelly:
@zshi-redhat For E800 series there is support in devlink to read the DDP state of NICs. So for E800 series, you do not need DDPTool at all. We do need it for X700 series however for loading package and viewing DDP state. 

Name | Type | Example | Description
-- | -- | -- | --
fw.app.name | running | ICE OS Default Package | The name of the DDP package that is active in the device. The DDP package is loaded by the driver during initialization. Each variation of the DDP package has a unique name.
fw.app | running | 1.3.1.0 | The version of the DDP package that is active in the device. Note that both the name (as reported by fw.app.name) and version are required to uniquely identify the package.
fw.app.bundle_id | running | 0xc0000001 | Unique identifier for the DDP package loaded in the device. Also referred to as the DDP Track ID. Can be used to uniquely identify the specific DDP package.

[ref](https://www.kernel.org/doc/html/latest/networking/devlink/ice.html)

I want to switch to devlink for E800 series.

zshi-redhat:
@martinkennelly thanks for the explanation and reference link!

zshi-redhat (approved):

PR #329, opened 2021-03-05
Add build GA

martinkennelly:
Signed-off-by: Kennelly, Martin <martin.kennelly@intel.com>

#327

PR #334, opened 2021-03-17
Add github actions

manuelbuil:
Based on the github actions in the sriov-operator project, add the same kind of actions here.

We did not have a non-rhel Dockerfile in this project, so I added one

Signed-off-by: Manuel Buil <mbuil@suse.com>

### DIFF
--- a/.github/workflows/image-push-release.yml
+++ b/.github/workflows/image-push-release.yml
@@ -1,0 +1,39 @@
+name: "Push images on release"
+on:
+  push:
+    tags:
+      - v*
+jobs:
+  build-and-push-image-device-plugin:
+    runs-on: ubuntu-20.04
+    env:
+      IMAGE_NAME: ghcr.io/${{ github.repository }}
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to Docker
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Docker meta
+        id: docker_meta
+        uses: crazy-max/ghaction-docker-meta@v1
+        with:
+          images: ${{ env.IMAGE_NAME }}
+          tag-latest: false
+
+      - name: Build and push sriov-network-device-plugin
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: true
+          tags: |
+            ${{ steps.docker_meta.outputs.tags }}
+          file: images/Dockerfile
--- a/.github/workflows/image-push-master.yml
+++ b/.github/workflows/image-push-master.yml
@@ -1,0 +1,33 @@
+name: "Push images on merge to master"
+on:
+  push:
+    branches:
+      - master
+jobs:
+  build-and-push-device-plugin:
+    runs-on: ubuntu-20.04
+    env:
+      IMAGE_NAME: ghcr.io/${{ github.repository }}
+    steps:
+      - name: Check out the repo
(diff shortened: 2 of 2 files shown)


manuelbuil:
There was a non-rhel Dockerfile in images, so we will use that one instead

.github/workflows/go.yml
```diff
@@ -0,0 +1,43 @@
...
+
+  build:
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.15
```
martinkennelly, line 15: What does the '^' symbol accomplish?
Latest golang is 1.16.
Also, do you think there's value iterating over a number of golang versions? 
I have caught a build error before when I have built on newer golang versions. I was hoping to keep a list of the last 4 minor golang versions e.g. 1.16.x, 1.15.x, etc to ensure any changes still build at least across different golang versions. The builds are in parallel so it doesn't effect the speed of the workflow.

```
jobs:
  build:
    strategy:
      matrix:
        go-version: [1.13.x, 1.14.x, 1.15.x, 1.16.x]
        os: [ubuntu-latest]
````

File: .github/workflows/go.yml [hunk 2]
```diff
@@ -0,0 +1,43 @@
...
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: fmt
```
martinkennelly, line 20: Is there any need for this task when ```make all``` performs 'fmt' anyway?

.github/workflows/image-push-release.yml
```diff
@@ -0,0 +1,39 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Docker meta
+        id: docker_meta
+        uses: crazy-max/ghaction-docker-meta@v1
+        with:
+          images: ${{ env.IMAGE_NAME }}
+          tag-latest: false
+
+      - name: Build and push sriov-network-operator
```
martinkennelly, line 32: ```sriov-network-operator``` -> ```sriov-network-device-plugin```

.github/workflows/go.yml
manuelbuil: In theory `^` is regex for "Start with", i.e. ^1.15 should match 1.15.x.

Your idea is good! I saw it in https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/332/files, I am just wondering if it works with the "x", can you please confirm it?

.github/workflows/go.yml
manuelbuil: good point, thanks. It makes sense to remove it

.github/workflows/image-push-release.yml
manuelbuil: thanks!

.github/workflows/go.yml
martinkennelly: Sure - See execution of multiple golang versions here: https://github.com/martinkennelly/network-resources-injector/actions/runs/627385214

adrianchiris:
Hmm, i see #328 #329 #332  address build and test, so perhaps we should limit this to just the image push parts ?
@martinkennelly @manuelbuil  WDYT?

About the PR labeler workflow, i think it may need additional inputs from @zshi-redhat @pliurh if they found it useful in network operator.

manuelbuil:
> Hmm, i see #328 #329 #332 address build and test, so perhaps we should limit this to just the image push parts ?
> @martinkennelly @manuelbuil WDYT?
> 
> About the PR labeler workflow, i think it may need additional inputs from @zshi-redhat @pliurh if they found it useful in network operator.

Ok! Let me just do the image push parts

martinkennelly (approved):
Thank you for the changes!

pliurh:
LGTM

PR #339, opened 2021-03-31
Fix broken URL link to host-device CNI plugin page

SilverSoldier:
Link to host-device CNI plugin page in the README is currently https://www.cni.dev/plugins/main/host-device/ which is broken.


adrianchiris (approved):
Thank you for this fix !

martinkennelly (approved):

Issue #321, opened 2021-01-31
Introduce Process for API changes, experimental features and feature gates


adrianchiris:
As Support for new features are added to sriov-network-device-plugin, these features can:

1. add and/or modify functionality of device plugin
2. introduce new user-facing APIs (and possibly deprecate an old user-facing API)
3. support new "south bound" facing APIs e.g kubelet, device-info-spec

For these features it may be desired to:
1. Introduce feature gates to enable-disable support and functionality in device plugin
2. Mark these features (and their APIs) as experimental as they are not fully mature and their user-facing APIs may change in subsequent releases or removed altogether.
3. Mark an existing API as deprecated and instruct users to use the new API (eventually removing the old API)

Ive created this issue to trigger an open discussion on the issues and see if we can come with a clean way to handle it.

zshi-redhat:
> **As Support for new features are added to sriov-network-device-plugin, these features can:**
> 
>     1. add and/or modify functionality of device plugin
> 
>     2. introduce new user-facing APIs (and possibly deprecate an old user-facing API)
> 
>     3. support new "south bound" facing APIs e.g kubelet, device-info-spec (this is similar to 1 IMO as the API is already defined and device plugin just implements a new functionality)
> 
> 
> **For these features it may be desired to:**
> 
>     1. Introduce feature gates to enable-disable support and functionality in device plugin

so this feature gate will be a new cli cmd option, and the corresponding new feature will only be evaluated when this feature gate is true?

> 
>     2. Mark these features (and their APIs) as experimental as they are not fully mature and their user-facing APIs may change in subsequent releases or removed altogether.

Is this to mark it as experimental or deprecated in the doc or in the code (e.g. warning message that user can be notified when configuring the feature)?

Since we are targeting new API process, do we really need featureGate? or just use the API created for that feature for enablement or disablement (e.g. if API is defined in cli or configMap, then it is enabled by user, vice versa)?

adrianchiris:
>so this feature gate will be a new cli cmd option, and the corresponding new feature will only be evaluated when this feature gate is true?

Yes, thats what i had in mind.  similarly to what is done in [K8s feature gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-stages). 
We can have the feature gates as part of configMap as well (eg a new `featureGates` field) if we want and cli flags will override those.

[...]

PR #344, opened 2021-04-12
Introducing FeatureGates and global feature config

ipatrykx:
Regards discussion: https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/issues/321

This PR introduces FeatureGates mechanism similar to one available in Kubernetes, that could be used by users to disable/enable features provided by SR-IOV DP. Additionally added global config structure that could hold configuration of features introduced in the future.

### DIFF
--- a/cmd/sriovdp/main.go
+++ b/cmd/sriovdp/main.go
@@ -20,6 +20,9 @@ import (
 	"os/signal"
 	"syscall"
 
+	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/config"
+	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/features"
+
 	"github.com/golang/glog"
 )
 
@@ -33,6 +36,8 @@ func flagInit(cp *cliParams) {
 		"JSON device pool config file location")
 	flag.StringVar(&cp.resourcePrefix, "resource-prefix", "intel.com",
 		"resource name prefix used for K8s extended resource")
+	flag.StringVar(&cp.featureGates, "feature-gates", "",
+		"enables or disables selected features")
 }
 
 func main() {
@@ -57,6 +62,41 @@ func main() {
 		glog.Fatalf("Exiting.. one or more invalid configuration(s) given")
 		return
 	}
+
+	config.NewConfig()
+	
+	cfg, err := config.GetConfig()
+	if err != nil {
+		glog.Fatalf("error while getting config: %v", err)
+		return
+	}
+
+	// Read global config
+	if err := cfg.ReadConfig(cp.configFile); err != nil {
+		glog.Error(err)
+		return
+	}
+
+	features.NewFeatureGate()
+
+	fg, err := features.GetFeatureGate()
+	if err != nil {
+		glog.Fatalf("error while getting feature gate: %v", err)
+		return
+	}
+
+	// Set FeatureGates with ConfigMap
+	if err := fg.SetFromMap(cfg.FeatureGates); err != nil {
+		glog.Error(err)
+		return
+	}
+
+	// Set FeatureGates with CLI arguments
+	if err := fg.SetFromString(cp.featureGates); err != nil {
+		glog.Error(err)
+		return
+	}
+
 	glog.Infof("Discovering host devices")
 	if err := rm.discoverHostDevices(); err != nil {
 		glog.Errorf("error discovering host devices%v", err)
--- a/cmd/sriovdp/main.go
+++ b/cmd/sriovdp/main.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"flag"
+	"fmt"
 	"os"
 	"os/signal"
 	"syscall"
@@ -64,7 +65,7 @@ func main() {
 	}
 
 	config.NewConfig()
-	
+
 	cfg, err := config.GetConfig()
 	if err != nil {
 		glog.Fatalf("error while getting config: %v", err)
@@ -77,9 +78,7 @@ func main() {
 		return
 	}
 
-	features.NewFeatureGate()
-
-	fg, err := features.GetFeatureGate()
+	fg, err := prepareFeaturegates()
 	if err != nil {
 		glog.Fatalf("error while getting feature gate: %v", err)
 		return
@@ -103,18 +102,11 @@ func main() {
 		return
 	}
 
-	glog.Infof("Initializing resource servers")
-	if err := rm.initServers(); err != nil {
-		glog.Errorf("error initializing resource servers %v", err)
+	if err := startServers(rm); err != nil {
+		glog.Error(err)
 		return
 	}
 
-	glog.Infof("Starting all servers...")
-	if err := rm.startAllServers(); err != nil {
-		glog.Errorf("error starting resource servers %v\n", err)
-		return
-	}
-	glog.Infof("All servers started.")
 	glog.Infof("Listening for term signals")
 	// respond to syscalls for termination
 	sigCh := make(chan os.Signal, 1)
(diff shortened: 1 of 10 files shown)


cmd/sriovdp/main.go
```diff
@@ -57,6 +61,25 @@ func main() {
 		glog.Fatalf("Exiting.. one or more invalid configuration(s) given")
 		return
 	}
+
+	// Read global config
+	if err := config.Config.ReadConfig(cp.configFile); err != nil {
```
zshi-redhat, line 66: Could you add an example of what the configMap looks like when `featureGates` data entry is added?
It looks like in the ReadConfig function, it is re-using the same cli cmd `configFile` which points to the resourceList data entry in configMap, but the umarshalling inside ReadConfig is failing when there is a resourceList data entry.

pkg/config/config.go
```diff
@@ -0,0 +1,53 @@
...
+	if Config != nil {
+		return Config
+	}
+	newCfg := &config{}
+	newCfg.FeatureGates = make(map[string]bool)
+	return newCfg
+}
+
+// ReadConfig loads config
+func (cfg *config) ReadConfig(configFile string) error {
```
zshi-redhat, line 30: There is a `readConfig` in manager.go file to parse the `resourceList` data in configMap.
I think perhaps we want to combine both `resourceList` and `featureGates` to the same `readConfig` function and make it a sub-element of `resourceManager`.

cmd/sriovdp/main.go
```diff
@@ -33,6 +35,8 @@ func flagInit(cp *cliParams) {
 		"JSON device pool config file location")
 	flag.StringVar(&cp.resourcePrefix, "resource-prefix", "intel.com",
 		"resource name prefix used for K8s extended resource")
+	flag.StringVar(&cp.featureGates, "feature-gates", "",
+		"resource name prefix used for K8s extended resource")
```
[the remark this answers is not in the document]
ipatrykx: Done

cmd/sriovdp/main.go
ipatrykx: Added the example as a part of manager_test, just for now: https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/26fce9ae66492aea18f6b355b0845d2556553fa1/cmd/sriovdp/manager_test.go#L118

pkg/config/config.go
ipatrykx: What I intended, was to make this config as independent of the existing 'manager config' as possible as it is still yet to be decided if if would like to use one big global config, or for example separate the device-discovery-related config and feature-related config (which I think is optional by it's own nature).

pkg/features/features.go
```diff
@@ -0,0 +1,145 @@
...
+	_, exists := fg.knownFeatures[f]
+	return exists
+}
+
+func (fg *featureGate) set(f feature, status bool) error {
+	if !fg.isFeatureSupported(f) {
+		return fmt.Errorf("Feature %s is not supported", f)
+	}
+	fg.enabled[f] = status
+	if status == true && fg.knownFeatures[f].Maturity == Deprecated {
```
martinkennelly, line 102: redundant comparison to true. Also, check if key in knownFeatures before indexing map.

pkg/features/features.go
```diff
@@ -0,0 +1,145 @@
...
+		glog.Warningf("WARNING: Feature %s will be deprecated soon", f)
+	}
+	return nil
+}
+
+// SetFromMap sets the enablement status of featuers accordig to a map
+func (fg *featureGate) SetFromMap(valuesToSet map[string]bool) error {
+	for k, v := range valuesToSet {
+		err := fg.set(feature(k), v)
```
martinkennelly, line 111: nit: you should wrap your if statement around the set:
if err := fg.set.... err != nil {

}

pkg/features/features.go
```diff
@@ -0,0 +1,145 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// SetFromString converts config string to map and sets the enablement status of the selected features
+// copied form k8s and slightly changed - TBC?
```
martinkennelly, line 120: form - > from

pkg/features/features.go
```diff
@@ -0,0 +1,145 @@
...
+func (fg *featureGate) SetFromString(value string) error {
+	featureMap := make(map[string]bool)
+	for _, s := range strings.Split(value, ",") {
+		if len(s) == 0 {
```
martinkennelly, line 124: @ipatrykx Did you think of using [flag.Var](https://pkg.go.dev/flag)? User would need to specify flag ```feature-gates``` repeatedly but we wouldn't have to do this extraction of features which is complex. Probably need to rename feature-gates to feature-gate

martinkennelly (commented):
@ipatrykx I don't see where each features  default "maturity" is located. My understanding is the user can toggle this via CLI or CM but where is the default? Thanks!

cmd/sriovdp/main.go
```diff
@@ -21,6 +21,8 @@ import (
 	"syscall"
 
 	"github.com/golang/glog"
+	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/config"
```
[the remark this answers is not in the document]
ipatrykx: Done

pkg/config/config.go
```diff
@@ -0,0 +1,53 @@
+package config
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+)
+
+// Config is global config object
+var Config *config
```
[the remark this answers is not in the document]
ipatrykx: Changed that to use getters. Also, after thinking about this I've decided that `NewConfig()` and `NewFeatureGate()` will not return the pointer to the config/fg but the usage of getter will be forced, so it should be clearly visible that this object is instantiated inside the package.

pkg/config/config.go
```diff
@@ -0,0 +1,53 @@
...
+		return fmt.Errorf("error reading file %s, %v", configFile, err)
+	}
+
+	if err = json.Unmarshal(rawBytes, &allCfg); err != nil {
+		return fmt.Errorf("error unmarshalling raw bytes %v", err)
+	}
+
+	fgMap := make(map[string]bool)
+
+	if err = json.Unmarshal(allCfg["featureGates"], &fgMap); err != nil {
```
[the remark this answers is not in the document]
ipatrykx: Done

pkg/features/features.go
```diff
@@ -0,0 +1,145 @@
...
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/golang/glog"
+)
+
+// FeatureGate is feature gate 'manager' to be used
+var FeatureGate *featureGate
```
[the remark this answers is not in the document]
ipatrykx: Same as above.

pkg/features/features.go
```diff
@@ -0,0 +1,145 @@
...
+type featureSpec struct {
+	// Default is the default enablement state for the feature
+	Default bool
+	// Maturity indicates the maturity level of the feature
+	Maturity maturity
+}
+
+var defaultSriovDpFeatureGates = map[feature]featureSpec{
+	AlphaFeature:      {Default: false, Maturity: Alpha},
+	BetaFeature:       {Default: true, Maturity: Beta},
```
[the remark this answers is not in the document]
ipatrykx: Removed beta

pkg/features/features.go
```diff
@@ -0,0 +1,145 @@
...
+	Beta = maturity("BETA")
+	// GA - general availability
+	GA = maturity("GA")
+
+	// Deprecated - feature that will be deprecated in 2 releases
+	Deprecated = maturity("DEPRECATED")
+)
+
+// Feature name
+type feature string
```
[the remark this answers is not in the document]
ipatrykx: I thought back then that it will make code more readable, but after thinking about this, changed that to use just string.

pkg/features/features.go
ipatrykx: Did you mean that is thousl be just `if status &&'?

As for the key checking, it's done in `isFeatureSupported()` which is called earlier in the code.

pkg/features/features.go
ipatrykx: Done

pkg/features/features.go
ipatrykx: Done

pkg/features/features.go
ipatrykx: Yes, I was thinking of that, but using flags we will have to add a flag for every feature introduced, so decided that it will be better to do this the same way the k8s handles that (as far as I remember). Would like to hear more input on this from other though.

ipatrykx:
Hi, @martinkennelly, @zshi-redhat, @adrianchiris is it possible we could move this thing further?

ipatrykx:
Hi all!
This PR has been rebased onto the current master.

SchSeba:
Is this PR still relevant or we can close it?

PR #350, opened 2021-05-04
Use exact match or regex for pfname selector

rozeps:
Use regular expression as PF selector.

### DIFF
--- a/pkg/resources/deviceSelectors.go
+++ b/pkg/resources/deviceSelectors.go
@@ -178,14 +178,13 @@ func getItem(hay []string, needle string, what string) string {
 		// May be an exact match or a regex, try both.
 		if strings.EqualFold(strings.Split(item, "#")[0], needle) {
 			return item
+		}
+		re, err := regexp.Compile(strings.Split(item, "#")[0])
+		if err != nil {
+			glog.Infof("getItem(): error compiling %s, may however not be a regex: %v", what, err)
 		} else {
-			re, err := regexp.Compile(strings.Split(item, "#")[0])
-			if err != nil {
-				glog.Infof("getItem(): error compiling %s, may however not be a regex: %v", what, err)
-			} else {
-				if re.Match([]byte(needle)) {
-					return item
-				}
+			if re.Match([]byte(needle)) {
+				return item
 			}
 		}
 	}
--- a/pkg/resources/deviceSelectors.go
+++ b/pkg/resources/deviceSelectors.go
@@ -2,9 +2,11 @@ package resources
 
 import (
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 
+	"github.com/golang/glog"
 	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
 )
 
@@ -104,7 +106,7 @@ func (s *pfNameSelector) Filter(inDevices []types.PciDevice) []types.PciDevice {
 			// Exclude devices that doesn't have a PF name
 			continue
 		}
-		selector := getItem(s.pfNames, pfName)
+		selector := getItem(s.pfNames, pfName, "pfname")
 		if selector != "" {
 			if isSelected(dev, selector) {
 				filteredList = append(filteredList, dev)
@@ -132,7 +134,7 @@ func (s *rootDeviceSelector) Filter(inDevices []types.PciDevice) []types.PciDevi
 			// Exclude devices that doesn't have a root PCI device
 			continue
 		}
-		selector := getItem(s.rootDevices, rootDevice)
+		selector := getItem(s.rootDevices, rootDevice, "rootDevice")
 		if selector != "" {
 			if isSelected(dev, selector) {
 				filteredList = append(filteredList, dev)
@@ -171,10 +173,20 @@ func contains(hay []string, needle string) bool {
 	return false
 }
 
-func getItem(hay []string, needle string) string {
+func getItem(hay []string, needle string, what string) string {
 	for _, item := range hay {
+		// May be an exact match or a regex, try both.
 		if strings.EqualFold(strings.Split(item, "#")[0], needle) {
 			return item
+		} else {
+			re, err := regexp.Compile(strings.Split(item, "#")[0])
+			if err != nil {
+				glog.Infof("getItem(): error compiling %s, may however not be a regex: %v", what, err)
+			} else {
+				if re.Match([]byte(needle)) {
+					return item
+				}
+			}
 		}
 	}
 	return ""
--- a/README.md
+++ b/README.md
@@ -324,8 +324,8 @@ Or
 
 Where:
 
-    `<PFName>`     - is the PF interface name
-    `<RootDevice>` - is the PF PCI address
+    `<PFName>`     - is the PF interface name (may be a regex compatible with https://github.com/google/re2/wiki/Syntax)
+    `<RootDevice>` - is the PF PCI address (may be a regex compatible with https://github.com/google/re2/wiki/Syntax)
     `<SingleVF>`   - is a single VF index (0-based) that is included into the pool
     `<FirstVF>`    - is the first VF index (0-based) that is included into the range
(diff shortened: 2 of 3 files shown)


pkg/resources/deviceSelectors.go
```diff
@@ -171,11 +173,20 @@ func contains(hay []string, needle string) bool {
...
 }
 
-func getItem(hay []string, needle string) string {
+func getItem(hay []string, needle string, what string) string {
 	for _, item := range hay {
+		// May be an exact match or a regex, try both.
 		if strings.EqualFold(strings.Split(item, "#")[0], needle) {
 			return item
 		}
+		re, err := regexp.Compile(strings.Split(item, "#")[0])
```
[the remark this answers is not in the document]
martinkennelly: I think you should split the string once into a var once you have checked it doesn't include more than one '#' char at the start of this loop. Ignore any string that have more than one '#' and log it.

pkg/resources/deviceSelectors.go
```diff
@@ -171,11 +173,20 @@ func contains(hay []string, needle string) bool {
 	return false
 }
 
-func getItem(hay []string, needle string) string {
+func getItem(hay []string, needle string, what string) string {
```
martinkennelly, line 176: I don't think the 'what' arg is needed.

pkg/resources/deviceSelectors.go
```diff
@@ -171,11 +173,20 @@ func contains(hay []string, needle string) bool {
...
 	for _, item := range hay {
+		// May be an exact match or a regex, try both.
```
martinkennelly, line 178: I think I'd like to have this regex matching an optional feature that would be off by default. I am worried a PF name could end up being a regex expression and cause strange matches to occur.

martinkennelly (commented):
Thank you for the patch. Its an interesting enhancement. Id like to have this feature turned off by default.

pkg/resources/deviceSelectors.go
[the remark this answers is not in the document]
rozeps: Sure I understand your concern. Not sure how to make this "turned off", but one way to address both issue could be to use a special notation like "re{<regexp}#<vfs>" so the "re{...}" is the way to turn on the feature and the brackets ensure that everthing included is the regexp and everything that follows is the VF list.

pkg/resources/deviceSelectors.go
rozeps: Sure can remove.

pkg/resources/deviceSelectors.go
rozeps: With the re{}, intent would be clear that it must compile/match.

martinkennelly:
@rozeps Why not use multiple pfNames [selector](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin#config-parameters)?

adrianchiris:
@rozeps can you explain the problem you are trying to solve with this PR ?

rozeps:
We are deploying in multi-nodes OpenShift clusters, and we support few vendors SR-IOV for use in our applications (matching on vendor PF is not sufficient here).

We are reserving VFs 0-2 for internal use then 3-7 for end user applications.

As interface name not fully predictive (systems end up with ens5s0f0 or enp65s0f0, and so on).

I am tring to come up with configuration that would match any PF from vendor/device XYZ, take only VFs 3-7.

martinkennelly:
> We are deploying in multi-nodes OpenShift clusters, and we support few vendors SR-IOV for use in our applications (matching on vendor PF is not sufficient here).
> 
> We are reserving VFs 0-2 for internal use then 3-7 for end user applications.
> 
> As interface name not fully predictive (systems end up with ens5s0f0 or enp65s0f0, and so on).
> 
> I am tring to come up with configuration that would match any PF from vendor/device XYZ, take only VFs 3-7.
@rozeps
Is the PCI address deterministic across your cluster for the PFs you wish to consume?  You can use that instead of PF name.
Or use udev rules to make your interfaces names deterministic?

rozeps:
Well apparently yes, but I just looked at few systems that are based on same Xeon family (same socket/numa architecture), just wonder how that may differ in more heterogeneous h/w environment (that customers would be allowed to use).

In my mind, would be much simpler to say selector: "match vendor=0x8086,device=0x158b,pf=/.*/#3-7"

With the re enclosed in special characters (illegal ifname ones) don't think that would break much existing implementations.

adrianchiris:
> I am tring to come up with configuration that would match any PF from vendor/device XYZ, take only VFs 3-7.

So if you would have a "`VFIndex`" selector it would also satisfy your use-case ?

rozeps:
Sure you mean no PFNames just the VFIndex as list of id that would be perfect (and cleaner I believe).

adrianchiris:
@rozeps this PR was discussed in today's technical meeting

see: https://docs.google.com/document/d/1bEyEh_K9lGwQWHgUHUH-vqodYd1W-LCWRGlW0-ZVR5A/edit#

We would like to understand more about your use-case and whether or not it is possible to get persistent and predictable net device names.

if it is, then using `pfNames` while specifying all relevant netdevs with the specific range should work.
e.g all nodes have 4 NICs and the netdevs are named the same across nodes.

if it isnt, then a "VFIndex" selector might do the trick.
A regex as you proposed will work but it requires to introduce a special notation for it.

rozeps:
We cannot predict all the combinations now and future, but on my test systems I have 2 nodes with mix of different cards (Mellanox/Intel) and the PCI addresses are different, and the network device names also. Anticipating all possible PF names might be error prone one day someone deploy node with slightly different config and the PF name is not on the list ... may take while to figure out.

The VFIndex I think would have been a good idea if initially tought (in addition to the PFName) but right now that could conflict right one may have non-overlapping pfname with "#0-2-3" part and VFIndex.

The regex enclosed withih "/" is backward compatible with existing syntax - when not present exact match on PF name is used otherwise the regex logic.

adrianchiris:
>The VFIndex I think would have been a good idea if initially tought (in addition to the PFName) but right now that could conflict right one may have non-overlapping pfname with "#0-2-3" part and VFIndex.

AND is performed between different selectors, so result will be consistent
It might not make sense to use the VF range notation and VFIndex selector.
there are other combinations that might not make sense.

@zshi-redhat, @martinkennelly  any [additional] thoughts on this ?

zshi-redhat:
> > The VFIndex I think would have been a good idea if initially tought (in addition to the PFName) but right now that could conflict right one may have non-overlapping pfname with "#0-2-3" part and VFIndex.
> 
> AND is performed between different selectors, so result will be consistent
> It might not make sense to use the VF range notation and VFIndex selector.
> there are other combinations that might not make sense.
> 
> @zshi-redhat, @martinkennelly any [additional] thoughts on this ?

Do you want to introduce a match on all PF interface names or do we need a regex for advanced matchings?
If it is about matching all PF interface names, perhaps we can just leave the name empty which indicates matching all.
If we have advanced matching cases, it looks to me that having a special notation like re{...} would provide more readability to user and less confliction with existing selectors.

Eoghan1232:
@zshi-redhat @adrianchiris @rozeps 
What's the status of this PR?

zshi-redhat:
> @zshi-redhat @adrianchiris @rozeps What's the status of this PR?

I think we didn't get conclusion on which approach to implement or use:
1) VF Index selector
2) Regex match in PfName selector
3) Exhaust all PF names in PfName selector w/ defect that PF names may change in heterogeneous env, no change required

rozeps:
Not sure if it can help to decide, but in our case we'd like to have a selector based on "vendor", "device", "driver", and "vfIndex" the later could have syntax like "pfnames: ["#2", "#6", "#8"] for example i.e. no regex needed only VF number.

I guess one could also want: "pfnames: ["enp66s0f0#1", "#2", "enp65s0f1"] i.e. with both, pfname only or vfIndex only.

SchSeba:
Hi @rozeps any update on this PR or we can close it?

SchSeba:
no update closing this one

PR #357, opened 2021-06-21
Make it easier for the end user to debug issues with their sriov config

johnoloughlin:
With this patch the config is logged before the error is returned.
When the user looks through the logs they will be able to see their config,
along with a message asking them to check if config is in json format.

Signed-off-by: John O'Loughlin <john.oloughlin@intel.com>

### DIFF
--- a/cmd/sriovdp/manager.go
+++ b/cmd/sriovdp/manager.go
@@ -81,7 +81,7 @@ func (rm *resourceManager) readConfig() error {
 
 	glog.Infof("raw ResourceList: %s", rawBytes)
 	if err = json.Unmarshal(rawBytes, resources); err != nil {
-		return fmt.Errorf("error unmarshalling raw bytes %v\n please make sure the config is in json format \n", err)
+		return fmt.Errorf("error unmarshalling raw bytes %v please make sure the config is in json format", err)
 	}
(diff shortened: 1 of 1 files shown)


cmd/sriovdp/manager.go
```diff
@@ -79,11 +79,11 @@ func (rm *resourceManager) readConfig() error {
 
 	}
 
+	glog.Infof("raw ResourceList: %s", rawBytes)
 	if err = json.Unmarshal(rawBytes, resources); err != nil {
-		return fmt.Errorf("error unmarshalling raw bytes %v", err)
+		return fmt.Errorf("error unmarshalling raw bytes %v\n please make sure the config is in json format \n", err)
```
adrianchiris, line 84: nit: mind dropping the `\n`  ? 
i.e

```
fmt.Errorf("error unmarshalling raw bytes %v. please make sure the config is in json format", err)
```

adrianchiris (approved):
Thanks!

cmd/sriovdp/manager.go
johnoloughlin: done

zshi-redhat (approved):

PR #348, opened 2021-04-29
Change to GHCR image path

martinkennelly:
Signed-off-by: Martin Kennelly <martin.kennelly@intel.com>

@zshi-redhat I can make further updates to this by creating a new tagged release, so we can update the rest of the deployment artifacts. Currently, we only have a version package for 'latest'.

PR #365, opened 2021-07-26
Updating DDP Docs

Eoghan1232:
version 1.0.0.0 of DDP tool doesn't support E810 Series NICs
Updating Docs to make note of this.

@zshi-redhat

### DIFF
--- a/docs/ddp/README.md
+++ b/docs/ddp/README.md
@@ -29,6 +29,8 @@ For Intel® Ethernet Controller X710, a DDP package can be loaded into the NIC u
 ## Short step by step configuration for E810 & X710 series NICs
 ### Install DDP packages
 #### Intel® Ethernet Controller E810
+Minimum verson of SRIOV Network Device Plugin needed is 3.3.2 to support DDP Profiles on E800 series NIC's. 
+
 By default, ice driver will automatically load the default DDP package. If you require additional protocals beyond the default set available, download and extract the DDP package into your device firmware folder, typically `/lib/firmware/intel/ice/ddp`.
 Additional packages to suit your use-case can be found at [Intel® download center](https://downloadcenter.intel.com/search?keyword=Dynamic+Device+Personalization)
 


Eoghan1232:
Bit of background on this, Tried to use DDP profiles whilst using an older image, which had ddptool version 1.0.0.0. It did not working when using a E810 Series NIC.

Does the doc need to be updated to make note of this? 

The image update in this PR fixed the issue.
https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/348

docs/ddp/README.md
```diff
@@ -29,6 +29,9 @@ For Intel® Ethernet Controller X710, a DDP package can be loaded into the NIC u
 ## Short step by step configuration for E810 & X710 series NICs
 ### Install DDP packages
 #### Intel® Ethernet Controller E810
+ddptool : Version 1.0.0.0 Doesn't support DDP on E810 series NICs.
```
martinkennelly, line 32: I think, it would be better to specify the min version of this software that will work for E810 instead of naming DDPTool which is an internal component and from a user perspective is transparent.

Eoghan1232: Do you mean the min version of SRIOV DP?
if so, it would be the version before #348  was merged.
As it was the old images that contained an old dpptool version.

martinkennelly: yes, min version of SRIOV DP. The latest tagged release 3.3.2 should be the min version which support E800 series DDP.
>Also, it was the ddptool that was the issue, and why it didn't work. so specifying the version of DP I think is irrelevant.

DDPTool is included in SRIOV DP (see images folder). From a user perspective, they don't care about version of DDPTool  version. We do not want users to mix and match this tool but take the DDPTool version which comes with this software, so from their perspective, they only care about what ver of DP works with E800 series.

Eoghan1232: Okay I understand now, that makes sense.
Since the "fix" was done recently and is not in the 3.3.2 release. Should this doc update be put on hold till next release?

Eoghan1232: Done.

martinkennelly (approved):

adrianchiris:
Mellanox CI issues not related to this PR. moving forward with merge

adrianchiris (approved):

PR #387, opened 2021-10-14
Support for acquiring DDP profile using devlink

ipatrykx:
This PR introduces support for devlink commands to be used to acquire DDP profile name on supported hardware. First, the devlink will be used, and if it is not supported then it will fallback to DDPTool as it was before.

Some of the devlink related code will be possibly moved to [vishvananda/netlink](https://github.com/vishvananda/netlink) library if the [PR there](https://github.com/vishvananda/netlink/pull/702) will be accepted.

Later this can be used to add support for DDP profiles in [k8snetworkplumbingwg/sriov-network-operator](https://github.com/k8snetworkplumbingwg/sriov-network-operator).


pkg/utils/devlink.go
```diff
@@ -0,0 +1,267 @@
...
+
+	return true
+}
+
+// DevlinkGetDDPProfiles returns DDP for selected device
+func (dic DevlinkInfoClient) DevlinkGetDDPProfiles(device string) (string, error) {
+	return dic.DevlinkGetDeviceInfoByNameAndKey(pciBus, device, FwAppNameKey)
+}
+
+// DevlinkGetDeviceInfoByNameAndKeys returns values for selected keys in device info
```
adrianchiris, line 73: should this be part of netlink package.

i.e in : https://github.com/vishvananda/netlink/blob/master/devlink_linux.go

adrianchiris: ah, it is (as of 19 hours ago) :)

https://github.com/vishvananda/netlink/commit/b10eb8fe5cf6ad397bcebe30a53d8607d9191b09

zshi-redhat:
@ipatrykx Would you mind updating this PR? I assume some of the util functions have been merged in netlink go library. Thank you!

ipatrykx:
Hi @zshi-redhat , sure, sorry for the delay on that. I am right now on a tight schedule, but will try to do this until Tuesday.

pkg/utils/ddp.go
```diff
@@ -61,14 +80,17 @@ func GetDDPProfiles(dev string) (string, error) {
 	return getDDPNameFromStdout(stdout.Bytes())
 }
 
+// ErrProfileNameNotFound error when DDPTool is supported, but package name is empty
```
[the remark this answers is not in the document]
ipatrykx: `DDP profile name not found` would suggest that the profile is supported but the value is empty, and the other one (in my opinion) could suggest that profile is not even supported. Can't see the second one anywhere so I am not sure what do you mean. :(

pkg/utils/devlink.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+func newDevlinkInfoClient(infoGetter devlinkInfoGetter) DevlinkInfoClient {
+	var dic DevlinkInfoClient
+	dic.infoGetter = infoGetter
+
+	return dic
+}
+
+// IsDevlinkSupportedByPCIDevice checks if PCI devie supports devlink info command
```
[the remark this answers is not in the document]
ipatrykx: Done :)

pkg/utils/devlink.go
```diff
@@ -0,0 +1,267 @@
...
+func (dic DevlinkInfoClient) IsDevlinkSupportedByPCIDevice(device string) bool {
```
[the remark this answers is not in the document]
ipatrykx: You're right. CHnaged that to `IsDevlinkDDPSupportedByDevice`.

pkg/netdevice/pciNetDevice.go
```diff
@@ -90,13 +94,27 @@ func NewPciNetDevice(dev *ghw.PCIDevice, rFactory types.ResourceFactory, rc *typ
 		linkType = la.EncapType
 	}
 
+	devlinkClient := utils.NewDevlinkInfoClient()
+
+	var ddpFunc ddpProfileGetFunc
+
+	if devlinkClient.IsDevlinkSupportedByPCIDevice(pciAddr) {
```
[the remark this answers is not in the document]
ipatrykx: If this would be a part of `GetDDPProfiles` it would be invoked each time the profile is queried for the device. Now the support will be only checked when the device object is being created. I think this could be discussed.

pkg/netdevice/pciNetDevice.go
```diff
@@ -90,13 +94,27 @@ func NewPciNetDevice(dev *ghw.PCIDevice, rFactory types.ResourceFactory, rc *typ
...
-		linkSpeed: "", // TO-DO: Get this using utils pkg
-		rdmaSpec:  rdmaSpec,
-		linkType:  linkType,
+		PciDevice:     pciDev,
+		ifName:        ifName,
+		pfName:        pfName,
+		linkSpeed:     "", // TO-DO: Get this using utils pkg
+		rdmaSpec:      rdmaSpec,
+		linkType:      linkType,
+		dic:           &devlinkClient,
```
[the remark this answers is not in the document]
ipatrykx: Yes, just deleted the `dic`. :)

pkg/utils/devlink.go
ipatrykx: In fact only general logic is in the netlink package. As we need to retrieve a specific value from the info map those functions have to be here.

pkg/utils/devlink.go
```diff
@@ -0,0 +1,96 @@
...
+	pciBus       = "pci"
+)
+
+// IsDevlinkDDPSupportedByPCIDevice checks if DDP of PCI device can be acquired with devlink info command
+func IsDevlinkDDPSupportedByPCIDevice(device string) bool {
+	return IsDevlinkDDPSupportedByDevice(pciBus, device)
+}
+
+// IsDevlinkDDPSupportedByDevice checks if DDP of device can be acquired with devlink info command
+func IsDevlinkDDPSupportedByDevice(bus, device string) bool {
```
e0ne, line 22: Do we really need to pass hard-coded 'bus' argument to each function? When do you expect it's value won't be 'pci'?

ipatrykx: You're right. And I have, finally, fixed that.

ipatrykx:
> Hi @zshi-redhat , sure, sorry for the delay on that. I am right now on a tight schedule, but will try to do this until Tuesday.

@zshi-redhat I know, I've said I'll deliver it until Tuesday, but Thursday isn't bad as well, isn't it? 🥳

Anyway, I've rebased this PR on top of current master, fixed some issues and added unit tests. Hope we can get it merged soon.

adrianchiris:
@Eoghan1232 @ipatrykx is there a plan to re-kindle this work ?

Eoghan1232:
> @Eoghan1232 @ipatrykx is there a plan to re-kindle this work ?

Hi @adrianchiris , yes of course.
Looks like just to resolve a merge conflict, will get to it this week :)
Then let's get it merged!

Eoghan1232:
will fix lint issues in a moment.

Eoghan1232:
I am currently testing this PR locally and will also provide update soon

Eoghan1232:
@SchSeba PTAL

Eoghan1232:
I've updated the PR and validated it now works.
I've also fixed lint issue.

I can squash the PR at the end if there are no comments/changes needed.

pkg/netdevice/pciNetDevice.go
```diff
@@ -95,24 +100,47 @@ func NewPciNetDevice(dev *ghw.PCIDevice,
...
+	ddpProfile, err := nd.getDDPProfile(pciAddr)
 	if err != nil {
-		glog.Infof("GetDDPProfiles(): unable to get ddp profiles for device %s : %q", pciAddr, err)
-		return ""
+		pfPCI := nd.GetPfPciAddr()
+		ddpProfile, err = nd.getDDPProfile(pfPCI)
+		if err != nil {
+			// default to ddptool if devlink failed
+			ddpProfile, err = utils.GetDDPProfiles(pciAddr)
+			if err != nil {
```
Eoghan1232, line 131: not sure I like this being 3 IF statements deep... 

Thoughts?

pkg/netdevice/pciNetDevice.go
```diff
@@ -95,24 +100,47 @@ func NewPciNetDevice(dev *ghw.PCIDevice,
...
 	return &pciNetDevice{
-		HostDevice:   hostDev,
-		GenPciDevice: *pciDev,
-		GenNetDevice: *netDev,
-		vdpaDev:      vdpaDev,
+		HostDevice:    hostDev,
+		GenPciDevice:  *pciDev,
+		GenNetDevice:  *netDev,
+		vdpaDev:       vdpaDev,
+		getDDPProfile: ddpFunc,
```
adrianchiris, line 118: can we roll these changes into GetDDPProfiles ?  thus there would be no changes when creating PciNetDevice ?

adrianchiris: that way you also dont need to call ddpFuc twice in case of error at L124/127

Eoghan1232: Yes we could move the check if devlink is supported to GetDDPProfiles. 
I will call it devlink query for clarity here:
we would still need:

call devlink query on pci address
call devlink query on the nd parent PF if the first attempt failed
call ddptool if both of the above fail

or am I missing something?

Eoghan1232: sure so,

devlink query only works on PF pci address

so if it first fails on pci address, this could means it's a vf pci
so we get the parent pci address, and try again.
vf will share the pf ddp profile.

as for the ddptool, it works on pf and vf.
thus, only requiring 1 call.

Eoghan1232:
@SchSeba @adrianchiris I've updated this PR. PTAL

pkg/netdevice/pciNetDevice.go
```diff
@@ -33,6 +33,8 @@ type pciNetDevice struct {
 }
 
 // NewPciNetDevice returns an instance of PciNetDevice interface
+//
+//nolint:gocyclo
```
adrianchiris, line 37: do we still need this ? function is not changed now :)

Eoghan1232: you are right - done

adrianchiris:
at some point we would need to switch to interfaces here like we did with sriov-network-operator.
not related to this PR.

pkg/utils/devlink.go
```diff
@@ -0,0 +1,87 @@
...
+// ErrDDPNotSupported error when device does not support DDP
+var ErrDDPNotSupported = errors.New("this device seems not to support DDP")
+
+// KeyNotFoundError returns ErrKeyNotFound
+func KeyNotFoundError(function, key string) error {
+	return fmt.Errorf("%s - %w: %s", function, ErrKeyNotFound, key)
+}
+
+// DDPNotSupportedError returns ErrDDPNotSupported
+func DDPNotSupportedError(device string) error {
```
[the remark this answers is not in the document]
Eoghan1232: not anymore :) will remove !

pkg/utils/ddp.go
```diff
@@ -61,6 +81,9 @@ func GetDDPProfiles(dev string) (string, error) {
...
+var ErrProfileNameNotFound = errors.New("DDP profile name not found")
```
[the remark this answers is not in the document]
Eoghan1232: https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/387/files#diff-83c02f96251b0c36e761108850cbbbeb1eeae8a9d6fb714b473ea8ac1242631aR59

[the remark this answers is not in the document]
Eoghan1232: error message is used here to check if error matches

pkg/utils/ddp.go
```diff
@@ -47,8 +49,26 @@ type DDPpackage struct {
...
 }
 
+// 8 is an exit code of ddptool when profile was not found
+const ddpNoDDPProfile = 8
+
 var ddpExecCommand = exec.Command
 
+// IsDDPToolSupportedByDevice checks if DDPTool can be used with device
+func IsDDPToolSupportedByDevice(dev string) bool {
+	if _, err := GetDDPProfiles(dev); err != nil && err != ErrProfileNameNotFound {
```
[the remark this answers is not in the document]
Eoghan1232: sure can update to match this

Eoghan1232:
@adrianchiris thanks for the feedback! I've updated the PR, PTAL

pkg/netdevice/pciNetDevice.go
```diff
@@ -104,11 +106,30 @@ func NewPciNetDevice(dev *ghw.PCIDevice,
...
-		return ""
+	if utils.IsDevlinkDDPSupportedByPCIDevice(nd.GetPfPciAddr()) {
+		var err error
+		ddpProfile, err = utils.DevlinkGetDDPProfiles(pciAddr)
+		if err != nil {
+			pfPCI := nd.GetPfPciAddr()
+			ddpProfile, err = utils.DevlinkGetDDPProfiles(pfPCI)
+			if err != nil {
+				// default to ddptool if devlink failed
+				ddpProfile, err = utils.GetDDPProfiles(pciAddr)
```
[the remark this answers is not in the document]
Eoghan1232: also good catch - that was leftover from my testing.... I updated it to be :

``utils.IsDDPToolSupportedByDevice(pciAddr)``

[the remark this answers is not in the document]
adrianchiris: say we have a device that supports devlink device info, but does not support ddp.

in this cace dont we want to check if ddp tool is supported ?
i guess end result is the same.

pkg/utils/ddp.go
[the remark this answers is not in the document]
Eoghan1232: I don't understand sorry, we just set a var for the string to check at L59. 
instead of just hardcoding it. 

no one returns the error, since it's a standard error that could occur, we use it for comparison only as a valid error. 
since method only returns true or false.

pkg/netdevice/pciNetDevice.go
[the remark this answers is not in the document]
adrianchiris: ack

pkg/utils/ddp.go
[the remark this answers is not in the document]
adrianchiris: errors.Is() check will not return true in this case

see[1].

we should explicitly return `ErrProfileNameNotFound` in GetDDPProfiles or wrap it in some other error

[1] https://go.dev/play/p/4sv-xe4SPd5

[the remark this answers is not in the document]
Eoghan1232: thank you for the extra clarification
after looking at the go playground code, I see your point. 
it would never return correctly when doing error.Is()

We could just wrap the error then do the comparison/check, so error.Is() will respond with correct response. 
or 
we could update function to return bool and error msg from this func ``IsDDPToolSupportedByDevice``

do you have a preference? I think we could just update the comparison and continue as done before when I had it as ``err != ErrProfileNameNotFound``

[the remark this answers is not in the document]
adrianchiris: what i would do is:

in GetDDPProfile , if err happend , check if it contains the string "DDP profile name not found" then either wrap it with `ErrProfileNameNotFound` or return `ErrProfieNameNotFound`

pkg/utils/ddp.go
```diff
@@ -47,8 +49,26 @@ type DDPpackage struct {
...
 }
 
+// 8 is an exit code of ddptool when profile was not found
+const ddpNoDDPProfile = 8
+
 var ddpExecCommand = exec.Command
 
+// IsDDPToolSupportedByDevice checks if DDPTool can be used with device
+func IsDDPToolSupportedByDevice(dev string) bool {
+	if _, err := GetDDPProfiles(dev); err != nil && !errors.Is(err, ErrProfileNameNotFound) {
```
SchSeba, line 59: sorry I don't follow here why we check both the string and the existCode?

pkg/utils/devlink.go
```diff
@@ -0,0 +1,79 @@
...
+	fwAppNameKey = "fw.app.name"
+	pciBus       = "pci"
+)
+
+var (
+	netlinkDevlinkGetDeviceInfoByNameAsMap = netlink.DevlinkGetDeviceInfoByNameAsMap
+)
+
+// IsDevlinkDDPSupportedByPCIDevice checks if DDP of PCI device can be acquired with devlink info command
+func IsDevlinkDDPSupportedByPCIDevice(device string) bool {
```
SchSeba, line 24: what is the different between the two functions?

pkg/utils/devlink.go
```diff
@@ -0,0 +1,79 @@
+package utils
```
SchSeba, line 1: I think we can make most of the functions private here right?

pkg/utils/ddp.go
Eoghan1232: As the error could imply that the DDP Tool is supported, but returns an empty package name, so it should return true for being supported. 
so we check the returned ExitCode with the ``ddpNoDDPProfile``

pkg/utils/ddp.go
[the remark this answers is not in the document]
Eoghan1232: I've updated so that ``ErrProfileNameNotFound`` is returned

pkg/utils/devlink.go
Eoghan1232: I've made some of the functions private, but could not make all private.

Eoghan1232:
just testing my changes - will push commit after

Eoghan1232:
@SchSeba @adrianchiris  PTAL

pkg/utils/devlink.go
Eoghan1232: I have removed ``IsDevlinkDDPSupportedByPCIDevice`` and used ``IsDevlinkDDPSupportedByDevice`` throughout instead

pkg/utils/devlink.go
```diff
@@ -0,0 +1,74 @@
...
+type NetlinkFunction func(string, string) (map[string]string, error)
+
+const (
+	// fwAppNameKey is a key to extract DDP name
+	fwAppNameKey = "fw.app.name"
+	pciBus       = "pci"
+)
+
+var (
+	netlinkDevlinkGetDeviceInfoByNameAsMap = netlink.DevlinkGetDeviceInfoByNameAsMap
```
adrianchiris, line 20: we already have a netlink provider [1] shouldnt we extend it with the new method ?

then, the rest of the functions in this file would be moved to ddp.go as they are pretty specific for ddp.

in tests we would use a mocked version of NetlinkProvider instead of defining a fake function.


[1] https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/d59e6abe605a4ce0a0e7dca82afff0bfd9fb25fc/pkg/utils/netlink_provider.go#L28

Eoghan1232: Hi @adrianchiris yes 100% agree.

I migrated the code, but have hit an issue with the mockery.
```
2024/05/23 14:26:14 internal error: package "bytes" without types was imported from "github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/utils"
```

I am trying to figure out the reason, but this is blocking me from making the changes, as I cannot regenerate the mocked version of the NetlinkProvider Interface. 

I also did a test of moving the code to a test project and trying to compile but still hit the same issues with mockery. 

Has anyone come across an issue like this with mockery before?

SchSeba: I found this one https://github.com/argoproj/argo-workflows/issues/8251
can you try to build the mockery version please?

Eoghan1232:
trying to fix lint issue.....

adrianchiris (approved):
LGTM

SchSeba (approved):
LGTM

BUT please squash all the commits to reasonable ones :)

adrianchiris:
Squashed to single commit, edited commit msg to contain brief description.

PR #394, opened 2021-10-28
Add the tun mounting with vhost is requested

SchSeba:
This commit add also the tun device into the container.
This is needed to create tap devices inside the container network namespace

Vhost-net is used to accelerate the dpdk connected to kernel networking using tap devices.

### DIFF
--- a/pkg/netdevice/netInfoProviders.go
+++ b/pkg/netdevice/netInfoProviders.go
@@ -76,7 +76,15 @@ func (rip *vhostNetInfoProvider) GetDeviceSpecs() []*pluginapi.DeviceSpec {
 		glog.Errorf("GetDeviceSpecs(): /dev/vhost-net doesn't exist")
 		return nil
 	}
-	return GetVhostNetDeviceSpec()
+	deviceSpec := GetVhostNetDeviceSpec()
+
+	if !TunDeviceExist() {
+		glog.Errorf("GetDeviceSpecs(): /dev/net/tun doesn't exist")
+		return nil
+	}
+	deviceSpec = append(deviceSpec, GetTunDeviceSpec()...)
+
+	return deviceSpec
 }
 
 func (rip *vhostNetInfoProvider) GetEnvVal() string {
--- a/README.md
+++ b/README.md
@@ -232,7 +232,7 @@ This selector is applicable when "deviceType" is "netDevice"(note: this is defau
 | "linkTypes"   | N        | The link type of the net device associated with the PCI device | `string` list Default: `null`                     | "linkTypes": ["ether"]                                                                           |
 | "ddpProfiles" | N        | A map of device selectors                                      | `string` list Default: `null`                     | "ddpProfiles": ["GTPv1-C/U IPv4/IPv6 payload"]                                                   |
 | "isRdma"      | N        | Mount RDMA resources                                           | `bool`  values `true` or `false` Default: `false` | "isRdma": `true`                                                                                 |
-| "needVhostNet"| N        | Share /dev/vhost-net                                           | `bool`  values `true` or `false` Default: `false` | "needVhostNet": `true`                                                                           |
+| "needVhostNet"| N        | Share /dev/vhost-net and /dev/net/tun                          | `bool`  values `true` or `false` Default: `false` | "needVhostNet": `true`                                                                           |
 
 
 [//]: # (The tables above generated using: https://ozh.github.io/ascii-tables/)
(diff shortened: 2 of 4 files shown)


pkg/netdevice/netInfoProviders.go
```diff
@@ -72,11 +72,21 @@ func NewVhostNetInfoProvider() types.DeviceInfoProvider {
...
+	deviceSpec := make([]*pluginapi.DeviceSpec, 2)
+
 	if !VhostNetDeviceExist() {
 		glog.Errorf("GetDeviceSpecs(): /dev/vhost-net doesn't exist")
 		return nil
 	}
-	return GetVhostNetDeviceSpec()
+	deviceSpec[0] = GetVhostNetDeviceSpec()[0]
+
+	if !TunDeviceExist() {
```
adrianchiris, line 83: This means we now require /dev/net/tun to exist when `needVhostNet` is set to `true` in resource config.

is there no use-case for vhost-net without tun device ?

SchSeba: Thanks for the comment!

That was my question I am not aware from our side of any use case that we use vhost until now.
Now we have a use-case (dpdk slow path/ vpp) there is a need to create a tap device and for that, the tun device should be available inside the pod.

The other solution is to have a new flag for the tun device but I will like to know if there is a use-case for only vhost-net mount if not I don't see a reason to create another flag to will be always equal to the vhost one

NicolasDichtel: With our use case, we also need /dev/net/tun, it is exported to via:
      volumes:                                                                                       
      - hostPath:                                                                                    
          path: /dev/net/tun                                                                         
          type: ""                                                                                   
        name: net

adrianchiris: @amorenoz I see you also participated in discussions on the original vhostNet support PR #228 
any thoughts on this ? do you see a case where we need vhost but not `/dev/net/tun` ?

amorenoz: Not with upstream DPDK. The vhost-net (virtio-user) PMD creates a tap device and configures it as the backend of the vhost-net device.

adrianchiris: ack, thanks for the input. if all use-cases sift down to needing tun as well, to me it makes sense to provide it as a mount when needVhostNet is specified.

lets see if we get additional inputs, maybe raise this in one of the NPWG network resource mgmt meeting ?

SchSeba: I will do it at the next meeting thanks everyone for the inputs!

pkg/netdevice/netInfoProviders.go
```diff
@@ -72,11 +72,21 @@ func NewVhostNetInfoProvider() types.DeviceInfoProvider {
...
 
 func (rip *vhostNetInfoProvider) GetDeviceSpecs() []*pluginapi.DeviceSpec {
+	deviceSpec := make([]*pluginapi.DeviceSpec, 2)
+
 	if !VhostNetDeviceExist() {
 		glog.Errorf("GetDeviceSpecs(): /dev/vhost-net doesn't exist")
 		return nil
 	}
-	return GetVhostNetDeviceSpec()
+	deviceSpec[0] = GetVhostNetDeviceSpec()[0]
```
adrianchiris, line 81: why not just 

```
deviceSpec := GetVhostNetDeviceSpec()

...
...

deviceSpec := append(deviceSpec, GetTunDeviceSpec()...)
return deviceSpec
```

adrianchiris (changes requested):
Could you also update documentation as well to mention this new mount ?

pkg/netdevice/netInfoProviders.go
SchSeba: I just didn't want to append as in the underling it will create a new slice and copy everything but again here is only a small slice so if you want I can change the code :)

adrianchiris: also, if later on, if either GetVhostNetDeviceSpec() or GetTunDeviceSpec() end up returning more than one element, you will not need to change here :)

SchSeba: works for me :)

SchSeba:
document updated.

@adrianchiris can you have another round of review please

pkg/netdevice/netInfoProviders.go
```diff
@@ -72,11 +72,20 @@ func NewVhostNetInfoProvider() types.DeviceInfoProvider {
 /* DeviceInfoProvider Interface */
 
 func (rip *vhostNetInfoProvider) GetDeviceSpecs() []*pluginapi.DeviceSpec {
+
```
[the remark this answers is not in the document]
SchSeba: done thanks, the linter was complaining about it :)

zshi-redhat (approved):

zshi-redhat:
merge with two approvals.

Issue #383, opened 2021-10-09
Fail to detect Mellanox VFs in switchdev mode using pfName nicSelector


zshi-redhat:
### What happened?

OpenShift cluster, Mellanox ConnectX-5 configured in switchdev mode using sriov-network-operator, VFs can be observed on the target worker node, but device plugin fail to detect VFs with pfName nicSelector, which results in zero resource reported in node status. 

It is noticed that when VF is released from sriov pod, its representor name appears as the first element in the host directory as below (because of the naming convention):

```
# ls /sys/bus/pci/devices/0000\:d8\:01.5/physfn/net/
58609951635cbfa  ens8f1  ens8f1_0  ens8f1_1  ens8f1_3  ens8f1_4  ens8f1_5
```

0000\:d8\:01.5: the VF PCI address
58609951635cbfa: the VF representor name after its VF released from sriov pod
ens8f1:  PF name of the VF.

This 58609951635cbfa name is considered as the PF name when [utils.GetPfName](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/master/pkg/utils/utils.go#L62) is called, which of course doesn't match with the pfName specified in the device plugin config, this in turn results in VFs be filtered out from the resource pool.

### What did you expect to happen?

VFs in switchdev mode be detected successfully using pfNames selector.

### What are the minimal steps needed to reproduce the bug?

- Deploy a baremetal kubernetes cluster with one of the workers containing Mellanox CX-5 card
- Deploy sriov-network-operator and apply sriov network node policy (use pfName in the nicSelector field) to configure VFs in switchdev mode on CX-5 interface
- Observe that VFs resource been reported to node status
- Create sriov pod requesting VF resource, delete the pod once it is successfully created
- Restart sriov network device plugin manually (delete the device plugin pod on target node)
- Observe that zero number of resource be reported by sriov device plugin

### Component Versions

Component | Version|
------------------------------|--------------------|
|SR-IOV Network Device Plugin |master|
|SR-IOV CNI Plugin            |master|
|OS                           |4.18.0-341.el8.x86_64|

### Config Files
Config file locations may be config dependent.

##### Device pool config file location (Try '/etc/pcidp/config.json')
```
{
  "resourceList": [
    {
      "resourceName": "mlxnics",
      "selectors": {
        "vendors": [
          "15b3"
        ],
        "pfNames": [
          "ens8f1"
        ],
        "rootDevices": [
[...]

PR #395, opened 2021-10-29
Fix detection of Mellanox VFs in switchdev mode

almaslennikov:
This fixes https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/issues/383 by calling the function implemented in  https://github.com/Mellanox/sriovnet/blob/ecc40df73c7c2d53a10fd86368583de1937c40f6/sriovnet_switchdev.go#L60 that correctly determines the name of PF in switchdev mode for given VF by checking phys_port_id for each device in `/sys/bus/devices/<vf_address>/physFn/net`

The original issue is that current logic selects the first device in `physFn/net` which might not be the PF if the device is in switchdev mode. The proposed solution is to use a function from https://github.com/Mellanox/sriovnet library that iterates over the devices and determines the PF by validating its physical port

Signed-off-by: Alexander Maslennikov <amaslennikov@nvidia.com>


go.mod
```diff
@@ -4,6 +4,7 @@ go 1.13
 
 require (
 	github.com/Mellanox/rdmamap v1.0.0
+	github.com/Mellanox/sriovnet v1.0.2 // indirect
```
moshe010, line 7: why indirect comment  ?

pkg/utils/utils.go
```diff
@@ -76,7 +78,13 @@ func GetPfName(pciAddr string) (string, error) {
 		}
 		return "", err
 	} else if len(files) > 0 {
-		return files[0].Name(), nil
+		name, err := sriovnet.GetUplinkRepresentor(pciAddr)
+
+		if err != nil {
+			return files[0].Name(), nil
```
moshe010, line 84: why not return the error? as I remember this method sriovnet.GetUplinkRepresentor support both old kernel and new kernel

pkg/utils/utils.go
```diff
@@ -76,7 +76,13 @@ func GetPfName(pciAddr string) (string, error) {
 		}
 		return "", err
 	} else if len(files) > 0 {
-		return files[0].Name(), nil
+		name, err := GetSriovnetOps().GetUplinkRepresentor(pciAddr)
```
adrianchiris, line 79: This will fail if device is not in switchdev mode.

we should imo.

1. check switchdev mode using devlink
2. if switchdev - use sriovnet
3. else use existing logic

adrianchiris (changes requested):
Hey !

Added some comments. also please add some more info in commit message explaining the problem and proposed solution

Thanks!

pkg/utils/utils.go
zshi-redhat: Agree to use the conditional check on switchdev mode.
It doesn't seem robust to return the right value based on a failure :-)

pkg/utils/sriovnet_ops.go
```diff
@@ -0,0 +1,29 @@
...
+	GetUplinkRepresentor(vfPciAddress string) (string, error)
+}
+
+type defaultSriovnetOps struct {
+}
+
+var sriovnetOps SriovnetOps = &defaultSriovnetOps{}
+
+// SetSriovnetOpsInst method would be used by unit tests in other packages
+func SetSriovnetOpsInst(mockInst SriovnetOps) {
```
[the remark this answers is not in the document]
adrianchiris: my bad, it does accept interface :) maybe just change the argument name to `impl` or something :)
as it might not be a mock instance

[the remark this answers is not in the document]
almaslennikov: Changed the argument name to `inst` :)

pkg/utils/utils.go
almaslennikov: Added a conditional check

pkg/utils/utils.go
```diff
@@ -76,7 +89,21 @@ func GetPfName(pciAddr string) (string, error) {
 		}
 		return "", err
 	} else if len(files) > 0 {
+		pfEswitchMode, err := getPfEswitchMode(pciAddr)
+		if err != nil {
+			return "", err
+		}
+		if pfEswitchMode == "switchdev" {
```
adrianchiris, line 96: nit: can you define "switchdev" as const ?

pkg/utils/utils.go
```diff
@@ -28,6 +28,19 @@ import (
 var (
 	sysBusPci = "/sys/bus/pci/devices"
 	devDir    = "/dev"
+	// getPfEswitchMode returns PF's eswitch mode for the given VF
+	// If device is not VF then it will return its own eswitch mode
+	getPfEswitchMode = func(pciAddr string) (string, error) {
```
adrianchiris, line 33: not a fan of this, any chance we follow the same pattern in netlink as we did in sriovnet ?

adrianchiris: i know its a bit of refactoring, but its going to make our life easier down the road.

pkg/utils/utils_test.go
```diff
@@ -1,6 +1,7 @@
 package utils
 
 import (
+	mocks "github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/utils/mocks"
```
adrianchiris, line 4: nit should be after external dependencies in its own group

pkg/utils/sriovnet_ops.go
```diff
@@ -0,0 +1,29 @@
+package utils
+
```
adrianchiris, line 2: nit: for consistency also maybe name this `sriovnet_provider.go` or rename `netlink_provider.go` to `netlink_ops.go ` if we end up doing refactoring.

pkg/utils/utils.go
```diff
@@ -76,7 +77,21 @@ func GetPfName(pciAddr string) (string, error) {
 		}
```
adrianchiris, line 77: I think we need to do the same logic for L#76 to support the case where pciAddr is not a VF (i.e it is a PF) and the device eswitch mode is switchdev.

adrianchiris (changes requested):

pkg/utils/netlink_provider.go
```diff
@@ -16,49 +16,44 @@ package utils
...
-type fakeLink struct {
-	nl.LinkAttrs
-}
-
-func (fl fakeLink) Attrs() *nl.LinkAttrs {
-	return &fl.LinkAttrs
-}
+// SetMockNetlinkProvider sets a mocked instance of NetlinkProvider to be used by unit test in other packages
+func SetMockNetlinkProvider() {
+	mockProvider := mocks.NetlinkProvider{}
```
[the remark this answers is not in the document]
almaslennikov: The majority of tests need the same default mock of NetLinkProvider, so I propose to keep both `SetNetlinkProviderInst(inst NetlinkProvider)` and this function but maybe rename it to something like `SetDefaultMockNetlinkProvider` to better emphasize what it does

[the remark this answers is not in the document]
adrianchiris: How about having this helper function as part of a test util pacakge ? IMO if possible we should mixing test code in the same file.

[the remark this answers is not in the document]
almaslennikov: Agree, will extract it to the testing package

pkg/utils/utils.go
```diff
@@ -60,6 +61,20 @@ func GetPfAddr(pciAddr string) (string, error) {
 // GetPfName returns SRIOV PF name for the given VF
 // If device is not VF then it will return its own ifname if exist else empty string
 func GetPfName(pciAddr string) (string, error) {
+	pfEswitchMode, err := GetPfEswitchMode(pciAddr)
```
adrianchiris, line 64: @Eoghan1232 @zshi-redhat for intel nics, i assume we can query the eswitch mode via devlink. can you approve that is the case ?

else we would need to log a warning and fallback on error

go.mod
```diff
@@ -4,17 +4,22 @@ go 1.13
...
+	github.com/google/uuid v1.3.0 // indirect
 	github.com/jaypipes/ghw v0.6.0
 	github.com/jaypipes/pcidb v0.5.0
 	github.com/k8snetworkplumbingwg/network-attachment-definition-client v1.1.1-0.20201119153432-9d213757d22d
 	github.com/onsi/ginkgo v1.12.0
 	github.com/onsi/gomega v1.9.0
 	github.com/pkg/errors v0.9.1
-	github.com/stretchr/testify v1.5.1
-	github.com/vishvananda/netlink v1.1.0
+	github.com/spf13/afero v1.6.0 // indirect
```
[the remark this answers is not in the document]
almaslennikov: Tidied up the deps

pkg/utils/utils.go
zshi-redhat: I tried on the i40e device, got the following error:
```
W1110 00:36:16.715844 2172814 pciNetDevice.go:81] unable to get PF name "error getting devlink device attributes for net device 0000:3b:00.0 no such device"
```

I suppose the eswitch mode query is not supported.
@Eoghan1232  Could you help to confirm?

pkg/utils/netlink_provider.go
```diff
@@ -20,45 +20,43 @@ import (
...
 
-// UseFakeLinks causes GetLinkByName to retrieve fake netlink Link object
-// return value intended to be used in unit-tests as deffered method to
-// restore GetLinkByName to be the actual netlink implementation.
-func UseFakeLinks() func() {
-	getLinkByName = getFakeLinkByName
-	return func() {
-		getLinkByName = nl.LinkByName
+// GetDevLinkDeviceEswitchAttrs returns a devlink device's attributes
+func (defaultNetlinkProvider) GetDevLinkDeviceEswitchAttrs(ifName string) (*nl.DevlinkDevEswitchAttr, error) {
```
zshi-redhat, line 51: Could you change the function argument to pfAddr. I think the PF pci address is used here, not ifName.

pkg/utils/utils.go
```diff
@@ -393,3 +408,17 @@ func GetVFID(pciAddr string) (vfID int, err error) {
 	vfID = -1
 	return vfID, nil
 }
+
+// GetPfEswitchMode returns PF's eswitch mode for the given VF
+// If device is not VF then it will return its own eswitch mode
+func GetPfEswitchMode(pciAddr string) (string, error) {
+	pfAddr, err := GetPfAddr(pciAddr)
+	if err != nil {
+		return "", fmt.Errorf("error getting PF eswitch mode for PCI device %s %v", pciAddr, err)
```
zshi-redhat, line 417: The error message is not accurate, maybe change to `error getting PF PCI address for device %s %v`.

pkg/utils/utils.go
adrianchiris: getting an error with `devlink` tool as well ?

```
root ~ $ devlink dev show
pci/0000:03:00.0
pci/0000:03:00.1
root ~ $ devlink dev eswitch show pci/0000:03:00.0
pci/0000:03:00.0: mode switchdev inline-mode none encap enable
```

zshi-redhat: Yes, error with devlink :
```
# devlink dev eswitch show pci/0000:3b:00.0
devlink answers: No such device
```

adrianchiris: ack, thanks for checking this ! I think some intel NIC do support devlink as we got PR #387 
So, i suggest we check for error (ENODEV) and fallback to previous logic.

Eoghan1232: Will test this today and reply back.

Eoghan1232: Hi @adrianchiris @zshi-redhat 

@ipatrykx has tested this in his env.
It seems currently it is only supported by 800 series Nics and not 700 series Nics.

```
root ~ $ devlink dev eswitch show pci/0000:83:00.0
pci/0000:83:00.0: mode switchdev
root ~ $ devlink dev eswitch show pci/0000:81:00.0
devlink answers: No such device
```
the first card is an 800 series 
the second one is a 700 series with i40e

We also updated i40e to latest 2.17.4 driver and downgraded 800 series to 1.5.8 and it's still the same

```
root ~ $ devlink dev eswitch show pci/0000:81:00.0
devlink answers: No such device
root ~ $ devlink dev eswitch show pci/0000:83:00.0
pci/0000:83:00.0: mode legacy
```

adrianchiris: Thanks for checking this out!

we should do fallback if device does not support devlink ("no such device" error). this would be good for other vendor's that dont support devlink.

almaslennikov: Thanks for testing the changes! Added a check for error message and fallback to the default impl if device doesn't support eswitch mode query

pkg/utils/utils_test.go
```diff
@@ -396,35 +403,89 @@ var _ = Describe("In the utils package", func() {
 	)
 
 	DescribeTable("getting PF names",
-		func(fs *FakeFilesystem, device string, expected string, shouldFail bool) {
+		func(fs *FakeFilesystem, eswitchMode string, device string, expected string, shouldFail bool) {
+			fakeNetlinkProvider := mocks.NetlinkProvider{}
+			fakeNetlinkProvider.
+				On("GetDevLinkDeviceEswitchAttrs",
```
adrianchiris, line 409: can you add also a case for when devlink device attributes return eswitch mode `legacy` ?

you can split the tables to "getting PF names legacy", "getting PF names switchdev" if you feel its getting to complex with the number of args here.

also maybe use other device name to trigger error like "nonDevlinkDevice", "devlinkDeviceSwitchdev", "devlinkDeviceLegacy" ?

passing device name as "eswitchUnsupportedError"/"unknownError" is a bit unclear to read imo.

almaslennikov: Done

adrianchiris (approved):
Thanks for addressing the comments ! LGTM

Eoghan1232:
LGTM

Eoghan1232 (approved):

zshi-redhat:
I tested with current version, it seems the pfName is still retrieved from sysfs, rather than devlink. the `pfEswitchMode` returned from devlink query is empty with cx-5 NIC. @almaslennikov  Does it work for you?

pkg/utils/utils.go
```diff
@@ -60,6 +63,25 @@ func GetPfAddr(pciAddr string) (string, error) {
 // GetPfName returns SRIOV PF name for the given VF
 // If device is not VF then it will return its own ifname if exist else empty string
 func GetPfName(pciAddr string) (string, error) {
+	pfEswitchMode, err := GetPfEswitchMode(pciAddr)
+	if err != nil {
+		// If device doesn't support eswitch mode query, fall back to the default implementation
+		if strings.Contains(strings.ToLower(fmt.Sprint(err)), "no such device") {
```
adrianchiris, line 69: There is an additional use-case which slipped my mind.

if `GetPfName` is called with a PF (i.e non-sriov) address and it does not have sriov VFs then devlink will fail with 

```
[root ~]# devlink dev eswitch show pci/0000:03:00.0
devlink answers: Operation not supported
```

in this case, we should also use sysfs

adrianchiris:
> I tested with current version, it seems the pfName is still retrieved from sysfs, rather than devlink. the pfEswitchMode returned from devlink query is empty with cx-5 NIC. @almaslennikov Does it work for you?

@zshi-redhat when you run `devlink dev eswitch show <devlink-device>`  what are you getting ?
do you have sr-iov enabled ?

pkg/utils/utils.go
almaslennikov: Added a condition handling this case

pkg/utils/utils.go
```diff
@@ -60,6 +63,29 @@ func GetPfAddr(pciAddr string) (string, error) {
 // GetPfName returns SRIOV PF name for the given VF
 // If device is not VF then it will return its own ifname if exist else empty string
 func GetPfName(pciAddr string) (string, error) {
+	pfEswitchMode, err := GetPfEswitchMode(pciAddr)
+	if err != nil {
+		errString := strings.ToLower(fmt.Sprint(err))
+		// If device doesn't support eswitch mode query or doesn't have sriov enabled,
+		// fall back to the default implementation
+		if strings.Contains(errString, "no such device") {
```
adrianchiris, line 71: can we do:

```
if strings.Contains(errString, "no such device") || strings.Contains(errString, "operation not supported" {
  glog.Infof("devlink query for eswitch mode is not supported for device %s. %v", pciAddr, err)
} else {
  return "", err
}
...
```

as its not really a warning, we might have devices that simply dont support devlink.

zshi-redhat:
> > I tested with current version, it seems the pfName is still retrieved from sysfs, rather than devlink. the pfEswitchMode returned from devlink query is empty with cx-5 NIC. @almaslennikov Does it work for you?
> 
> @zshi-redhat when you run `devlink dev eswitch show <devlink-device>` what are you getting ? do you have sr-iov enabled ?

Manually run the devlink cmd, got the following result:
```
# devlink dev eswitch show pci/0000:3b:00.1
devlink answers: Operation not supported
```

With sriov device plugin, it returns empty value:

added the following debug message:
```
@@ -76,6 +79,7 @@ func GetPfName(pciAddr string) (string, error) {
                        return "", err
                }
        }
+       glog.Warningf("Eswitchdev mode for device %s is %s", pciAddr, pfEswitchMode)
        if pfEswitchMode == eswitchModeSwitchdev {
                name, err := GetSriovnetProvider().GetUplinkRepresentor(pciAddr)
```

got the empty output:
```
W1116 02:51:38.215605 2262387 utils.go:82] Eswitchdev mode for device 0000:3b:00.1 is
```

almaslennikov:
@zshi-redhat this could be an issue in the netlink library. The proposed solution is to log a warning when we get an empty eswitch mode and then use the default implementation

adrianchiris (approved):

zshi-redhat (approved):

zshi-redhat:
> @zshi-redhat this could be an issue in the netlink library. The proposed solution is to log a warning when we get an empty eswitch mode and then use the default implementation

I think you're right, the netlink ilbrary returns empty for the PF (w/o sriov configured), in which case it doesn't error out in the GetPfEswitchMode and continue to hit the empty string check and then the previous pfname logic.

zshi-redhat:
Merged per three approvals.

zshi-redhat:
@almaslennikov Thanks for your persistence in fixing the comments!

zshi-redhat:
hmm.. @almaslennikov I just tried the latest master, it seems not getting the eswitch mode correctly.

`0000:af:0a.0` is the VF pci address:
```
W1117 12:23:01.315293 1155807 utils.go:79] Failed to retrieve eswitch mode with devlink query for device 0000:af:0a.0
```

`0000:af:00.1` is the PF pci address
```
# devlink dev eswitch show pci/0000:af:00.1
pci/0000:af:00.1: mode switchdev inline-mode none encap-mode basic
```
Do you see the same issue in your environment?

almaslennikov:
@zshi-redhat I've tested netlink code against devlink cmd on my setup and got the following:
* both netlink and devlink correctly report legacy and switchdev eswitch modes
* for non-sriov case netlink code reports the same error as you get (`Failed to retrieve eswitch mode with devlink query for device`) while devlink reports `operation not supported`

What's your setup?

zshi-redhat:
My setup is simple that one CX-5 with two ports, one port is switchdev enabled and sriov configured.

ens4f1 is the PF with pci address 0000:af:00.1

```
# ethtool -i ens4f1
driver: mlx5e_rep
version: 4.18.0-322.el8.mr942_210708_154
firmware-version: 16.31.1014 (MT_0000000008)
expansion-rom-version: 
bus-info: 0000:af:00.1
supports-statistics: yes
supports-test: no
supports-eeprom-access: no
supports-register-dump: no
supports-priv-flags: no
```
devlink shows the right switchdev mode:

```
devlink dev eswitch show pci/0000:af:00.1
pci/0000:af:00.1: mode switchdev inline-mode none encap-mode basic
```

but I got this message (0000:af:0a.0 is the VF pci address of the above PF):
```
W1117 13:55:41.022025  171331 utils.go:79] Failed to retrieve eswitch mode with devlink query for device 0000:af:0a.0
```

almaslennikov:
@zshi-redhat the version of netlink used in this commit is broken. I've created another PR fixing it. I've described the netlink issue there. Please, take a look: https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/398

Issue #392, opened 2021-10-25
Attached devices are not reported to kubelet upon device plugin restarts


zshi-redhat:
### What happened?

Device plugin (with linkType selector) reports X devices to kubelet, Y number (Y < X) of devices be allocated to SR-IOV Pods.
Device plugin restarts and reports X-Y devices to kubelet.

What happened was linkType attribute (retrieved via netlink in host namespace) is empty when devices are attached to SR-IOV pods, this results in attached devices are filtered out when running linkType selector.

### What did you expect to happen?

Device plugin reports X devices to kubelet after restarting

### Anything else we need to know?

Netlink interface is used to retrieve linkType in the network namespace where device plugin runs:
```
var (
        // getLinkByName is a function that retrieves nl.Link object according to
        // a provided netdev name.
        getLinkByName = nl.LinkByName
)

// GetLinkAttrs returns a net device's link attributes.
func GetLinkAttrs(ifName string) (*nl.LinkAttrs, error) {
        link, err := getLinkByName(ifName)
        if err != nil {
                return nil, fmt.Errorf("error getting link attributes for net device %s %v", ifName, err)
        }
        return link.Attrs(), nil
}

        linkType := ""
        if len(ifName) > 0 { 
                la, err := utils.GetLinkAttrs(ifName)
                if err != nil {
                        return nil, err 
                }
                linkType = la.EncapType
        }
```

### Component Versions
Please fill in the below table with the version numbers of components used.

Component | Version|
------------------------------|--------------------|
|SR-IOV Network Device Plugin |master|

### Config Files
Config file locations may be config dependent.

##### Device pool config file location (Try '/etc/pcidp/config.json')
```
{
  "resourceList": [
    {
      "resourceName": "intelnics",
      "selectors": {
        "vendors": [
          "8086"
        ],
        "pfNames": [
          "ens1f1#0-9"
        ],
        "rootDevices": [
          "0000:3b:00.1"
        ],
        "linkTypes": [
          "ether"
        ],
        "IsRdma": false,
        "NeedVhostNet": false
      },
      "SelectorObj": null
    },
    {
      "resourceName": "mlxnics",
      "selectors": {
        "vendors": [
          "15b3"
        ],
        "pfNames": [
          "ens8f1"
        ],
        "rootDevices": [
          "0000:d8:00.1"
        ],
        "linkTypes": [
          "ether"
        ],
        "IsRdma": false,
        "NeedVhostNet": false
[...]

PR #399, opened 2021-11-23
Use Devlink when possible for NewPciNetDevice

SchSeba:
Try to use devlink api to configure the linkType to be `ether`.
If the card doesn't support this we fall back to netlink.

This commit is a partial fix for https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/issues/392
There are still cards the doesn't support the netlink api like intel xv710


pkg/netdevice/pciNetDevice.go
```diff
@@ -82,7 +82,13 @@ func NewPciNetDevice(dev *ghw.PCIDevice, rFactory types.ResourceFactory, rc *typ
 	}
 
 	linkType := ""
-	if len(ifName) > 0 {
+	if _, err = utils.GetNetlinkProvider().GetDevLinkDevice(pciAddr); err == nil {
+		linkType = "ether"
```
SchSeba, line 86: @adrianchiris question if the nic is InfiniBand will it get reported in `devlink dev` command?

I don't have a card to test this

zshi-redhat: Is there a linkType field in devlink device ?
I only see the port Type field in [devlink port attribute](https://www.kernel.org/doc/html/latest/networking/devlink/devlink-port.html), but not in devlink device.
I was thinking we can reliably get the linkType from devlink interface for devlink device.

zshi-redhat:
As we are adding devlink support in various functional calls, it would good to have a generic `isDevlinkEnabled` function to distinguish devlink call and non-devlink call.

zshi-redhat:
> As we are adding devlink support in various functional calls, it would good to have a generic `isDevlinkEnabled` function to distinguish devlink call and non-devlink call.

This is being added through https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/387

pkg/netdevice/pciNetDevice.go
adrianchiris: > @adrianchiris question if the nic is InfiniBand will it get reported in devlink dev command?

it will report a devlink device. (tested on cx5)

however you need devlink port as Zenghui mentioned.
Now, for infiniband port, i dont think mlx driver has implemented the feature :(.

running on my system i didnt see devlink port for infiniband PF
```
// PF 03:00.1 is infiniband
$ devlink dev show
pci/0000:03:00.0
pci/0000:03:00.1
$ devlink port show
pci/0000:03:00.0/65535: type eth netdev enp3s0f0 flavour physical port 0
```

SchSeba: Thanks for the comment I have to say this is not working for me.

I try it on my system
```
lspci | grep Mellanox
5e:00.0 Ethernet controller: Mellanox Technologies MT27800 Family [ConnectX-5]
5e:00.1 Ethernet controller: Mellanox Technologies MT27800 Family [ConnectX-5]
5e:00.2 Ethernet controller: Mellanox Technologies MT27800 Family [ConnectX-5 Virtual Function]
5e:00.3 Ethernet controller: Mellanox Technologies MT27800 Family [ConnectX-5 Virtual Function]
5e:00.4 Ethernet controller: Mellanox Technologies MT27800 Family [ConnectX-5 Virtual Function]
5e:00.5 Ethernet controller: Mellanox Technologies MT27800 Family [ConnectX-5 Virtual Function]
5e:00.6 Ethernet controller: Mellanox Technologies MT27800 Family [ConnectX-5 Virtual Function]
5e:00.7 Ethernet controller: Mellanox Technologies MT27800 Family [ConnectX-5 Virtual Function]
5e:01.0 Ethernet controller: Mellanox Technologies MT27800 Family [ConnectX-5 Virtual Function]
5e:01.1 Ethernet controller: Mellanox Technologies MT27800 Family [ConnectX-5 Virtual Function]
5e:01.2 Ethernet controller: Mellanox Technologies MT27800 Family [ConnectX-5 Virtual Function]
5e:01.3 Ethernet controller: Mellanox Technologies MT27800 Family [ConnectX-5 Virtual Function]
```

I try the devlink port show and it's empty on my system do I miss something?
```
[root@test core]# devlink port show
[root@test core]# 
```

adrianchiris: > I try the devlink port show and it's empty on my system do I miss something?

hmmm, maybe you need a newer kernel ?

bn222: Chiming in, it works for me on a recent RHEL 8.4 kernel (4.18.0-305.28.1.el8_4.x86_64)

SchSeba: your version is a bit newer

```
[core@cnfdt04 ~]$ uname -a
Linux cnfdt04 4.18.0-305.19.1.el8_4.x86_64 #1 SMP Tue Sep 7 07:07:31 EDT 2021 x86_64 x86_64 x86_64 GNU/Linux
[core@cnfdt04 ~]$ devlink port
[core@cnfdt04 ~]$ 
```

SchSeba: Hi @bn222 I try it with the same kernel you did and I am still not able to get the information

```
[root@cnfdd8 core]# devlink dev
pci/0000:19:00.0
pci/0000:19:00.1
pci/0000:3b:00.0
pci/0000:3b:00.1
pci/0000:86:00.0
pci/0000:86:00.1
[root@cnfdd8 core]# devlink port
[root@cnfdd8 core]# 
[root@cnfdd8 core]# 
[root@cnfdd8 core]# uname -a
Linux cnfdd8 4.18.0-305.28.1.el8_4.x86_64 #1 SMP Mon Nov 8 07:45:47 EST 2021 x86_64 x86_64 x86_64 GNU/Linux
```

bn222: Hi SchSeba, I just checked this again. I was confused by other NICs on my system. I can confirm that I also don't have it showing up in devlink port.
I'm using a slightly different ConnectX-5:

`
d8:00.0 Ethernet controller: Mellanox Technologies MT28800 Family [ConnectX-5 Ex]
d8:00.1 Ethernet controller: Mellanox Technologies MT28800 Family [ConnectX-5 Ex]
`

I also tried the latest RHEL kernel, also not showing anything there.

However, I also just built an upstream kernel, and there it works. I get:

`
pci/0000:d8:00.0/65535: type eth netdev ens8f0np0 flavour physical port 0 splittable false
pci/0000:d8:00.1/131071: type eth netdev ens8f1np1 flavour physical port 1 splittable false
`

I suspect RHEL is missing some patches. @adrianchiris Do you have an idea what patches could be missing?

I could start a bisecting effort to pinpoint the patches if we have no clue.

SchSeba: @bn222 will the type by `ib` if the nic is in InfiniBand mode?

I must say I have a problem needing a specific kernel and the solution working only for some time of network nics.

Is it possible to drop this one now that https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/203 was merged?

Or we are going to have a problem with we want to use InfiniBand?

bn222: It's not that we need a specific kernel. Rather, I think we just miss a fix/patch in the kernel. Once I can pinpoint what changes we need, we would be in a better place to decide.

Regarding k8snetworkplumbingwg/sriov-network-operator#203 which removes the selector, I'm just wondering if there are other issues elsewhere with having an empty link type.

I think that having an additional way to query the link type is still useful. That, or the question becomes what the real validity is of having the link type stored in the first place.

e0ne (approved):

PR #408, opened 2022-02-17
add support for vfOnly in the selector

SchSeba:
This commit add a new variable in the selector.

default will be false and no change in the current device plugin implementation.

When this variable will be true we will only add VFs to the pools even if the PF have 0 VFs

### DIFF
--- a/pkg/accelerator/accelDeviceProvider.go
+++ b/pkg/accelerator/accelDeviceProvider.go
@@ -22,6 +22,7 @@ import (
 	"github.com/jaypipes/ghw"
 
 	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
+	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/utils"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/accelerator/accelDeviceProvider.go
```diff
@@ -124,6 +125,17 @@ func (ap *accelDeviceProvider) GetFilteredDevices(devices []types.PciDevice, rc
...
+		vfOnlyDevices := make([]types.PciDevice, 0)
+		for _, dev := range filteredDevice {
+			if !utils.IsSriovPF(dev.GetPciAddr()) {
+				vfOnlyDevices = append(vfOnlyDevices, dev)
+			}
+		}
+
+		filteredDevice = vfOnlyDevices
+	}
+
```
bn222, line 138: Would it make sense to have some kind of DeviceSelector for this as well?

SchSeba: I try to use it first but the problem is the `DeviceSelector` use a list of string in the interface. that  is why I follow the same case as [https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/408/files/62867e6214b9d1aa45b78ce19144e6785bbf8b88#diff-640d0436ec4d8e1927e54b82a9d791bfb0127eff373c073edc53f69627b4b43aR202]

SchSeba:
As requested in the community meeting closing the PR in favor of #410

PR #409, opened 2022-02-21
Exclude PFs even if they don't have any VFs configured

SchSeba:

### DIFF
--- a/pkg/netdevice/netDeviceProvider.go
+++ b/pkg/netdevice/netDeviceProvider.go
@@ -86,7 +86,7 @@ func (np *netDeviceProvider) AddTargetDevices(devices []*ghw.PCIDevice, deviceCo
 			// exclude netdevice in-use in host
 			if isDefaultRoute, _ := hasDefaultRoute(device.Address); !isDefaultRoute {
 				aPF := utils.IsSriovPF(device.Address)
-				if aPF && utils.SriovConfigured(device.Address) {
+				if aPF {
 					// do not add this device in net device list
 					continue
 				}


zshi-redhat:
reproduced some errors when running golangci-lint locally:
```
$ golangci-lint run --config .golangci.yml
pkg/utils/testing.go:33:50: mnd: Magic number: 0755, in <argument> detected (gomnd)
		err := os.MkdirAll(path.Join(fs.RootDir, dir), 0755)
		                                               ^
pkg/utils/testing.go:39:66: mnd: Magic number: 0600, in <argument> detected (gomnd)
		err := ioutil.WriteFile(path.Join(fs.RootDir, filename), body, 0600)
		                                                               ^
pkg/utils/testing.go:44:63: mnd: Magic number: 0755, in <argument> detected (gomnd)
	err = os.MkdirAll(path.Join(fs.RootDir, "usr/share/hwdata"), 0755)
	                                                             ^
pkg/utils/testing.go:56:85: mnd: Magic number: 0600, in <argument> detected (gomnd)
	err = ioutil.WriteFile(path.Join(fs.RootDir, "usr/share/hwdata/pci.ids"), pciData, 0600)
```

pkg/netdevice/netDeviceProvider.go
```diff
@@ -86,7 +86,7 @@ func (np *netDeviceProvider) AddTargetDevices(devices []*ghw.PCIDevice, deviceCo
 			// exclude netdevice in-use in host
 			if isDefaultRoute, _ := hasDefaultRoute(device.Address); !isDefaultRoute {
 				aPF := utils.IsSriovPF(device.Address)
-				if aPF && utils.SriovConfigured(device.Address) {
+				if aPF {
```
adrianchiris, line 89: We currently use sriov-device-plugin to assign PF to container for Baremetal and VM use-cases
with this change, it will cause device plugin to skip PFs. some Mellanox nics come preconfigured with SR-IOV enabled and 8 VFs although not used.

SchSeba: @adrianchiris will you prefer the second solution?

https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/408

zshi-redhat: > We currently use sriov-device-plugin to assign PF to container for Baremetal and VM use-cases with this change, it will cause device plugin to skip PFs. some Mellanox nics come preconfigured with SR-IOV enabled and 8 VFs although not used.

@adrianchiris I recall that we didn't claim PF support. I think this feature should be documented if we all agree to support PF use case in sriov-network-device-plugin, otherwise it may be ignored by others who contribuing to this project. wdyt?

bn222: I vote for @SchSeba #408 solution since with the default setting it does not have any change

adrianchiris: > @adrianchiris I recall that we didn't claim PF support. I think this feature should be documented if we all agree to support PF use case in sriov-network-device-plugin, otherwise it may be ignored by others who contribuing to this project. wdyt?

I recall we had discussion on it a while ago, where we discussed the use-case which is similar to VM use-case where you use DP to assign a PCI device without it having a parent.

agree, I will some documentation and we can discuss there.

bn222:
ok

SchSeba:
continue the solution in PR https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/410

Issue #374, opened 2021-08-11
Create a release template


adrianchiris:
## What would you like to be added?

A template depicting the release process of the project. it should be used when creating an issue for creating a new release
defined process should be then copied to other resource management projects.

## What is the use case for this feature / enhancement?

formalize the release process so it is not an intrinsic knowledge

Issue #367, opened 2021-07-26
Do we want a stable tag?


martinkennelly:
Now that #348  is merged..

## What would you like to be added? 
Add ```stable-${major version}-${arch}``` where arch is amd64, arm64 and ppc64le.
## How?
We need to change deployments ```imagePullPolicy``` to make this work. I don't know if that's appealing.. 
We can extract major version number from docker meta to create the image tag.
```
        name: Docker meta
        id: docker_meta
        uses: crazy-max/ghaction-docker-meta@v1
        with:
          images: name/app
          tag-semver: |         
            {{major}}
```

## What is the use case for this feature / enhancement?
Allow consumers to consume latest major version updates without updating their yamls. May aid automated testing infra.

## Images we produce today
| Tag                   | Arch  | Description                                                                              |
|-----------------------|-------|------------------------------------------------------------------------------------------|
| latest                | x86   | Latest commit                                                                            |
| latest-amd64          | x86   | Latest commit                                                                            |
| ${GitHub SHA}         | x86   | Git commit e.g "sha256:540af56cbd92b2dfdaea5bac3c44a6f46a277f0dbed000a0993d116fc377f332" |
| ${GitHub tag}         | x86   | GitHub Release tag e.g "v1.0"                                                            |
| latest-arm64          | arm   | Latest commit                                                                            |
| ${GitHub tag}-arm64   | arm   | GitHub Release tag e.g "v1.0-arm64"                                                      |
| latest-ppc64le        | power | Latest commit                                                                            |
| ${GitHub tag}-ppc64le | power | GitHub Release tag e.g "v1.0-ppc64le"                                                    |

adrianchiris:
This was discussed in K8s NPWG network & resource mgmt meeting on 08.10.2021. this enhancement proposal is approved.

documentation should be added as well, explaining the nature of the stable tag.
while it might be relevant for dev deployments and CI/testing, it will not be suitable for production.

while updating documentation it would also be great to document the images we are creating as depicted in this issue.

zshi-redhat:
@martinkennelly @adrianchiris  do we know anyone who is working on this proposal?

martinkennelly:
No. 
@ipatrykx  anyone on your side?

ipatrykx:
@martinkennelly Sorry, I was out of office for over 2 weeks.

After checking with @Eoghan1232 it seems that we don't have anyone working on this.

PR #419, opened 2022-03-29
Bump plugin version to v3.4.0 in deployment examples

almaslennikov:
Proposal for the future:
Using a specific version of the plugin in deployment examples requires updating it for each release. This poses a risk of users
deploying an older version of the plugin than the specific one they checked out.

The proposal is to always use `latest` and then in every release, update yaml for the release and upload it to the release page.

### DIFF
--- a/deployments/k8s-v1.10-v1.15/sriovdp-daemonset.yaml
+++ b/deployments/k8s-v1.10-v1.15/sriovdp-daemonset.yaml
@@ -31,7 +31,7 @@ spec:
       serviceAccountName: sriov-device-plugin
       containers:
       - name: kube-sriovdp
-        image: ghcr.io/k8snetworkplumbingwg/sriov-network-device-plugin:v3.3.2
+        image: ghcr.io/k8snetworkplumbingwg/sriov-network-device-plugin:v3.4.0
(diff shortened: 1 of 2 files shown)


adrianchiris:
Hi @almaslennikov, i think its fine to bump the .yaml to last release like we did before (i.e bump to v3.4.0) 
So we dont change customer experience to much.

we should first agree on this approach in a community meeting, 

there are a couple of issues around release proceess and image tag management::

https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/issues/367
https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/issues/374

almaslennikov:
@adrianchiris Let's bump it to the v3.4.0 now and discuss the transition to `latest` or `stable` tags in the future

adrianchiris (approved):

e0ne:
@SchSeba @zshi-redhat could you please take a look on this PR once you have a time?

zeeke:
LGTM

zshi-redhat (approved):

zshi-redhat:
Mellanox CI failed, shall we rerun? I cannot see the details of the CI job.

PR #429, opened 2022-05-17
Use `go install` to download binaries

### LINKED ISSUE #427: Avoid go getting binaries in Makefile
The way tools binaries are downloaded in [Makefile](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/master/Makefile#L82) may change the project dependencies in go.mod / go.sum files, bringing to CI error like the one [fixed here](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/426).

Furthermore, go getting binaries is [deprecated](https://go.dev/doc/go-get-install-deprecation) in go1.17 and dropped in go1.18.

zeeke:
Downloading binaries with go get may change the
content of go.mod file and it is deprecated in go1.17.
`go install` should be used instead.

Fixes #427

Signed-off-by: Andrea Panattoni <apanatto@redhat.com>


adrianchiris (approved):
Thanks @zeeke  !

SchSeba (approved):

PR #431, opened 2022-06-02
Bump golang 1.18 and k8s 1.24

SchSeba:


adrianchiris (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/k8snetworkplumbingwg/sriov-network-device-plugin
 
-go 1.13
+go 1.17
```
e0ne, line 3: Why thre is 1.17 instead of 1.18?

.github/workflows/build-test-lint.yml
```diff
@@ -5,13 +5,13 @@ jobs:
     name: build
     strategy:
       matrix:
-        go-version: [1.13.x, 1.14.x, 1.15.x, 1.16.x]
+        go-version: [1.17.x, 1.18.x]
```
e0ne, line 8: Do we still need to support 1.17.x?

go.mod
SchSeba: there is no 1.18 version in go.mod files yet

.github/workflows/build-test-lint.yml
SchSeba: I think it's in the build verification (we only publish 1.18 built container images)

Issue #433, opened 2022-06-27
Add ability to select devices based on NIC names


ian-howell:
## What would you like to be added?

I would like to be able to create configmaps which allow for the sriov-device-plugin to select NICs based on their names, e.g.

```
{
    "resourceList": [{
            "resourceName": "sriov_netdevice",
            "selectors": {
                "nicNames": ["net1", "net2"]
            }
        }
}
```

## What is the use case for this feature / enhancement?

This is useful when creating a device in libvirt, as the targets for the current selectors are unknown at VM creation time.

ian-howell:
I'm generating the SRIOV-DP configmap prior to the creation of any VMs. In my case, I'm creating a VM template which is then used by kubevirt to later generate the actual VM. Even though I could specify a pci address for the device in the template, libvirt doesn't guarantee that the resulting device will actually get that address ([libvirt docs](https://libvirt.org/pci-addresses.html)). However, that isn't the case for other device info such as NIC name or MAC address

ian-howell:
@SchSeba Any updates on this?

SchSeba:
@ian-howell Thanks for the info!

SchSeba:
Hi @ian-howell 

can you please have a look on this links and let me know if that solve your use case

https://kubevirt.io/user-guide/virtual_machines/interfaces_and_networks/
https://github.com/kubevirt/kubevirt/blob/main/docs/sriov.md
https://dougbtv.com/nfvpe/2019/05/15/kubevirt-sriov/

SchSeba:
Hi @ian-howell can we close this isse?

ian-howell:
Hey @SchSeba, after some investigation, I've found that the pfNames selector isn't sufficient for our use cases. I need to be able to target the device as it is named within the VM.

SchSeba:
closing this one as we implement the ACPI instead https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/488

PR #434, opened 2022-06-27
Add selectors `nicNames`

ian-howell:
This addresses #433 by adding the `nicNames` fields as selectors

(8/8/2022 Update): This no longer addresses `macAddresses`


adrianchiris:
@ian-howell does PfName selector satisfy your use-case based on community discussion we had on jul 18th ? 

if so we can close this PR and related issue

ian-howell:
Hey @adrianchiris, after some investigation, I've found that the `pfNames` selector isn't sufficient for our use cases. I need to be able to target the device as it is named within the VM. This change seems to satisfy that.

pkg/utils/utils.go
```diff
@@ -24,11 +24,13 @@ import (
 	"strings"
 
 	"github.com/golang/glog"
+	"github.com/jaypipes/ghw"
 )
 
 var (
 	sysBusPci = "/sys/bus/pci/devices"
```
e0ne, line 31: since we're adding sysDir, we can change sysBusPci to something like:
sysBusPci  = sysDir + '/bus/pci/devices'

pkg/resources/pciDevice.go
```diff
@@ -44,6 +45,9 @@ func NewPciDevice(dev *ghw.PCIDevice, rFactory types.ResourceFactory, rc *types.
 		return nil, err
 	}
 
+	// Get NIC name. Ignoring the error because not all PCI Devices will be NICs
+	nicName, _ := utils.GetNicName(pciAddr)
```
e0ne, line 49: There could be different errors returned in by GetNicName function. I'm not sure we should ignore all of them

pkg/utils/utils.go
```diff
@@ -59,6 +61,21 @@ func GetPfAddr(pciAddr string) (string, error) {
...
+	if err != nil {
+		return "", fmt.Errorf("GetNicName(): error getting network info: %v", err)
+	}
+
+	for _, nic := range networkInfo.NICs {
+		if nic.PCIAddress != nil && *nic.PCIAddress == pciAddr {
+			return nic.Name, nil
+		}
+	}
+	return "", fmt.Errorf("error getting NIC name for PCI device %s %v", pciAddr, err)
```
e0ne, line 76: Maybe it's better to return nil, nil in this case

e0ne (changes requested):
Please, add unit-tests for new device selector

ian-howell:
@e0ne Thanks for the review! I think I've resolved all of your comments. Let me know if there's anything else I need to do.

zshi-redhat:
@ian-howell Is this for VF passthrough to VM or is it also applied to virtio interfaces?

ian-howell:
This is just for VF passthrough

ian-howell:
Gentle bump on this one

SchSeba:
I will review this PR sorry for the long wait

ian-howell:
> I will review this PR sorry for the long wait

@SchSeba Any update on this?

README.md
```diff
@@ -243,6 +243,7 @@ This selector is applicable when "deviceType" is "netDevice"(note: this is defau
 |---------------|----------|----------------------------------------------------------------|---------------------------------------------------|--------------------------------------------------------------------------------------|
 | "pfNames"     | N        | VFs from PF matches list of PF names                           | `string` list Default: `null`                     | "pfNames": ["enp2s2f0"] (See follow-up sections for some advance usage of "pfNames")             |
 | "rootDevices"     | N        | VFs from PF matches list of PF PCI addresses               | `string` list Default: `null`                     | "rootDevices": ["0000:86:00.0"] (See follow-up sections for some advance usage of "rootDevices") |
+| "nicNames"    | N        | VFs from PF matches list of interface names                    | `string` list Default: `null`                     | "nicNames": ["net1"]                                                                             |
```
SchSeba, line 246: I think the description is not right here. the nic doesn't need to be a VF from a PF it can be a regular nic. special if you run this device plugin inside a VM there is no PF supporter nic

SchSeba (commented):
I am really sorry for the late review..

this looks fine to me just a small comment for the documentation the code looks good

ian-howell:
Thanks @SchSeba! I've addressed the documentation when you get another chance

ian-howell:
@e0ne Could you give this a review when you get a moment? I think I just need one more approval for merge. Thanks in advance!

ian-howell:
Not trying to be pushy, but could I please get an update on this when someone finds the time?

adrianchiris:
Hi Ian, appologies for the late response.

looking at this PR you essentially add a netdevice selector right ? it will select devices based on their netdev name.

this is problematic since sriov-network-device plugin is often used with CNI which means netdevice will be moved to the container network namespace.

if at that point device plugin is restarted, resources that are already allocated to running pods will not be discovered.

ian-howell:
Hey Adrian, thanks for looking over this.

Yes, the goal is to create a netdevice selector. At the time of instantiation of the device-plugin ConfigMap, all that we have is the interface name. This is because we are creating templates of kubevirt machines, and the concrete VM definitions won't come until afterward. We could specify PCI addresses, but due to [libvirt limitations](https://libvirt.org/pci-addresses.html), we can't guarantee that what we specified will be what appears in the VM.

We did some testing to get a better understanding of the problem you've pointed out, and sure enough, the resources are lost when the device-plugin restarts. We will continue to look for a solution for this, but if you happen to have any ideas that might help us, I would be greatly appreciative.

ian-howell:
@adrianchiris I have a solution which appears to work. Basically, we iterate over the child namespaces, switching as we go, and search within for an interface whose alias matches the interface as it were named in the parent namespace. I've verified that this works on our hardware.

When you get a second, I would appreciate it if you could take a second look at this and let me know what you think. Thanks!

ian-howell:
@adrianchiris Just a gentle bump on this one. Please let me know if there's anything wrong with the latest approach

pkg/utils/utils.go
```diff
@@ -372,6 +377,86 @@ func GetNetNames(pciAddr string) ([]string, error) {
...
+			// search for the interface
+			interfaceDir := fmt.Sprintf("%v/bus/pci/devices/%v/net", tmpDir, pciAddr)
+			fInfos, err := os.ReadDir(interfaceDir)
+			if err != nil {
+				return fmt.Errorf("GetNetNamesFromNetns(): unable to read %v: %w", interfaceDir, err)
+			}
+
+			for _, fInfo := range fInfos {
+				ifaceName := fInfo.Name()
+				alias, _ := GetInterfaceAlias(ifaceName)
```
adrianchiris, line 416: would this contain the original interface name ?

if statement below indicates that this might not be set for netdev ?

i need to give this one a try on a system to be able to provide proper feedback.

with this, we could potentially determine the link type as well when creating the device in `gen_net.go`

we have a community meeting tommorow so i will add this PR to the agenda to get some attention to this PR and have others provide feedback as well.

Eoghan1232:
Hey @ian-howell , couple quick questions:

When you say iterate over namespaces, are these pod namespaces or host namespaces?
SRIOV NDP already operates in the host namespace.

NDP should not be iterating over pod namespaces.

What is the difference between pfName and ifaceName ? only that it is looking for an alias?
What would happen if there is more than one alias with the same name? since they could be named similar but in different namespaces.

ian-howell:
@Eoghan1232 

> NDP should not be iterating over pod namespaces.

I'm curious why not? I don't think what we're trying to achieve is possible without doing so.

Eoghan1232:
> @Eoghan1232
> 
> > NDP should not be iterating over pod namespaces.
> 
> I'm curious why not? I don't think what we're trying to achieve is possible without doing so.

NDP scope should be limited to host network namespace for device discovery on the host.

if NDP starts to snoop over other net namespaces, it is kind of a breach of trust.
The network isolation is there for a reason.

This could pose further issues if NDP starts messing with pod network namespaces

Can you explain the use case a bit more?
from my understanding, you are essentially creating VM's with kubevirt, which is a K8s pod, and then passing the PF to the pod, then want NDP to manage the VF pools from the host side?
i.e. control VF's inside VM from the host.

ian-howell:
The use case is a bit more involved. We're creating a VM with kubevirt, and within that VM, we're spinning up an inner k8s cluster. So the "PF" in this case is technically a VF within the VM. The only information about that function we have to work with is its name as viewed from within the VM

What I want to do is schedule a pod to the inner k8s cluster, and pass the kubevirt pod's VF into the inner pod.

With our testing, we've found that `pfNames` is insufficient for this purpose. The `pciAddresses` selector works, but due to ordering, we can't know the PCI address of the kubevirt pod's VF until after it's running, which is too late.

ian-howell:
> would you be able to join a community meeting on the 4th April - [link](https://docs.google.com/document/d/1sJQMHbxZdeYJPgAWK1aSt6yzZ4K_8es7woVIrwinVwI/edit#heading=h.a169lfuz7mbq) to the meeting doc, there is a zoom link at the top.
> The meeting takes place at 14:00 UTC.

I will be present.

Eoghan1232:
Hi @ian-howell ,

we discussed the new proposal you mentioned above to use the Probe func, to solve the use case.
We need to see what the original intention by Kuberentes was for this probe method and perhaps see what other device plugins have done.
you can also see the discuss notes in the google doc that was linked previously. [link again](https://docs.google.com/document/d/1sJQMHbxZdeYJPgAWK1aSt6yzZ4K_8es7woVIrwinVwI/edit#)

ian-howell:
Hi @Eoghan1232, looks like I might have just missed the meeting. Thanks for keeping this on the agenda, and yes, I will plan on attending the meeting on the 18th.

SchSeba:
Hi @ian-howell I contact the kubevirt team.

I wanted to ask you if you can try to add apci config to the VM definition

```
interfaces:
      - masquerade: {}
        name: default
      - bridge: {}
        name: br10
        acpiIndex: 101
```

Then check `cat /sys/bus/pci/devices/<PCI>/acpi_index

if this exist on your node we can implement a new acpi selector that will always work :)

adrianchiris:
@ian-howell any updates ? 

were the alternatives working for you ?  (acpi index or vm tags) ?

ian-howell:
Hi @SchSeba, @adrianchiris, sorry for the prolonged delay.

I took a look into the acpi index and have found that it fits out use case, while also being dramatically simpler than the previous solutions. I've created a new PR at https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/488. I'll close this PR when we've finalized the new one

adrianchiris:
as PR #488 has merged 🚀 🚀  this one can be closed.

Issue #430, opened 2022-05-25
Scalable Function support


DmytroLinkin:
## What would you like to be added?
Utilizing of Scalable Functions (SFs) as resources exposed to the cluster.

## What is the use case for this feature / enhancement?
SFs are relatively new type of functions which can be created in switchdev mode. Detailed information can be found at this [wiki page](https://github.com/Mellanox/scalablefunctions/wiki). Arm host of the Bluefild DPU is capable to act as node in the cluster. Due to limitations VFs are created on x86 parent host, meaning only SFs can be created on arm host itself.

From plugin point of view SFs are unusual devices. Because SFs doesn't have own dedicated PCI device, it is not possible to implement new DeviceType in addition to NetDeviceType and AcceleratorType. However SFs are created on parent PF PCI device.
[This](https://github.com/DmytroLinkin/sriov-network-device-plugin/pull/1) pull request implements support for SFs as auxiliary devices, by extending configuration with additional field. If this field is provided resource pool of requested auxiliary devices created instead of parent PCI devices.

Perhaps handling SFs (and auxiliary network devices) with sriovdp is not correct since SFs unrelated to SRIOV at all and it is better to create brand new plugin. What do You think?

adrianchiris:
Greetings @DmytroLinkin !

I have raised this issue in today's community meeting. As its a new technology and not a trivial change in device plugin i think
it would be best to have a design doc (usually we use google doc) which will serve as common ground for discussions.

This design doc should include:
- general overview of SFs (with additional links for folks who want abit more of a deepdive)
-  supported kernel / userspace (e.g devlink) versions
- currently supported NICs (that you are aware of ofc)
- use-cases
- proposed external facing API changes (if any)
- proposed coded changes
  - diagrams to ease understanding of code changes if needed
- proposed testing coverage (e.g feature will be covered by unit-tests, integration tests, e2e tests)



Some notes on the implementation approach of PR#435:

I think we should strive to split into a separate device type (e.g AuxDeviceType) with its own set of selectors as:
- different selectors may be used for aux/sf devices which will not be applicable for a pci network device and via versa
-  an aux/sf device is a device on the Kernel Auxiliary bus and not on PCI bus
- an aux/sf device does not have PCI attributes. it does have a parent which is a PCI network device

to achieve this some refactoring may be needed but i think we would end up with a cleaner API and internal model.


Looking forward to seeing the design doc !

[...]

PR #435, opened 2022-06-30
Sf support

DmytroLinkin:
This series adds mechanism to discover SFs or any other auxiliary devices on parent PCI device if they requested in configuration. For example, this config defines SF resources:

```
{
    "resourceList": [{
        "resourceName": "bf2_sf",
        "resourcePrefix": "nvidia.com",
        "selectors": {
            "pfNames": ["p0"],
            "auxDevices": ["sf"]
        }
    }]
}
```

When `auxTypes` field is provided, for each discovered PCI device all related auxiliary devices will be discovered making resource pool from this devices instead of original PCI devices. If `auxTypes` field were omitted (not just empty), then resource pool created as usual. Values for `auxTypes` are not hard coded but driver specific.
Because auxiliary device doesn't have its own dedicated PCI device, but plugin expects that all devices are PCI devices, auxiliary devices extends PCI device data type, where all PCI related fields are taken from parent PCI device. In this way plugin can apply selectors and allocate device on request. However id selectors doesn't work currently. This issue will be fixed by additional merge request.

Fixes: #430

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.17
 
 require (
 	github.com/Mellanox/rdmamap v1.0.0
-	github.com/Mellanox/sriovnet v1.0.3
+	github.com/Mellanox/sriovnet v1.1.0
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
 	github.com/jaypipes/ghw v0.6.0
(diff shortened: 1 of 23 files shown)


moshe010:
Poc PR

PR #436, opened 2022-07-07
Implement and use probe function

bn222:
Instead of having to rely on restarts to detect changes to the devices,
use the probe function (which is called periodically) to scan for
(changes to) devices. If a change is detected, the new list of devices
is advertised to kubelet.

### DIFF
--- a/cmd/sriovdp/manager.go
+++ b/cmd/sriovdp/manager.go
@@ -20,7 +20,6 @@ import (
 	"io/ioutil"
 
 	"github.com/golang/glog"
-	"github.com/jaypipes/ghw"
 
 	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/factory"
 	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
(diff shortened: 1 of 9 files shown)


SchSeba:
/cc @adrianchiris 

@bn222 question can't we have a watch on the config map? so when it gets remounted we make the changes

DmytroLinkin:
> @bn222 question can't we have a watch on the config map? so when it gets remounted we make the changes

Does config map keep tracks the changes of number of VFs? I think that is an idea of probing. Probing also make sense for SFs support - they are created dynamically, one by one, without affecting existing SFs.

bn222:
@DmytroLinkin The config map changes is not tracked dynamically by this PR at this point. Reacting dynamically to config map changes is something we could address in a future PR. This PR focuses on detecting changes as is.

bn222:
@DmytroLinkin Correct

bn222:
note to self: need to add lock around access to `allocated`

SchSeba:
Hi @bn222 any update on this one or we can close it?

bn222:
No, I didn't work on this anymore. I will close it now. I think we need to redo the whole logic. First, we need a cleaning PR. There are unnecessary levels of abstractions in it that need to be cleaned up first. Currently, we have a pod restart to "recheck" the state of devs which is far from optimal. Therefore, I still think we need to do the work that this PR intended to tackle.

PR #438, opened 2022-07-15
Remove IsSriovPF as unused method of PciDevice interface

DmytroLinkin:
This check is done by utility function with the same name, hence
remove method as unused.

Signed-off-by: Dmytro Linkin <dlinkin@nvidia.com>

### DIFF
--- a/pkg/resources/pciDevice.go
+++ b/pkg/resources/pciDevice.go
@@ -108,10 +108,6 @@ func (pd *pciDevice) GetDriver() string {
 	return pd.driver
 }
 
-func (pd *pciDevice) IsSriovPF() bool {
-	return false
-}
-
 func (pd *pciDevice) GetDeviceSpecs() []*pluginapi.DeviceSpec {
 	dSpecs := make([]*pluginapi.DeviceSpec, 0)
 	for _, infoProvider := range pd.infoProviders {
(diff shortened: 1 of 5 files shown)


DmytroLinkin:
/CC @adrianchiris 

There is also unused GetSubClass method, but I supposed it gonna be used with PCI SubClass implementation in the future.

adrianchiris:
Hi @DmytroLinkin ! thanks for working on this cleanup.

> There is also unused GetSubClass method, but I supposed it's gonna be used with PCI SubClass implementation in the future.

it just returns PCI subclass attribute which is not being used throughout the code. im not sure what you are referring in rgards to future use of this method.

IMO it can also be removed. that said, it can be done in a followup PRd.

adrianchiris (approved):

DmytroLinkin:
Mostly I'm talking about this lines
https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/master/pkg/types/types.go#L63-L78
Which along with `GetSubClass` suggests there can be such support in the future.

Eoghan1232 (approved):

adrianchiris:
@DmytroLinkin can you rebase ? we can merge this one IMO

PR #440, opened 2022-07-17
Remove API devices from reosurcePool

adrianchiris:
This commit removes `devices` map from ResourcePoolImpl
and uses devicePool map instead in GetDevices() method.
This is done to avoid duplication by holding API devices
in two places.

In addition, add unit-test for GetDevices().

Signed-off-by: Adrian Chiris <adrianc@nvidia.com>


adrianchiris:
@Eoghan1232 @SchSeba can we merge this one ?

Eoghan1232 (approved):
LGTM

adrianchiris:
Thanks for reviewing !

PR #443, opened 2022-08-30
Support GetPreferredAllocation

wattmto:
This PR adds support of GetPreferredAllocation API which PR #267 tried to support. This PR is based on the latest master branch to make it easier to review and merge (as I wrote in [#267 (comment)](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/267#issuecomment-1217524548)). 

This PR doesn't include Balanced policy I also mentioned in [#267 (comment)](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/267#issuecomment-1217524548) because balanced policy doesn't seem to be the scope of PR #267.

Fixes [#255](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/issues/255)


README.md
```diff
@@ -217,6 +220,7 @@ This plugin creates device plugin endpoints based on the configurations given in
 | "deviceType"     | N        | Device Type for a resource pool.                                                                                  | string value of supported types. Default: "netDevice" | Currently supported values: "accelerator", "netDevice"          |
 | "excludeTopology" | N       | Exclude advertising of device's NUMA topology          | bool Default: "false"    | "excludeTopology": true          |
 | "selectors"      | N        | A map of device selectors. The "deviceType" value determines the "selectors" options.                             | json object as string Default: null                   | Example: "selectors": {"vendors": ["8086"],"devices": ["154c"]} |
+| "allocatePolicy" | N        | Preferred device allocation policy for a resource pool. The "allocatePolicy" value determines which device in a resource pool is allocated first | string value of supported allocate policy. Default: "" | Example: "allocatePolicy": "packed" |
```
e0ne, line 223: Please, add a list of supported values. It should be ["", "packed"] now

pkg/factory/factory.go
```diff
@@ -53,11 +53,22 @@ func (rf *resourceFactory) GetResourceServer(rp types.ResourcePool) (types.Resou
...
 	}
 	return nil, fmt.Errorf("factory: unable to get resource pool object")
 }
 
+// GetAllocator returns an instance of Allocator using preferredAllocationPolicy
+func (rf *resourceFactory) GetAllocator(policy string) types.Allocator {
+	switch policy {
+	case "packed":
+		return resources.NewPackedAllocator()
+	default:
```
e0ne, line 67: Please, add logging for "unsupported" values

pkg/resources/allocator.go
```diff
@@ -0,0 +1,109 @@
+// Copyright 2018 Intel Corp. All Rights Reserved.
```
e0ne, line 1: Should it be 2022?

e0ne (changes requested):

pkg/resources/allocator.go
milran: Yes, changed.
https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/443/files#diff-f19d6492bf6a94ca5a69d2f87d38a7de8430d4c85bd33b76e0dbb80bb0327326R1

README.md
milran: Added supported values.
https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/443/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R223

milran:
@e0ne Thank you for your review! I think I've resolved 2 of your suggestion. Logging for unsupported values is main update, so could you check one? Please let me know if there are anything else I can fix.

e0ne (approved):
LGTM. Thanks for addressing my comments

milran:
Thank you!

README.md
```diff
@@ -217,6 +220,7 @@ This plugin creates device plugin endpoints based on the configurations given in
 | "deviceType"     | N        | Device Type for a resource pool.                                                                                  | string value of supported types. Default: "netDevice" | Currently supported values: "accelerator", "netDevice"          |
 | "excludeTopology" | N       | Exclude advertising of device's NUMA topology          | bool Default: "false"    | "excludeTopology": true          |
 | "selectors"      | N        | A map of device selectors. The "deviceType" value determines the "selectors" options.                             | json object as string Default: null                   | Example: "selectors": {"vendors": ["8086"],"devices": ["154c"]} |
+| "allocatePolicy" | N        | Preferred device allocation policy for a resource pool. The "allocatePolicy" value determines which device in a resource pool is allocated first | string value of supported allocate policy. Default: "" | Currently supported values: "", "packed" |
```
Eoghan1232, line 223: Am I assuming this correctly ? 
Packed : VFs are allocated under the same PF

pkg/factory/factory.go
```diff
@@ -53,11 +53,28 @@ func (rf *resourceFactory) GetResourceServer(rp types.ResourcePool) (types.Resou
...
 
+// GetAllocator returns an instance of Allocator using preferredAllocationPolicy
+func (rf *resourceFactory) GetAllocator(policy string) (types.Allocator, error) {
+	switch policy {
+	case "":
+		return nil, nil
+	case "packed":
+		return resources.NewPackedAllocator(), nil
+	default:
+		return nil, fmt.Errorf("GetSelector(): invalid policy %s", policy)
```
Eoghan1232, line 74:
```suggestion
		return nil, fmt.Errorf("GetAllocator(): invalid policy %s", policy)
```

Should it be like this?

pkg/factory/factory_test.go
```diff
@@ -47,6 +47,28 @@ var _ = Describe("Factory", func() {
 			})
 		})
 	})
+	DescribeTable("getting allocator",
+		func(policy string, shouldSucceed bool, expected reflect.Type) {
+			f := factory.NewResourceFactory("fake", "fake", true)
+			a, e := f.GetAllocator(policy)
```
Eoghan1232, line 53: Could we use different var names here?
a -> allocator ?
e -> error  ?

It would make it easier to read.

pkg/resources/allocator_test.go
```diff
@@ -0,0 +1,194 @@
...
+	var (
+		f  types.ResourceFactory
+		rc *types.ResourceConfig
+	)
+	BeforeEach(func() {
+	})
+	Describe("creating new packed allocator", func() {
+		Context("with valid policy", func() {
+			It("should return valid allocator", func() {
+				pa := resources.NewPackedAllocator()
```
Eoghan1232, line 28: same comment as above.

README.md
milran: Yes, correct.

pkg/factory/factory_test.go
milran: Renamed variables.

pkg/resources/allocator_test.go
milran: Done.

milran:
@Eoghan1232 I think I resolved all of your suggestion. Please let me know anything else I can fix.

Eoghan1232 (approved):
LGTM!

Eoghan1232:
@adrianchiris can you take a look?

milran:
@adrianchiris Please let me know anything I can fix ;-)

adrianchiris:
sure, @milran could you rebase this one ?

milran:
@adrianchiris I rebased on top of current master.

pkg/types/types.go
```diff
@@ -297,6 +301,16 @@ type DeviceInfoProvider interface {
 	GetMounts() []*pluginapi.Mount
 }
 
+// Allocator is an interface to get preferred device allocation
+type Allocator interface {
+	Allocate(*pluginapi.ContainerPreferredAllocationRequest, ResourcePool) []string
+}
+
+// PackedAllocator extends Allocator interface
```
adrianchiris, line 309: do we need this ATM ? we can just use Allocator.

README.md
```diff
@@ -261,14 +264,15 @@ This plugin creates device plugin endpoints based on the configurations given in
...
+| "selectors"       | N        | A map of device selectors. The "deviceType" value determines the "selectors" options.                                                            | json object as string Default: null                    | Example: "selectors": {"vendors": ["8086"],"devices": ["154c"]}                      |
+| "additionalInfo"  | N | A map of map to add additional information to the pod via environment variables to devices                                                       | json object as string Default: null                    | Example: "additionalInfo": {"*": {"token": "3e49019f-412f-4f02-824e-4cd195944205"}}  |
+| "allocatePolicy"  | N        | Preferred device allocation policy for a resource pool. The "allocatePolicy" value determines which device in a resource pool is allocated first | string value of supported allocate policy. Default: "" | Currently supported values: "", "packed"                                             |
```
adrianchiris, line 275: nit: `allocationPolicy` WDYT ?

pkg/resources/allocator.go
```diff
@@ -0,0 +1,109 @@
...
+
+// DeviceSet is to hold and manipulate a set of HostDevice
+type DeviceSet map[string]types.HostDevice
+
+// PackedAllocator implements the Allocator interface
+type PackedAllocator struct {
+}
+
+// Allocator implements the Allocator interface
+type Allocator struct {
```
adrianchiris, line 33: nit: rename Allocator -> DefaultAllocator

pkg/resources/allocator.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// NewPackedAllocator create instance of PackedAllocator
+func NewPackedAllocator() types.Allocator {
```
adrianchiris, line 37: can you return *PackedAllocator here ?
users of this method can then assign this value to types.Allocator typed variable

pkg/resources/allocator.go
```diff
@@ -0,0 +1,109 @@
...
+	return &PackedAllocator{}
+}
+
+// NewAllocator create instance of Allocator
+func NewAllocator() types.Allocator {
```
adrianchiris, line 42: if you accept my rename proposal, this should be NewDefaultAllocator.

also same as above, an you return *DefaultAllocator here ?

pkg/resources/allocator.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// NewDeviceSet is to create an empty DeviceSet
+func NewDeviceSet() DeviceSet {
+	set := make(DeviceSet)
+	return set
+}
+
+// Insert is to add a HostDevice in DeviceSet
+func (ds DeviceSet) Insert(pciAddr string, device types.HostDevice) {
```
adrianchiris, line 53: can you make those pointer receiver ?

pkg/resources/allocator.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"sort"
+
+	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
+
+	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
+)
+
+// DeviceSet is to hold and manipulate a set of HostDevice
+type DeviceSet map[string]types.HostDevice
```
adrianchiris, line 26: id prefer this to be a package internal type (change to lowercase)

pkg/resources/allocator.go
```diff
@@ -0,0 +1,109 @@
...
+		return preferredDevices
+	}
+
+	availableSet := NewDeviceSet()
+	for _, available := range rqt.AvailableDeviceIDs {
+		dev, ok := rp.GetDevicePool()[available]
+		if ok {
+			availableSet.Insert(available, dev)
+		} else {
+			return preferredDevices
```
adrianchiris, line 86: its unexpected to get here right ?
can you emit a warning in log ?

pkg/resources/allocator.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+	}
+	for _, required := range rqt.MustIncludeDeviceIDs {
+		_, ok := rp.GetDevicePool()[required]
+		if ok {
+			availableSet.Delete(required)
+		} else {
+			return preferredDevices
```
adrianchiris, line 94: log warning as above

pkg/resources/allocator.go
```diff
@@ -0,0 +1,109 @@
...
+	ds[pciAddr] = device
+}
+
+// Delete is to delete a HostDevice in DeviceSet
+func (ds DeviceSet) Delete(pciAddr string) {
+	delete(ds, pciAddr)
+}
+
+// Sort is to sort the DeviceSet and return the sorted keys
+func (ds DeviceSet) Sort() []string {
```
adrianchiris, line 63: this reads like you can sort the DeviceSet. IMO its a bit confusing

how about renaming this method to : AsSortedStrings/AsSorted/AsSortedSlice ?

pkg/resources/allocator.go
```diff
@@ -0,0 +1,109 @@
...
+	sort.Strings(keys)
+	return keys
+}
+
+// Allocate return the preferred allocation
+func (pa *PackedAllocator) Allocate(rqt *pluginapi.ContainerPreferredAllocationRequest, rp types.ResourcePool) []string {
+	size := rqt.AllocationSize
+	preferredDevices := make([]string, 0)
+	if size <= 0 || len(rqt.AvailableDeviceIDs) < int(size) || len(rqt.MustIncludeDeviceIDs) > int(size) {
+		return preferredDevices
```
adrianchiris, line 77: emit warning in logs ?

pkg/resources/server.go
```diff
@@ -115,6 +117,19 @@ func (rs *resourceServer) NotifyRegistrationStatus(ctx context.Context,
 	return &registerapi.RegistrationStatusResponse{}, nil
 }
 
+func (rs *resourceServer) GetPreferredAllocation(ctx context.Context,
+	rqt *pluginapi.PreferredAllocationRequest) (*pluginapi.PreferredAllocationResponse, error) {
+	glog.Infof("GetPreferredAllocation called with %+v", rqt)
+	resp := new(pluginapi.PreferredAllocationResponse)
+	for _, container := range rqt.ContainerRequests {
```
adrianchiris, line 124: nit: rename container -> req  or containerReq

pkg/resources/server.go
```diff
@@ -193,7 +202,7 @@ func (rs *resourceServer) PreStartContainer(ctx context.Context,
 func (rs *resourceServer) GetDevicePluginOptions(ctx context.Context, empty *pluginapi.Empty) (*pluginapi.DevicePluginOptions, error) {
 	return &pluginapi.DevicePluginOptions{
 		PreStartRequired:                false,
-		GetPreferredAllocationAvailable: false,
+		GetPreferredAllocationAvailable: (rs.allocator != nil),
```
adrianchiris, line 205: do we really need to set this to true if we end up with the "default" allocator ?

pros to keeping this as is is that we have one flow for allocations.
con is that we add an additional rpc call which is not needed in case of the default allocator

im fine with either, just want to hear other opinions on this one.

adrianchiris:
@milran  appologies for the delayed review. 

I gave this one a proper read, please see comments and thanks for you patience :)

pkg/resources/server.go
milran: I think we don't need to set this to true if we doesn't need a specific allocator. Because, as you pointed out, we don't want  extra meaningless rpc call in case of the default allocator.

README.md
milran: I think `allocationPolicy` is better.

pkg/resources/allocator.go
milran: I think we need this to be public, so that we can use `DeviceSet` in tests.

https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/443/files#diff-8dc00b1b9c75fe62651b7f963c2fb2b56d0d0da00f2aa353c1e511cc1dea3151R79-R84

adrianchiris (commented):
Getting close :)

added a few comments.

also appreciate if you restructure the PR to contain less commits

e.g

1. adding allocator
2.  modifications to resource pool
3. modifications to server
4.  readme updates

pkg/resources/server.go
```diff
@@ -115,6 +117,19 @@ func (rs *resourceServer) NotifyRegistrationStatus(ctx context.Context,
...
 }
 
+func (rs *resourceServer) GetPreferredAllocation(ctx context.Context,
+	rqt *pluginapi.PreferredAllocationRequest) (*pluginapi.PreferredAllocationResponse, error) {
+	glog.Infof("GetPreferredAllocation called with %+v", rqt)
+	resp := new(pluginapi.PreferredAllocationResponse)
+	for _, containerReq := range rqt.ContainerRequests {
+		containerResp := new(pluginapi.ContainerPreferredAllocationResponse)
+		containerResp.DeviceIDs = rs.allocator.Allocate(containerReq, rs.resourcePool)
+		resp.ContainerResponses = append(resp.ContainerResponses, containerResp)
```
e0ne, line 127: We need to use default allocator or check for rs.allocator == nil`

adrianchiris:
@milran hey ! do you still plan to work on this one ?

milran:
@adrianchiris sorry for late response. I am working on this PR and I will finish by next week.

pkg/resources/allocator.go
```diff
@@ -0,0 +1,126 @@
...
+
+// DeviceSet is to hold and manipulate a set of HostDevice
+type DeviceSet map[string]types.HostDevice
+
+// PackedAllocator implements the Allocator interface
+type PackedAllocator struct {
+}
+
+// DefaultAllocator implements the Allocator interface
+type DefaultAllocator struct {
```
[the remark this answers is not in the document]
milran: no, so deleted

pkg/resources/server.go
```diff
@@ -115,6 +117,19 @@ func (rs *resourceServer) NotifyRegistrationStatus(ctx context.Context,
 	return &registerapi.RegistrationStatusResponse{}, nil
 }
 
+func (rs *resourceServer) GetPreferredAllocation(ctx context.Context,
+	rqt *pluginapi.PreferredAllocationRequest) (*pluginapi.PreferredAllocationResponse, error) {
```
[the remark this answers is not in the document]
milran: Done.

pkg/resources/server.go
milran: added nil check.

milran:
@adrianchiris @e0ne Thank you for reviewing! I think all of the points are fixed.

pkg/resources/allocator.go
```diff
@@ -0,0 +1,112 @@
...
+	for _, required := range rqt.MustIncludeDeviceIDs {
+		_, ok := rp.GetDevicePool()[required]
+		if ok {
+			availableSet.Delete(required)
+		} else {
+			glog.Warningf("Allocator(): not available device was included: %s", required)
+			return []string{}
+		}
+	}
+	sortedAvailableSet := availableSet.AsSortedStrings()
```
adrianchiris, line 105: This will work if device to be allocated is a PCI device and device ID is a PCI device ID.

This  is not the case for AuxiliaryDevices of type SF (subfunction).
the ID/Name of such device is as follows: mlx5_core.sf.4  (4 being sort of device/instance index)

so allocator will not work as expected in this case unless the auxiliary device was created  in certain order.

to achieve correctness in such case, best approach is to have a separate allocator for auxiliaryNetDevices where it would perform the check of being allocated from the same parent PCI device. based on auxiliaryNetDevice pfAddr field.

(have buckets of auxiliaryNetDevice by their parent PCI device then allocate from the same bucket)

Alternative is to block creation of a resource pool if the type is AuxNetDevice and non default policy is provided.
Then deal with that as a followup.

adrianchiris: This is something that needs to brought up in our weekly community meetings to get an agreement on which direction its preferrable to go.

technically i think we can use the same allocator for all if implemented in a different way than whats proposed in this PR

README.md
```diff
@@ -269,17 +270,27 @@ This plugin creates device plugin endpoints based on the configurations given in
...
+| "selectors"        | N        | Either a single device selector map or a list of maps. The list syntax is preferred. The "deviceType" value determines the device selector options.                                                  | json list of objects or json object. Default: null                   | Example: "selectors": [{"vendors": ["8086"],"devices": ["154c"]}]        |
+| "additionalInfo"   | N        | A map of map to add additional information to the pod via environment variables to devices                                             | json object as string Default: null  | Example: "additionalInfo": {"*": {"token": "3e49019f-412f-4f02-824e-4cd195944205"}} |
 
 Note: "resourceName" must be unique only in the scope of a given prefix, including the one specified globally in the CLI params, e.g. "example.com/10G", "acme.com/10G" and "acme.com/40G" are perfectly valid names.
 
+#### Allocation Policy
+
+The "allocationPolicy" value determines which device in a resource pool is allocated. Each policy acts as following:
```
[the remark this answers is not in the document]
adrianchiris: That  is , we need to say what happens in case a policy cannot be satisfied for an allocation.

e.g say we have 3 devices to be allocated but only 2 can be allocated from the same "parent"

Eoghan1232:
@milran is this PR still being worked on?

milran:
I've been working on it for a while, but I've gotten busy. Do you mind asking for a review in a few months?

SchSeba:
Hi @wattmto,

I am doing some cleanup do you still working on this PR or I can close it?

wattmto:
Hi, @SchSeba 

I won’t be working on this anymore, I’ll go ahead and close it.

PR #458, opened 2023-01-03
Update sriov device plugin deployment

adrianchiris:
- move yaml to deployment root dir as we dont maintain yamls for multiple k8s versions
- align amd64 daemonset to deploy latest like other arch
- use kubernetes.io/arch label instead of beta.kubernetes.io/arch as it is deprecated

Signed-off-by: adrianc <adrianc@nvidia.com>


zeeke (approved):
/lgtm

PR #463, opened 2023-01-22
Move ddptool build to its own continer

adrianchiris:
due to recent update of gcc in builder base image
compilation of ddptool fails.

split ddptool build to its own container and fix alpine version.

this will allow to control the base image which is used to build device plugin and ddptool separately.

once ddptool is updated, its alpine base image version can be bumped

Signed-off-by: adrianc <adrianc@nvidia.com>

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -1,15 +1,22 @@
 FROM golang:1.18-alpine as builder
 
 COPY . /usr/src/sriov-network-device-plugin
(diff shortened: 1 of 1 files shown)


Eoghan1232 (approved):
lgtm
thanks adrian!

adrianchiris:
super ! thx for reviewing 

merging this one

Issue #470, opened 2023-02-26
Push mainfest on merge to master is failing


adrianchiris:
### What happened?

Push manifest Job is failing:
https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/actions/runs/4274943872/jobs/7442033582

we should fix it.

adrianchiris:
After additional investigation it seems that docker build push action builds and pushes manifest (with single arch) when platform field is specified.

then the manifest job fails since
1. it tries to pull manifest instead of the container image itself, fails on arm64 as the job is running on amd64
2. even if we use the `--platform` flag to pull image, it would still fail on mainfest creation, as it pulls a manifest and not a container image.

adrianchiris:
fixed by #478

PR #471, opened 2023-02-27
Bump github actions to their latests versions

adrianchiris:
workflows complain about deprecated actions.
align actions to their latest version.

it may be relates to #470 , if it doesnt solve it, its a good start :)


Eoghan1232 (approved):
lgtm

adrianchiris:
thx for the review @Eoghan1232 !

PR #475, opened 2023-03-13
Enable array of NetDevice selectors for one resource name

### LINKED ISSUE #444: Allow mixing RDMA and non-RDMA devices in one pool
Currently, RDMA devices must use the `isRdma: true` selector to have their RDMA resources mounted into a pod. This selector (to my understanding) prevents non-RDMA devices from being included in a resource pool with RDMA devices. So, from the user perspective, this selector has the dual responsibility of ensuring RDMA resources are added to containers using RDMA devices and filtering devices which are a part of a resource pool based on their RDMA capabilities.

I propose adding the ability to mix RDMA and non-RDMA devices into the same resource pool. 

An example use case is mixing Intel and Mellanox VFs for DPDK applications. DPDK allows applications to use both of those VFs without any application code changes. Therefore, DPDK applications don't need to care what kind of VF they run on, and should be able to request a DPDK compatible VF. Instead, due to the boolean nature of the `isRdma` selector, the application must request either an Intel VF or Mellanox VF.

## Implementation
I've put together a patch (upstreaming is awaiting internal review) that adds an `auto` value to the `isRdma` selector. When `isRdma` is set to `auto`, then RDMA and non-RDMA devices can be added to the same resource pool. When an RDMA-capable device (like a Mellanox VF) is requested, then the RDMA resources are attached to the pod. When a non-RDMA device is allocated to a pod, then whatever normal resources are attached to the pod. I've tested the patch against an Intel X710 VF and a Mellanox Connect X-5 VF without noticeable issues.

I am not sure how much backward compatibility this project aims to provide. By moving `isRdma` from a `bool` to `string`, I am not sure if some existing configurations that contain `isRdma: true` or `isRdma: false` will need to be updated or not. If backwards compatibility is a higher concern, then perhaps it will be better to add an additional flag rather than change the type of the existing one.

## Questions
Is the use case presented a valid use case? Is the implementation approach mentioned acceptable, or would another way be preferred?

nlbrown2:
This patch enables an end user to provide either a single object or an array of selector objects for the `selectors` field in a single resource name. This enables one resource name to refer to devices with different or conflicting configurations/selectors.

To support multiple selector objects per resource name, the discovery of device attributes has to be changed/delayed slightly. This is because some selector values (like `isRdma`) change what information is learned about a `pciNetDevice`. When there are multiple selector objects per resource name, the values of these selectors aren't known until the device list is filtered.

The flow for a resource pool/server creation was:
1. Get all devices of a certain class (net device, accelerator, etc.). This occurs in the `resourceManager`'s `discoverHostDevices()` and device provider's `AddTargetDevices()` functions.
2. Convert device structs into more specific types defined in `types.go`. This is started in the `initServers()` function, which calls `GetDevices()` for each device provider. For the `netDeviceProvider`, this converts the devices into `PciNetDevice`s. During this process, a lot of device information is learned, such as: driver, vendor, device code, PCI address, vDPA/RDMA information, etc. For the vDPA or RDMA information, the value of the `selectors` field is used to determine if that information should be learned or not.
3. Filter the list of all devices using the `selectors` field to determine which devices belong under which resource names. Use the list of devices to create the resource pool & resource server.

As mentioned before, allowing multiple selector objects for a single resource name means that certain information about a device can't be known until some filtering has occurred. To accommodate this, some per-device information that is learned during step 2 is instead deferred to step 3:

1. same as above
2. Convert devices into more specific types with only a subset of what is learned today. Information that can be learned independent of selector values is learned, while selector-dependent information is left unknown (ex: RDMA or vDPA capability).
3. For each selector, filter the devices as normal. Once the RDMA and vDPA selectors are left, the values for those are used to learn that information about the remaining devices. Once the appropriate information is learned, the results are used to either include or exclude that device. Any `infoProviders` added to a device as a part of this process occurs during this step.

To see this new flow in action, `GetFilteredDevices` of the `netDeviceProvider` can be used as an example.

This PR is just a PoC of the multi-selector idea, showing how it may be implemented. I have not updated the tests to pass yet, but I have compiled and run this code in a lab with Mellanox CX-5's and Intel XL710's. Both NICs were able to be included under the same resource name, with the Mellanox NICs being configured with `isRdma: true` and the Intel XL710 NICs configured with `isRdma: false`.

Fixes #444


adrianchiris:
Hi Nathan, 

Apologies for the delay in review. I gave your PR a good read and id like to propose a slightly different approach.

I think we can end up with a simpler implementation while avoiding the change in handling vdpa and rdma selectors.

- device selectors is a list of selectors (technically this can be supported generically for all device types) if json is not a list, treat it as a list of 1
-  in resourceManager.initServers() , for each resource config in rm.configList call device provider GetFilteredDevices() N times, N being the number of device selector items in the list of device selectors.
-  in resourceManager.initServers(), filteredDevices will aggregate all devices from the above

thoughts ?

my motivation for the approach above is to avoid delayed initialization of some device providers /attributes.
that is, some are added at device creation and some during filtering (vpda/rdma).

nlbrown2:
Hi Adrian,

Thank you for taking the time to review this!

I like your suggested approach, and I think it will help prevent the delayed initialization problem. I think both `deviceProvider.GetDevices` and `deviceProvider.GetFilteredDevices` will need to be adjusted to know which selector object to apply, but that will be simpler than the approach used now.

Let me take some time this week to adopt this approach, and see if I run into any issues.

Thanks again!

nlbrown2:
Hello,

I am attaching a few files in a zip so that folks can use them to test this PR. I will ensure the updated docs will also use the syntax implemented by this PR, but I still need some time to prepare those and update this PR. With these files, folks who want to test this on various types of devices can do so without needing to wait for the doc changes.

Prerequisites:
Multus and SR-IOV CNI have been installed. VFs have been created.

Files:
1. net-attach-def.yaml: defines a Multus network attachment definition used by the example pods
2. sriovdp-configmap-base.yaml: basic config that groups Intel and Mellanox VFs under the `dpdk.org/nic` resource name. Note this won't work for the Mellanox VFs unless the pods that run are privileged. 
3. sriovdp-configmap-base-slice.yaml: The same config as (2), but with array/slice brackets. Can be used to show that the parsing logic is the same for single selector configurations.
4. sriovdp-configmap-separate.yaml: This should look familiar, as it is basically copy/paste from examples under the doc/ directory. This puts Intel VFs under the `intel_sriov_dpdk` name, and Mellanox VFs under the `mlnx_sriov_rdma` name. This will let the VFs work properly, but under different resource names.
5. sriovdp-configmap-new.yaml: this uses the new syntax to combine the Intel and Mellanox VFs under the `dpdk.org/nic` resource name. It correctly uses the `isRdma` attribute to ensure the pods work correctly.
6. sriovdp-configmap-duplicate.yaml: this duplicates the selectors of (5) to show that the plugin correctly handles the case when the same resource/VF falls into multiple selector objects under the same resource name.
7. pod.yaml: A sample pod definition that can be used to have the device plugin allocate VFs to a pod. The device plugin logs can show which resources (like `/dev/vfio/vfio`) are mounted into the container.
8. deployment.yaml: Puts pod.yaml into a deployment to easily scale the replicas up. Useful to help deploy a pod for every VF, if that is of interest.

Please let me know if you have any questions, and I plan to properly update this PR in the near future.
[sriovdp-test-configs.zip](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/files/11436874/sriovdp-test-configs.zip)

SchSeba (commented):
Looks great nice work!

I added some small comments.
but please add also a documentation update and unit tests for this new feature

cmd/sriovdp/manager_test.go
```diff
@@ -273,7 +273,7 @@
 					rc := &types.ResourceConfig{
 						ResourceName: "fake",
 						DeviceType:   types.NetDeviceType,
-						SelectorObj:  types.NetDeviceSelectors{},
+						SelectorObjs: types.NetDeviceSelectors{},
```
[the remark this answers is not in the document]
nlbrown2: Done. Please let me know if the test needs to be extended in any way.

pkg/accelerator/accelDeviceProvider.go
```diff
@@ -113,10 +113,12 @@ func (ap *accelDeviceProvider) GetFilteredDevices(devices []types.HostDevice, rc
...
 
 func (ap *accelDeviceProvider) ValidConfig(rc *types.ResourceConfig) bool {
-	_, ok := rc.SelectorObj.(*types.AccelDeviceSelectors)
-	if !ok {
-		glog.Errorf("unable to convert SelectorObj to AccelDeviceSelectors")
-		return false
+	for _, selector := range rc.SelectorObjs {
+		_, ok := selector.(*types.AccelDeviceSelectors)
+		if !ok {
+			glog.Errorf("unable to convert SelectorObj to AccelDeviceSelectors")
```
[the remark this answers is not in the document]
nlbrown2: Done.

pkg/auxnetdevice/auxNetDeviceProvider.go
```diff
@@ -90,9 +90,9 @@ func (ap *auxNetDeviceProvider) AddTargetDevices(devices []*ghw.PCIDevice, devic
 }
 
 //nolint:gocyclo
-func (ap *auxNetDeviceProvider) GetFilteredDevices(devices []types.HostDevice, rc *types.ResourceConfig) ([]types.HostDevice, error) {
+func (ap *auxNetDeviceProvider) GetFilteredDevices(devices []types.HostDevice, rc *types.ResourceConfig, selectorIndex int) ([]types.HostDevice, error) {
 	filteredDevice := devices
-	nf, ok := rc.SelectorObj.(*types.AuxNetDeviceSelectors)
+	nf, ok := rc.SelectorObjs[selectorIndex].(*types.AuxNetDeviceSelectors)
```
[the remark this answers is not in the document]
nlbrown2: Done. This actually has a bug as well, the `selectorIndex > len(rc.SelectorObjs)` needs to be `>=`.

pkg/accelerator/accelDeviceProvider.go
```diff
@@ -73,9 +73,9 @@ func (ap *accelDeviceProvider) AddTargetDevices(devices []*ghw.PCIDevice, device
 	return nil
 }
 
-func (ap *accelDeviceProvider) GetFilteredDevices(devices []types.HostDevice, rc *types.ResourceConfig) ([]types.HostDevice, error) {
+func (ap *accelDeviceProvider) GetFilteredDevices(devices []types.HostDevice, rc *types.ResourceConfig, selectorIndex int) ([]types.HostDevice, error) {
 	filteredDevice := devices
-	af, ok := rc.SelectorObj.(*types.AccelDeviceSelectors)
+	af, ok := rc.SelectorObjs[selectorIndex].(*types.AccelDeviceSelectors)
```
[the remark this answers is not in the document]
nlbrown2: Same as above.

nlbrown2:
I have updated the unit tests for the functionality change. Please let me know if there are any additional test cases I should add, or if any test cases should be re-worked for any reason.

I will push another update for the documentation changes later this week/early next week.

Thanks for your time, effort, and energy in reviewing this PR. I am excited to see where it will end up!

pkg/types/types.go
```diff
@@ -202,9 +202,11 @@ type DeviceProvider interface {
 	GetDiscoveredDevices() []*ghw.PCIDevice
 
 	// GetDevices runs through the Discovered Devices and returns a list of fully populated HostDevices according to the given ResourceConfig
-	GetDevices(*ResourceConfig) []HostDevice
+	GetDevices(*ResourceConfig, int) []HostDevice
```
adrianchiris, line 205: you could replace the int (index) parameter with interface{} passing in the selectorObj.
then you will not need to deal with index handling throughout the stack. this is also relevant for GetFIlteredDevices below.

An alternative way would be to have a single "DeviceConfig" struct which will contain all needed information and it would be passed around from device provider to device implementation (pciNetDevice, AuxNetDevice etc...)

Thoughts ?

personally i prefer the latter as we would then decouple user facing API from internal one.

pkg/factory/factory.go
```diff
@@ -177,37 +177,64 @@ func (rf *resourceFactory) GetDeviceProvider(dt types.DeviceType) types.DevicePr
...
 
-// GetDeviceFilter unmarshal the "selector" values from ResourceConfig and returns an instance of DeviceSelector based on
+// GetDeviceFilter unmarshal the "selector" values from ResourceConfig and returns a slice of *DeviceSelectors based on
 // DeviceType in the ResourceConfig
-func (rf *resourceFactory) GetDeviceFilter(rc *types.ResourceConfig) (interface{}, error) {
+func (rf *resourceFactory) GetDeviceFilter(rc *types.ResourceConfig) ([]interface{}, error) {
 	switch rc.DeviceType {
 	case types.NetDeviceType:
-		netDeviceSelector := &types.NetDeviceSelectors{}
+		netDeviceArray := make([]types.NetDeviceSelectors, 1)
```
adrianchiris, line 185: nit: netDeviceSelectors ? here and below

technically you are creating a slice (there are no arrays in go).
also we generally dont add the type in the name (e.g[ hungarian style](https://en.wikipedia.org/wiki/Hungarian_notation))

pkg/factory/factory.go
```diff
@@ -177,37 +177,64 @@ func (rf *resourceFactory) GetDeviceProvider(dt types.DeviceType) types.DevicePr
...
 
-		if err := json.Unmarshal(*rc.Selectors, netDeviceSelector); err != nil {
-			return nil, fmt.Errorf("error unmarshalling NetDevice selector bytes %v", err)
+		if err := json.Unmarshal(*rc.Selectors, &netDeviceArray[0]); err != nil {
```
adrianchiris, line 187: this is repeating itself for all types.

unmarshal selectors to either the first element of the slice or replacing the entire slice.
then finally moving it to a slice of `interface{}`

can we move this to a helper method and call it for all cases ?

you can try using [generics ](https://go.dev/doc/tutorial/generics) for it

pkg/types/types.go
nlbrown2: Hi Adrian,

I wish to understand your second idea in more detail. 

Right now, the `resourceManager` reads the config & forwards it to the device plugins. Are you suggesting that the `resourceManager` converts the config from a `resourceConfig` style object to a `deviceConfig`, where `deviceConfig` then gets piped through the stack, like the index currently does?

For this change, a single `resourceConfig` would likely generate multiple `deviceConfigs` (one for each selector) in order to avoid the same index issue, where parts of the stack need to be told which selector/data to read. Is this what you had in mind? Or is there a simpler approach I am not seeing?

I like the idea of decoupling user-facing changes from internal changes. If we decide to go this route, we will have to decide if that work should live in a separate PR, and if so, which PR should be completed first. I think a separate PR for separating the internal and user-facing APIs without any functionality changes makes sense.

pkg/factory/factory.go
nlbrown2: I wasn't sure if generics were allowed. I don't think I've seen them anywhere else in the code base, but maybe I just missed it.

I think generics will help this logic a lot. A question then: should `any` be preferred instead of `interface{}`? I think `any` is the new alias for `interface{}`, but I am not familiar with the differences (if `any`) between them.

cmd/sriovdp/manager_test.go
```diff
@@ -127,6 +127,63 @@ var _ = Describe("Resource manager", func() {
 				Expect(len(rm.configList)).To(Equal(2))
 			})
 		})
+		Context("when the multi-selector config reading is successful", func() {
+			var err error
+			BeforeEach(func() {
```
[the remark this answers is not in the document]
nlbrown2: Thanks for the idea to refactor the tests in another PR.

I am not sure what you mean by
>  e.g., merging the setup logic, the function under test call, and the assertion under the same `It`

Can you elaborate?

I think some smaller readability improvements would be to convert `Expect(len(thing)).To(Equal(number))` to `Expect(thing).To(HaveLen(number))`.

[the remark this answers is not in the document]
zeeke: > I think some smaller readability improvements would be to convert Expect(len(thing)).To(Equal(number)) to Expect(thing).To(HaveLen(number)).

That's a good point and can be implemented as is. Please do that.

A different way to implement the same test can be

```
It("when the multi-selector config reading is successful", func() {

			err := os.MkdirAll("/tmp/sriovdp", 0755)
			Expect(err).NotTo(HaveOccurred())

			err = os.WriteFile("/tmp/sriovdp/test_config", []byte(`{
						"resourceList": [{
								"resourceName": "intel_sriov_netdevice",
								"selectors": {
									"isRdma": false,
									"vendors": ["8086"],
									"devices": ["154c", "10ed"],
									"drivers": ["i40evf", "ixgbevf"]
								}
							},
							{
								"resourceName": "dpdk",
								"selectors": [{
									"vendors": ["8086"],
									"devices": ["154c", "10ed"],
									"drivers": ["vfio-pci"]
								}, {
									"vendors": ["15b3"],
									"devices": ["1018"],
									"drivers": ["mlx5_core"],
									"isRdma": true
								}]
							}
						]
					}`), 0644)
			Expect(err).NotTo(HaveOccurred())
			defer os.RemoveAll("/tmp/sriovdp")

			// Function under test
			err = rm.readConfig()
			Expect(err).NotTo(HaveOccurred())

			Expect(len(rm.configList)).To(Equal(2))
			Expect(len(rm.configList[0].SelectorObjs)).To(Equal(1))
			Expect(len(rm.configList[1].SelectorObjs)).To(Equal(2))
		})
```

It follows the [Arrange,Act,Assert](https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/) test style, and it's more clear to me what the test wants to verify.

BTW, refactoring these tests is out of this PR's scope, as it already contains significant changes.

Eoghan1232:
Hi @nlbrown2,

I was able to test this PR against an accelerator device type:

Standard config:
```
"resourceList": [{
                "resourceName": "intel_sriov_netdevice",
                "deviceType": "accelerator",
                "selectors": {
                    "vendors": ["8086"],
                    "devices": ["0d5d"]
                                   }
                         }
		]
```
Updated config:
```
resourceList": [{
			"resourceName": "intel_sriov_netdevice",
			"deviceType": "accelerator",
			"selectors": [{
				"vendors": ["8086"],
				"devices": ["0d5d"]
			}, {
				"vendors": ["8086"],
				"devices": [
				  "154c"
				],
				"drivers": ["vfio-pci"]
			}]
		}
	]
```


standard config result:
```
PCIDEVICE_INTEL_COM_INTEL_SRIOV_NETDEVICE=0000:4c:00.0
PCIDEVICE_INTEL_COM_INTEL_SRIOV_NETDEVICE_INFO={"0000:4c:00.0":{"generic":{"deviceID":"0000:4c:00.0"},"vfio":{"dev-mount":"/dev/vfio/306","mount":"/dev/vfio/vfio"}}}
```

updated config result:
```
PCIDEVICE_INTEL_COM_INTEL_SRIOV_NETDEVICE=0000:4c:00.0
PCIDEVICE_INTEL_COM_INTEL_SRIOV_NETDEVICE_INFO={"0000:4c:00.0":{"generic":{"deviceID":"0000:4c:00.0"},"vfio":{"dev-mount":"/dev/vfio/306","mount":"/dev/vfio/vfio"}}}
```

I can confirm that this accelerator functionality still works against your PR.

@adrianchiris @SchSeba

cmd/sriovdp/manager_test.go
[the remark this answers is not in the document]
nlbrown2: Thanks for the detailed example and insightful link! I agree such a change should be out of scope for this PR.

pkg/factory/factory.go
adrianchiris: i think its fine to use any

pkg/accelerator/accelDeviceProvider.go
```diff
@@ -73,9 +73,12 @@ func (ap *accelDeviceProvider) AddTargetDevices(devices []*ghw.PCIDevice, device
 	return nil
 }
 
-func (ap *accelDeviceProvider) GetFilteredDevices(devices []types.HostDevice, rc *types.ResourceConfig) ([]types.HostDevice, error) {
+func (ap *accelDeviceProvider) GetFilteredDevices(devices []types.HostDevice, rc *types.ResourceConfig, selectorIndex int) ([]types.HostDevice, error) {
```
nlbrown2, line 76: For this function, should there be any refactoring to adjust the cyclomatic complexity? I think the other `GetFilteredDevices` functions have `//nolint:gocyclo` for them (e.g., `netDeviceProvider.GetFilteredDevices`). So, I am in favor of adding the same exclusion to this function.

pkg/factory/factory.go
nlbrown2: Hi Adrian, `go.mod` lists the version of go as `1.17`. Generics need go `1.18` or higher. Is it fine to bump the go version to introduce generics? If so, should that be done in this PR or a separate one?

adrianchiris:
> These commits address review comments and add documentation, respectively. One outstanding comment is regarding using generics. The go.mod for this project lists go 1.17, which prevents use of generics until that is bumped to 1.18 or later.

Thx Nathan!

RE generics bit, yes this can be done in a separate PR. if you have bandwidth to do it. if not please create an issue on it  so it can  be picked up later.

pkg/accelerator/accelDeviceProvider.go
adrianchiris: im fine with adding it.

docs/dpdk/configMap-all-nics.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: sriovdp-config
+  namespace: kube-system
+data:
+  config.json: |
```
Eoghan1232, line 7: ah, or you could simply get the readme to point to here if you want to use selectors as a slice instead?

Eoghan1232:
> > These commits address review comments and add documentation, respectively. One outstanding comment is regarding using generics. The go.mod for this project lists go 1.17, which prevents use of generics until that is bumped to 1.18 or later.
> 
> Thx Nathan!
> 
> RE generics bit, yes this can be done in a separate PR. if you have bandwidth to do it. if not please create an issue on it so it can be picked up later.

#484 :)

README.md
```diff
@@ -180,75 +180,75 @@ This plugin creates device plugin endpoints based on the configurations given in
 {
     "resourceList": [{
             "resourceName": "intel_sriov_netdevice",
-            "selectors": {
+            "selectors": [{
```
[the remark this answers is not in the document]
nlbrown2: Done

nlbrown2:
The merge conflict should be resolved. Also, I changed the parsing logic within the factory to use generics following the merging of #484. Thanks for getting that resolved so quickly!

README.md
```diff
@@ -267,17 +267,17 @@ This plugin creates device plugin endpoints based on the configurations given in
...
 | "excludeTopology" | N        | Exclude advertising of device's NUMA topology                                                                                          | bool Default: "false"                                 | "excludeTopology": true                                                |
-| "selectors"       | N        | A map of device selectors. The "deviceType" value determines the "selectors" options.                                                  | json object as string Default: null                   | Example: "selectors": {"vendors": ["8086"],"devices": ["154c"]}        |
+| "selectors"       | N        | An array of device selector maps. The "deviceType" value determines the device selector options.                                                  | json object as string Default: null                   | Example: "selectors": [{"vendors": ["8086"],"devices": ["154c"]}]        |
```
adrianchiris, line 270: Also i think we should mention we support two forms here, 
a selector map or  a list of selector map WDYT ?

pkg/factory/factory.go
```diff
@@ -177,37 +177,41 @@ func (rf *resourceFactory) GetDeviceProvider(dt types.DeviceType) types.DevicePr
...
-		netDeviceSelector := &types.NetDeviceSelectors{}
-
-		if err := json.Unmarshal(*rc.Selectors, netDeviceSelector); err != nil {
-			return nil, fmt.Errorf("error unmarshalling NetDevice selector bytes %v", err)
+// parseObjectOrSlice unmarshal's the "Selector" values from the ResourceConfig into a slice of *DeviceSelectors.
+// Each *DeviceSelector has been converted to any before being returned. parseObjectOrSlice will parse both
+// kinds of valid "selector" values - a slice or a single object.
+func parseObjectOrSlice[O types.NetDeviceSelectors | types.AccelDeviceSelectors | types.AuxNetDeviceSelectors](
+	rc *types.ResourceConfig) ([]any, error) {
+	slice := make([]*O, 1)
```
adrianchiris, line 185: this should be : `make([]O, 1)` ?

else `selectors: null` is valid and it will be parsed without error.

pkg/factory/factory.go
```diff
@@ -177,37 +177,41 @@ func (rf *resourceFactory) GetDeviceProvider(dt types.DeviceType) types.DevicePr
...
-	case types.NetDeviceType:
-		netDeviceSelector := &types.NetDeviceSelectors{}
-
-		if err := json.Unmarshal(*rc.Selectors, netDeviceSelector); err != nil {
-			return nil, fmt.Errorf("error unmarshalling NetDevice selector bytes %v", err)
+// parseObjectOrSlice unmarshal's the "Selector" values from the ResourceConfig into a slice of *DeviceSelectors.
+// Each *DeviceSelector has been converted to any before being returned. parseObjectOrSlice will parse both
+// kinds of valid "selector" values - a slice or a single object.
+func parseObjectOrSlice[O types.NetDeviceSelectors | types.AccelDeviceSelectors | types.AuxNetDeviceSelectors](
+	rc *types.ResourceConfig) ([]any, error) {
```
adrianchiris, line 184: you can return here ([]*O, error) imo and it will still work in GetDeviceFilter below.

that way its a bit more explicit on what this does.

README.md
```diff
@@ -189,66 +189,66 @@ This plugin creates device plugin endpoints based on the configurations given in
         {
             "resourceName": "intel_sriov_dpdk",
             "resourcePrefix": "intel.com",
-            "selectors": {
+            "selectors": [{
```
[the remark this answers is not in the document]
nlbrown2: Hi Adrian, maybe I misremembered or misinterpreted the discussion from May 2nd. The meeting minutes for that day note "New way of providing selectors should be as a list going forward". I jotted down that the existing examples should be rewritten to use the list format to showcase its usage, encourage more people to use it (& test it), and be the new desired format.

I'm fine to keep the existing format as the preferred approach and encouraging users to only adopt a list when it makes sense for them.

README.md
```diff
@@ -300,14 +300,14 @@ This selector is applicable when "deviceType" is "netDevice"(note: this is defau
...
 | "linkTypes"    | N        | The link type of the net device associated with the PCI device           | `string` list Default: `null`                       | "linkTypes": ["ether"]                                                                           |
-| "isRdma"       | N        | Mount RDMA resources. Incompatible with vdpaType                         | `bool` values `true` or `false` Default: `false`    | "isRdma": `true`                                                                                 |
 | "ddpProfiles"  | N        | A map of device selectors                                                | `string` list Default: `null`                       | "ddpProfiles": ["GTPv1-C/U IPv4/IPv6 payload"]                                                   |
+| "isRdma"       | N        | Mount RDMA resources. Incompatible with vdpaType                         | `bool` values `true` or `false` Default: `false`    | "isRdma": `true`                                                                                 |
```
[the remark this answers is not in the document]
nlbrown2: The list here should match the list the selectors are applied in the code. Looking at the relevant function, we see `isRdma` is applied after `ddpProfiles`: https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/7e7f979087286ee950bd5ebc89d8bbb6723fc625/pkg/netdevice/netDeviceProvider.go#L152

pkg/factory/factory.go
nlbrown2: Thanks, I had not considered this case. I will have to see if `make([]0, 1)` is easy to convert into the expected function output, or if manually checking that `slice[0] != nil` is more straightforward. The solution should hopefully also handle a case like `selectors: [{ /* something valid */}, nil]` and so forth.

nlbrown2:
Thanks Adrian for the detailed review. I haven't had enough time today to address all of your comments, but plan to do so by the end of the week.

README.md
nlbrown2: Agreed.

docs/ddp/README.md
```diff
@@ -88,35 +88,35 @@ data:
         "resourceList": [
              {
                 "resourceName": "e800_default",
-                "selectors": {
+                "selectors": [{
```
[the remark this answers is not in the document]
nlbrown2: Leaving as-is until the README comment is resolved.

pkg/factory/factory.go
nlbrown2: I tried this change, and it did not compile. The error was: `cannot use interfaceArray (variable of type []any) as type []*O in return statement`. I could do the interface object conversion outside of this function, but I think it makes more sense to have that conversion logic here rather than in `GetDeviceFilter`.

Unless I misunderstood your suggestion, I think the return type should to remain `([]any, error)`.

pkg/factory/factory.go
nlbrown2: I was attempting to write an additional unit test for this `selectors: null` case, and ran into a small confusion. By `selectors: null` is that equivalent to an end user writing a full on config akin to:
```json
{
    "resourceList": [{
            "resourceName": "intel_sriov_netdevice",
            "selectors": "null"
        }]
}
``` 
or
```json
{
    "resourceList": [{
            "resourceName": "intel_sriov_netdevice",
            "selectors": null
        }]
}
```
? For the latter case, I am not sure how to encode the `null` into a Go program. I would want to use `nil`, but I'm not sure how to embed that into a JSON string/[]byte.

In at least the first case (a string with value `"null"`), it seems almost impossible to detect this case within this method. From the [json docs](https://pkg.go.dev/encoding/json#Unmarshal):

> The JSON null value unmarshals into an interface, map, pointer, or slice by setting that Go value to nil. Because null is often used in JSON to mean “not present,” unmarshaling a JSON null into any other Go type has no effect on the value and produces no error. 

When I run this case against the head of `master`, the log line is:
```
manager.go:95] unable to get SelectorObj from selectors list:'"null"' for deviceType: netDevice error: error unmarshalling NetDevice selector bytes json: cannot unmarshal string into Go value of type types.NetDeviceSelectors
```
As the log line shows, there isn't anything special about the `"null"` string compared to any other string value.

For the latter case, the pod crashes. Using the upstream code (without this PR), the logs are:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xd58630]

goroutine 1 [running]:
github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/factory.(*resourceFactory).GetDeviceFilter(0x40002aa600?, 0x4000876c60)
	/usr/src/sriov-network-device-plugin/.gopath/src/github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/factory/factory.go:187 +0x70
main.(*resourceManager).readConfig(0x40008d02a0)
	/usr/src/sriov-network-device-plugin/.gopath/src/github.com/k8snetworkplumbingwg/sriov-network-device-plugin/cmd/sriovdp/manager.go:92 +0x244
main.main()
	/usr/src/sriov-network-device-plugin/.gopath/src/github.com/k8snetworkplumbingwg/sriov-network-device-plugin/cmd/sriovdp/main.go:45 +0xc8
```

From this, I assume the issue is that `rc.Selectors` is a null pointer, which gets dereferenced as part of
```go
json.Unmarshal(*rc.Selectors, netDeviceSelector)
```

The exact same logs occur if the config map doesn't have any `"selectors"` field at all. This can be fixed by checking `rc.Selectors` is not `nil` both here and in the `resourceManager`.

The PR has the same behavior for these cases as the `master` branch.

README.md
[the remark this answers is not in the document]
nlbrown2: Fixed.

nlbrown2: Updated.

README.md
```diff
@@ -267,17 +267,17 @@ This plugin creates device plugin endpoints based on the configurations given in
...
 | "additionalInfo" | N | A map of map to add additional information to the pod via environment variables to devices                                             | json object as string Default: null  | Example: "additionalInfo": {"*": {"token": "3e49019f-412f-4f02-824e-4cd195944205"}} |
 
 Note: "resourceName" must be unique only in the scope of a given prefix, including the one specified globally in the CLI params, e.g. "example.com/10G", "acme.com/10G" and "acme.com/40G" are perfectly valid names.
 
 #### Device selectors
 
-The "deviceType" value determines which selectors are supported for that device. Each selector evaluated in order as listed in selector tables below.
+The "selectors" field accepts both a single object and an array of selector objects. When using the array syntax, each selector object is evaluated in the order listed in the array. A given selector object comprises of "selectors".  The "deviceType" value determines which selectors are supported for that device. Each selector evaluated in order as listed in selector tables below.
```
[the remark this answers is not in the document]
nlbrown2: Updated.

README.md
```diff
@@ -405,13 +405,13 @@ Usage of ./sriovdp:
...
+This plugin does not bind or unbind any driver to any device whether it's PFs or VFs. It also doesn't create virtual functions either. Usually, the virtual functions are created at boot time when kernel module for the device is loaded. Same with SFs. Required device drivers could be loaded on system boot-up time by allow-listing/deny-listing the right modules. But plugin needs to be aware of the driver type of the resources (i.e. devices) that it is registering as K8s extended resource so that it's able to create appropriate Device Specs for the requested resource.
 
 For example, if the driver type is uio (i.e. igb_uio.ko) then there are specific device files to add in Device Spec. For vfio-pci, device files are different. And if it is Linux kernel network driver then there is no device file to be added.
 
-The idea here is, user creates a resource config for each resource pool as shown in [Config parameters](#config-parameters) by specifying the resource name, a list resource "selectors".
+The idea here is, user creates a resource config for each resource pool as shown in [Config parameters](#config-parameters) by specifying the resource name and a list of resource "selectors".
```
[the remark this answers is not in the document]
nlbrown2: Updated this and below paragraph which mentions "list of selectors"

README.md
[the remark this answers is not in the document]
SchSeba: From my point of view I will like to start killing the single selector and have in the examples the full potential of the device plugin config.

I think we can add a small section to say that is possible to use one selector (without the `[]`) but all the others should have the list

README.md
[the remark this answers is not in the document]
SchSeba: :+1:

pkg/factory/factory.go
SchSeba: I think we should only cover the nil case and not the `"nil"` one that should let the device plugin crash..

README.md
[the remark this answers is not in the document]
Eoghan1232: I had mentioned previously to have 2 examples, one as we have now, without the ``[]`` and one with the ``[]``.
I am okay to move forward with promoting the use of it, however I would still like a small example and a disclaimer that we still support not using it.

pkg/factory/factory.go
adrianchiris: yes ! you are right, lets keep it as is.
(tried it myself, current impl makes more  sense)

pkg/factory/factory.go
adrianchiris: agree, lets handle it in a different PR.
some checks will need to be added to prevent these cases.

README.md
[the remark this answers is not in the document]
adrianchiris: i dont think there is this strict requirement where the order should match the order applied in the code.
moreover the order is not expected to affect the end result.

in any case, if this bothered you im fine with keeping this change although not really related to this PR.

README.md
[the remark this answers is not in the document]
adrianchiris: OK,
here are my takes from this.

1. update all examples to use the list form of selectors
2. add example for the non list form as well
3. documentation should mention that we support both forms and that  the list form is preferred

agreed ?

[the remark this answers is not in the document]
Eoghan1232: +1

README.md
```diff
@@ -267,17 +267,32 @@ This plugin creates device plugin endpoints based on the configurations given in
...
 | "additionalInfo" | N | A map of map to add additional information to the pod via environment variables to devices                                             | json object as string Default: null  | Example: "additionalInfo": {"*": {"token": "3e49019f-412f-4f02-824e-4cd195944205"}} |
 
 Note: "resourceName" must be unique only in the scope of a given prefix, including the one specified globally in the CLI params, e.g. "example.com/10G", "acme.com/10G" and "acme.com/40G" are perfectly valid names.
 
 #### Device selectors
 
-The "deviceType" value determines which selectors are supported for that device. Each selector evaluated in order as listed in selector tables below.
+The "selectors" field accepts both a single object and a list of selector objects. When using the array syntax, each selector object is evaluated in the order listed in the array. For example, a single object would look like:
```
adrianchiris, line 277: nit: array -> list

nlbrown2: Thanks, addressed.

README.md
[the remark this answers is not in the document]
nlbrown2: All 3 points should be adopted now. Please let me know what you think.

adrianchiris (approved):
LGTM ! 🥳

Eoghan1232 (approved):

Eoghan1232:
lgtm :)

PR #479, opened 2023-04-13
Derive link type from PF if the interface name is not available

almaslennikov:
in case a workload is already consuming a VF and sriov-device-plugin is restarted, when this selector is specified, the device plugin will fail to get the VF link type since getting this information requires the VF netdev. The netdev does not exist in the host namespace as its assigned to pod. This causes sriov-network-device-plugin to fail adding this VF to resource pool causing it to report the wrong number of resources.

To fix this, derive the link type of the device from its PF, if the netdev itself is not available.


pkg/devices/gen_net.go
```diff
@@ -82,8 +82,13 @@ func NewGenNetDevice(deviceID string, dt types.DeviceType, isRdma bool) (*GenNet
 	}
 
 	linkType := ""
-	if len(ifName) > 0 {
-		la, err := utils.GetNetlinkProvider().GetLinkAttrs(ifName)
+	linkTypeProviderDevice := ifName
```
adrianchiris, line 85: Hey,

can you add a test for this one ?

pkg/devices/gen_net.go
```diff
@@ -82,8 +82,13 @@ func NewGenNetDevice(deviceID string, dt types.DeviceType, isRdma bool) (*GenNet
...
+	// If interface name is not available, derive link type from PF
+	if linkTypeProviderDevice == "" {
+		linkTypeProviderDevice = pfName
+	}
+	if linkTypeProviderDevice != "" {
```
SchSeba, line 90: I don't think we need this if state here.
we have two cases only
1. vf exist use vf
2. vf doesn't exist is pf 

there is no other case so it can't be empty

adrianchiris: There is, in case of VF passthrough, and device plugin runs in VM.

we have no PF and VF Ifname may not exist if its already assigned to a container

pkg/devices/gen_net.go
almaslennikov: Done

e0ne (approved):

adrianchiris (approved):
Thx for working on it @almaslennikov  ! LGTM

pkg/devices/gen_net.go
SchSeba: right!

SchSeba (approved):

Issue #473, opened 2023-02-28
PR456 leftover - address remaining open comment as followup


adrianchiris:
@SchSeba it would be great if you could take this one.

essentially its updating comment of `GetEnvVal` function to reflect what it does now (returns information to be exposed via env var)

adrianchiris:
#456

PR #480, opened 2023-04-23
Update GetEnvVal docstring

adrianchiris:
As recently the implementation changed
to expose additional device informaiton,
we should align comment to describe new
behaviour.


adrianchiris:
#473

SchSeba (approved):

SchSeba:
Thanks for heling with this!

PR #484, opened 2023-05-23
bump go version to 1.20

Eoghan1232:
bumping go version to 1.20.
#475 needs a go version greater than 1.17, SRIOV CNI is go 1.20, so updating to stay in line.

all these passed:
```
make build
make image
go test ./...
```


Eoghan1232:
@SchSeba @adrianchiris PTAL

adrianchiris (approved):

PR #487, opened 2023-06-02
Add `acpiIndex` selector

ian-howell:


ian-howell:
Somehow broke this one, will try again

PR #488, opened 2023-06-02
Add `acpiIndex` selector

ian-howell:


SchSeba (commented):
this looks great! nice work!

ian-howell:
Fixed the merge conflicts.

@Eoghan1232 @adrianchiris @SchSeba Gentle nudge to request reviews when you find time. Thanks in advance!

pkg/netdevice/netDeviceProvider.go
```diff
@@ -122,6 +122,13 @@ func (np *netDeviceProvider) GetFilteredDevices(devices []types.HostDevice,
 		}
 	}
 
+	// filter by acpiIndexes list
```
zeeke, line 125: We need a refactor here. The smallest that comes to my mind is 
a. move each if block to a package private function like:

```
func filterByVendor(nf *types.NetDeviceSelectors, rf types.ResourceFactory, devices []types.HostDevice) []types.HostDevice {
if nf.Vendors != nil && len(nf.Vendors) > 0 {
		if selector, err := rf.GetSelector("vendors", nf.Vendors); err == nil {
			return selector.Filter(filteredDevice)
		}
	}
}
```

b. quick and dirty: remove all `x != nil` as `len(x) == 0`  
https://go.dev/play/p/L7gVCSUcHZu

c. Rethink a different object structure than ResourceFactory and Selector, moving some responsibility. IMO this should be addressed in a different PR

ian-howell: Hi @zeeke! Thanks for the review.

I went ahead and implemented your "option B" from above. I also moved the filtering into an abstracted `FilterBySelector`. Let me know if this solution looks ok.

zeeke: Even better than what I was thinking about. Thanks!

zeeke (approved):

pkg/types/types.go
```diff
@@ -106,9 +106,10 @@ type ResourceConfig struct {
 
 // DeviceSelectors contains common device selectors fields
 type DeviceSelectors struct {
-	Vendors []string `json:"vendors,omitempty"`
-	Devices []string `json:"devices,omitempty"`
-	Drivers []string `json:"drivers,omitempty"`
+	Vendors     []string `json:"vendors,omitempty"`
+	Devices     []string `json:"devices,omitempty"`
+	Drivers     []string `json:"drivers,omitempty"`
+	AcpiIndexes []string `json:"acpiIndexes,omitempty"`
```
adrianchiris, line 112: this probably needs to be under : GenericPciDeviceSelectors
we dont want it for Auxiliary device ATM (also i dont know if its even supported for it) hence lets limit it for now to PCI devices

pkg/types/types.go
```diff
@@ -93,7 +93,7 @@ var SupportedVdpaTypes = map[VdpaType]string{
 
 // ResourceConfig contains configuration parameters for a resource pool
 type ResourceConfig struct {
-	// optional resource prefix that will overwrite	global prefix specified in cli params
+	// optional resource prefix that will overwrite global prefix specified in cli params
```
adrianchiris, line 96: unrelated change

pkg/devices/host.go
```diff
@@ -31,6 +31,7 @@ type HostDeviceImpl struct {
 	vendorID   string
 	deviceCode string
 	driver     string
+	acpiIndex  string
```
adrianchiris, line 34: needs to be in `GenPciDevice`

pkg/factory/factory.go
```diff
@@ -100,6 +102,15 @@ func (rf *resourceFactory) GetSelector(attr string, values []string) (types.Devi
 	}
 }
 
+func (rf *resourceFactory) FilterBySelector(attr string, values []string, filteredDevices []types.HostDevice) []types.HostDevice {
```
adrianchiris, line 105: nits:
attr -> selectorName
filteredDevices -> devicesToFilter or just devices.

README.md
```diff
@@ -320,6 +320,7 @@ These selectors are applicable when "deviceType" is "netDevice" (note: this is d
 | "devices"      | N        | Target Devices' device Hex code as string                                | `string` list Default: `null`                       | "devices": ["154c", "1889", "1018"]                                                              |
 | "drivers"      | N        | Target device driver names as string                                     | `string` list Default: `null`                       | "drivers": ["vfio-pci"]                                                                          |
 | "pciAddresses" | N        | Target device's pci address as string                                    | `string` list Default: `null`                       | "pciAddresses": ["0000:03:02.0"]                                                                 |
+| "acpiIndexes"  | N        | Target device's acpi index as string                                     | `string` list Default: `null`                       | "acpiIndexes": ["101"]                                                                           |
```
adrianchiris, line 323: with how you implemented it today, this selector is available for auxiliay network devices  and accelerators as well.

acpiIndex selector should appear in all relevant sections.
(but see my previous comments first as i think we need to have this one for accelerator and network devices only for now)

README.md
```diff
@@ -440,10 +441,11 @@ The device plugin will initially discover all PCI network resources in the host
 2. "devices"      - The device hex code of device
 3. "drivers"      - The driver name the device is registered with
 4. "pciAddresses" - The pci address of the device in BDF notation (if device type is accelerator or netDevice)
-5. "auxTypes"     - The type of auxiliary network device (if device type is auxNetDevice).
-6. "pfNames"      - The Physical function name
-7. "rootDevices"  - The Physical function PCI address
-8. "linkTypes"    - The link type of the net device associated with the PCI device
+5. "acpiIndexes"  - The acpi index of the device
```
adrianchiris, line 444: it seems to me we need a bit more info about acpi index, when its available etc...

pkg/types/types.go
SchSeba: I just checked and to use scalable function inside a VM you can use VDPA so I think it's fine to move this to the PCI device section

https://metonymical.hatenablog.com/entry/2021/07/28/201833

pkg/utils/utils.go
```diff
@@ -382,6 +382,24 @@ func GetDriverName(pciAddr string) (string, error) {
 	return filepath.Base(driverInfo), nil
 }
 
+// GetAcpiIndex returns the ACPI index attached to a pci device from its pci address
+func GetAcpiIndex(pciAddr string) (string, error) {
+	acpiIndexLink := filepath.Join(sysBusPci, pciAddr, "acpi_index")
```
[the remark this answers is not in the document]
SchSeba: This is inside a virtual machine 

https://libvirt.org/formatdomain.html#network-interfaces

README.md
SchSeba: I think we should support this for network and accelerators I don't think there is a use-case for this on auxiliary devices

ian-howell: Added to the accelerators section

pkg/utils/utils.go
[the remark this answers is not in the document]
ian-howell: I'm not certain what actually causes a machine to get an acpi_index. As SchSheba pointed out, this is inside of a libvirt VM, (or more specifically for my case, a kubevirt VM using [this field](https://github.com/kubevirt/kubevirt/blob/main/staging/src/kubevirt.io/api/core/v1/schema.go#L1227-L1231)).

pkg/types/types.go
ian-howell: Moved the PCI device section

pkg/types/types.go
ian-howell: reverted

SchSeba (approved):
/LGTM

Eoghan1232 (approved):

Eoghan1232:
@adrianchiris are you okay for this to be merged?

adrianchiris (approved):

PR #489, opened 2023-06-04
CDI implementation

e0ne:


cmd/sriovdp/manager.go
```diff
@@ -17,10 +17,11 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"os"
-
+	"github.com/container-orchestrated-devices/container-device-interface/pkg/cdi"
```
adrianchiris, line 20: nit: import order

cmd/sriovdp/manager.go
```diff
@@ -117,19 +119,76 @@ func (rm *resourceManager) initServers() error {
...
+					ContainerEdits: containerEdit,
+				}
+				cdiSpec.Devices = append(cdiSpec.Devices, device)
+			}
+
+			if err != nil {
+				glog.Errorf("initServers(): error getting filtered devices for config %+v: %q", rc, err)
+			}
+
+			cdiCache, err := cdi.NewCache()
```
adrianchiris, line 180: you should use Registry to interact with CDI

e.g

```
	err := cdi.GetRegistry().SpecDB().WriteSpec(spec, specName)
	if err != nil {
		return nil, err
	}
```

from cdi lib:
```
// GetRegistry returns the CDI registry. If any options are given, those
// are applied to the registry.
func GetRegistry(options ...Option) Registry {
```

adrianchiris: is there a reason why you used Cache directly ?

cmd/sriovdp/manager.go
```diff
@@ -117,19 +119,76 @@ func (rm *resourceManager) initServers() error {
...
 		}
+
 		rPool, err := rm.rFactory.GetResourcePool(rc, filteredDevices)
 		if err != nil {
 			glog.Errorf("initServers(): error creating ResourcePool with config %+v: %q", rc, err)
 			return err
 		}
+
+		if rm.useCdi {
+
```
adrianchiris, line 137: nit: remove newline

cmd/sriovdp/manager.go
```diff
@@ -117,19 +119,76 @@ func (rm *resourceManager) initServers() error {
...
+			cdiDevices := make([]specs.Device, 0)
+			cdiSpec := specs.Spec{
+				Version: "0.5.0",
```
adrianchiris, line 140: you can use `CurrentVersion` from CDI specs package

cmd/sriovdp/manager.go
```diff
@@ -117,19 +119,76 @@ func (rm *resourceManager) initServers() error {
...
+
 		rPool, err := rm.rFactory.GetResourcePool(rc, filteredDevices)
 		if err != nil {
 			glog.Errorf("initServers(): error creating ResourcePool with config %+v: %q", rc, err)
 			return err
 		}
+
+		if rm.useCdi {
+
+			cdiDevices := make([]specs.Device, 0)
```
adrianchiris, line 138: nit: you can initialize with a capacity  of `len(filteredDevices)`

cmd/sriovdp/manager.go
```diff
@@ -117,19 +119,76 @@ func (rm *resourceManager) initServers() error {
...
+					glog.Errorf("Unable to create environment variables: %s", err)
+					return fmt.Errorf("Unable to create environment variables")
+				}
+				for name, value := range envs {
+					containerEdit.Env = append(containerEdit.Env, name+"="+value)
+				}
+
+				for i, spec := range dev.GetDeviceSpecs() {
+					glog.Infof("Checking device spec (%d): %s", i, spec.String())
+					deviceNode := specs.DeviceNode{
```
adrianchiris, line 162: can you rename either specs to cdiSpecs in imports or spec in the range statement above. its confusing to read as the two refer to two different things.

cmd/sriovdp/manager.go
```diff
@@ -117,19 +119,76 @@ func (rm *resourceManager) initServers() error {
...
+						Path:        spec.ContainerPath,
```
adrianchiris, line 163: there are both container path and host path (and permissions as well) in device plugin device specs.

cmd/sriovdp/manager.go
```diff
@@ -117,19 +119,76 @@ func (rm *resourceManager) initServers() error {
...
 			continue
 		}
+
 		rPool, err := rm.rFactory.GetResourcePool(rc, filteredDevices)
 		if err != nil {
 			glog.Errorf("initServers(): error creating ResourcePool with config %+v: %q", rc, err)
 			return err
 		}
+
+		if rm.useCdi {
```
adrianchiris, line 136: can we move this logic to a cdi pkg ?

pkg/resources/server.go
```diff
@@ -253,6 +285,14 @@ func (rs *resourceServer) Start() error {
 		}
 	}
 
+	cdiCache, err := cdi.NewCache()
```
adrianchiris, line 288: you can pass in an instance of cache from manager  in NewResourceServer method.

pkg/resources/server.go
```diff
@@ -118,18 +122,46 @@ func (rs *resourceServer) NotifyRegistrationStatus(ctx context.Context,
 func (rs *resourceServer) Allocate(ctx context.Context, rqt *pluginapi.AllocateRequest) (*pluginapi.AllocateResponse, error) {
 	glog.Infof("Allocate() called with %+v", rqt)
 	resp := new(pluginapi.AllocateResponse)
+
+	err := rs.cdiCache.Refresh()
```
adrianchiris, line 126: why do we  need to refresh here ?  (when cdi is not enabled we dont, but i also dont see why we need to refresh when CDI is enabled)

pkg/resources/server.go
```diff
@@ -118,18 +122,46 @@ func (rs *resourceServer) NotifyRegistrationStatus(ctx context.Context,
...
 			glog.Errorf("failed to get environment variables for device IDs %v: %v", container.DevicesIDs, err)
 			return nil, err
 		}
-		containerResp.Envs = envs
 
-		containerResp.Mounts = rs.resourcePool.GetMounts(container.DevicesIDs)
+		if rs.useCdi {
+			containerResp.Annotations = make(map[string]string, 0)
+
+			for _, id := range container.DevicesIDs {
```
adrianchiris, line 143: device IDs to container annotation can be a function/method in cdi pkg (related to previous comment)

pkg/resources/server.go
```diff
@@ -118,18 +122,46 @@ func (rs *resourceServer) NotifyRegistrationStatus(ctx context.Context,
...
+						cdiDevice = rs.cdiCache.GetDevice(device)
+					}
+				}
+				if cdiDevice != nil {
+					// TODO: (e0ne): add support of several devices
+					containerResp.Annotations["cdi.k8s.io/"+strings.ReplaceAll(cdiDevice.GetSpec().Kind, "/", "_")] = cdiDevice.GetSpec().Kind + "=" + cdiDevice.Name
+				} else {
+					glog.Errorf("cant find device: %s", id)
+				}
+
```
adrianchiris, line 157: nit: remove newline

adrianchiris (commented):
please note that device plugin daemonset would need to mount cdi dir
(/var/run/cdi) IIRC

cmd/sriovdp/main.go
```diff
@@ -33,6 +33,10 @@ func flagInit(cp *cliParams) {
 		"JSON device pool config file location")
 	flag.StringVar(&cp.resourcePrefix, "resource-prefix", "intel.com",
 		"resource name prefix used for K8s extended resource")
+	flag.BoolVar(&cp.useCdi, "use-cdi", true,
+		"Use Container Device Interface to expose devices in containers")
+	flag.BoolVar(&cp.systemd, "systemd", false,
```
moshe010, line 38: do you think adding the systemd service to sriov-device-plugin is better that add it to SR-IOV operator?

cmd/sriovdp/manager.go
```diff
@@ -100,6 +106,77 @@ func (rm *resourceManager) readConfig() error {
...
+	}
+	return nil
+}
+
+func (rm *resourceManager) createCdiSpec() error {
+	glog.Infof("number of config: %d\n", len(rm.configList))
+	deviceAllocated := make(map[string]bool)
+	for _, rc := range rm.configList {
+		// Create new ResourcePool
+		glog.Infof("")
```
moshe010, line 138: why this is needed?

cmd/sriovdp/manager.go
```diff
@@ -131,11 +208,21 @@ func (rm *resourceManager) initServers() error {
...
 		}
+
 		rPool, err := rm.rFactory.GetResourcePool(rc, filteredDevices)
 		if err != nil {
 			glog.Errorf("initServers(): error creating ResourcePool with config %+v: %q", rc, err)
 			return err
 		}
+
+		if rm.useCdi {
+			err = cdi.CreateCDISpec(rm.resourcePrefix, filteredDevices, rPool)
```
moshe010, line 219: I see you regenerated it because the logic is in device plugin and you don't want to duplicate it

deployments/cdi/sriovdp-daemonset.yaml
```diff
@@ -0,0 +1,119 @@
...
+            - mountPath: /etc/pcidp/
+              name: config-volume
+              readOnly: true
+            - mountPath: /var/run/k8s.cni.cncf.io/devinfo/dp
+              name: device-info
+            - mountPath: /etc/cdi/
+              name: default-cdi
+            - mountPath: /var/run/cdi
+              name: dynamic-cdi
+            - mountPath: /host//etc/pcidp/
```
moshe010, line 67: remove extra /

pkg/cdi/cdi.go
```diff
@@ -0,0 +1,74 @@
...
+	//"strings"
+
+	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
+)
+
+// CreateCDISpec creates CDI spec file with specified devices
+func CreateCDISpec(resourcePrefix string, filteredDevices []types.HostDevice, rPool types.ResourcePool) error {
+	cdiDevices := make([]cdiSpecs.Device, 0)
+	cdiSpec := cdiSpecs.Spec{
+		Version: "0.5.0",
```
moshe010, line 30: please use constant for version  in https://github.com/container-orchestrated-devices/container-device-interface/blob/v0.5.0/specs-go/config.go#L6

pkg/cdi/cdi.go
```diff
@@ -0,0 +1,74 @@
...
+		for name, value := range envs {
+			containerEdit.Env = append(containerEdit.Env, name+"="+value)
+		}
+
+		for i, spec := range dev.GetDeviceSpecs() {
+			glog.Infof("Checking device spec (%d): %s", i, spec.String())
+			deviceNode := cdiSpecs.DeviceNode{
+				Path:        spec.ContainerPath,
+				HostPath:    spec.HostPath,
+				Permissions: "rwm",
```
moshe010, line 55: did you check if the m in Permissions is needed?

cmd/sriovdp/main.go
e0ne: Since we're implementing CDI support as a part of device plugin project, I think it's reasonable it have systemd service implementation in the same place. SR-IOV Operatir will be responsible for service provisioning

adrianchiris: WDYT about having a separate binary for it under cmd ?

also the flag name systemd is confusing.
what this does is: it runs once to re-generate CDI configuration.

cmd/sriovdp/manager.go
```diff
@@ -17,23 +17,29 @@ package main
...
 
 	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/factory"
 	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
 	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/utils"
 )
 
 const (
-	socketSuffix = "sock"
+	socketSuffix  = "sock"
+	hostConfigDir = "/host/etc/pcidp/"
```
adrianchiris, line 35: This needs to be configurable via flag

cmd/sriovdp/manager.go
```diff
@@ -100,6 +106,77 @@ func (rm *resourceManager) readConfig() error {
 	return nil
 }
 
+func (rm *resourceManager) storeConfig() error {
```
adrianchiris, line 109: storing the last config the daemon is using is also a nice debug feature IMO.
can you split this to a separate PR ? 

and should be optionally executed based on config from user via flag (e.g hostConfigDir was provided)

cmd/sriovdp/manager.go
```diff
@@ -100,6 +106,77 @@ func (rm *resourceManager) readConfig() error {
...
+		glog.Errorf("storeConfig(): error during copy config file to host: %v", err)
+		return err
+	}
+	return nil
+}
+
+func (rm *resourceManager) createCdiSpec() error {
+	glog.Infof("number of config: %d\n", len(rm.configList))
+	deviceAllocated := make(map[string]bool)
+	for _, rc := range rm.configList {
```
adrianchiris, line 136: This was copied for L184, why not just put into helper function to reduce code dupl ?

deployments/cdi/systemd/cdi-spec.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
```
adrianchiris, line 1: this is just an example, can you add some comments explaining this ?
or better yet add a README here for it.

in actual deployment you would want also to have an "After" key as this would need to run after the service which configures SR-IOV.

pkg/cdi/cdi.go
e0ne: it doesn't needed for VFs. I'll check with SFs and create a separate PR to fix it over the whole project

moshe010: Did you check with and without RDMA CM?

e0ne: yes, I checked both for VFs

cmd/cdi-service/main.go
```diff
@@ -0,0 +1,42 @@
+// Copyright 2018 Intel Corp. All Rights Reserved.
```
moshe010, line 1: update copyright as this is new file

deployments/cdi/systemd/cdi-spec.service
```diff
@@ -0,0 +1,12 @@
...
+Description=Creates CDI spec file
+# Removal of this file signals firstboot completion
+# This service is used to configure the SR-IOV VFs on NICs
+Wants=network-pre.target
+Before=network-pre.target
```
moshe010, line 6: why it network-pre.target ?
should it be after network so that we will be sure that the driver is loaded?

Makefile
```diff
@@ -73,6 +74,12 @@ build: $(BUILDDIR)/$(BINARY_NAME) | ; $(info Building $(BINARY_NAME)...) @ ## Bu
 $(BUILDDIR)/$(BINARY_NAME): $(GOFILES) | $(BUILDDIR)
 	@cd $(BASE)/cmd/$(BINARY_NAME) && CGO_ENABLED=0 go build $(LDFLAGS) -o $(BUILDDIR)/$(BINARY_NAME) -tags no_openssl -v
 
+build-cdi: $(BUILDDIR)/$(CDI_BINARY_NAME) | ; $(info Building $(CDI_BINARY_NAME)...) @ ## Build SR-IOV CDI Service
```
adrianchiris, line 77: nit: build-cdi-service ?

deployments/cdi/systemd/cdi-spec.service
e0ne: according to https://systemd.io/NETWORK_ONLINE/: 

> network-pre.target is used to order services before any network interfaces start to be configured. Its primary purpose is for usage with firewall services that want to establish a firewall before any network interface is up.

e0ne:
@SchSeba  could you please take a look on this PR once you have a time?

cmd/sriovdp/manager.go
```diff
@@ -16,64 +16,114 @@ package main
...
-	socketSuffix = "sock"
+	defaultConfig = "/etc/pcidp/config.json"
+	socketSuffix  = "sock"
 )
 
-type cliParams struct {
-	configFile     string
-	resourcePrefix string
+// CliParams presents CLI parameters for SR-IOV Network Device Plugin
+type CliParams struct {
```
zeeke, line 40: Is there any particular reason to make all symbols in this file public?

It's not a big issue, but having those private allows changing signatures in a safer manner

e0ne: I need to revert this change. It will be required for a follow-up PR

deployments/cdi/README.md
```diff
@@ -0,0 +1,16 @@
+# SR-IOV Network Device Plugin in CDI mode deployment
+
+SR-IOV Network Device Plugin supports [Container Device Interface (CDI)](https://github.com/container-orchestrated-devices/container-device-interface).
+
+To enable CDI mode should be started with `--use-cdi` CLI argument. There are also additional volume mounts required
```
zeeke, line 5: This is not congruent with [FlagInit()](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/489/files#diff-f1a4e99ad0d1a580dda8274ab56d83450f41088bae0f4d0df0808d03c851d20dR64) where it looks like the feature is enabled by default.
I don't know if it's better to have it enabled by default, but in that case, we have to document the opposite usage (i.e. how to disable it)

images/entrypoint.sh
```diff
@@ -7,7 +7,7 @@ LOG_DIR=""
 LOG_LEVEL=10
 RESOURCE_PREFIX=""
 CONFIG_FILE=""
-CLI_PARAMS=""
+CLI_PARAMS="-v $LOG_LEVEL"
```
adrianchiris, line 10: why was this moved ?

this breaks setting rog level as cmdline parsing happens at L25

cmd/sriovdp/manager.go
```diff
@@ -16,59 +16,109 @@ package main
...
+// FlagInit parse command line flags
+func FlagInit(cp *cliParams) {
+	flag.StringVar(&cp.configFile, "config-file", defaultConfig,
+		"JSON device pool config file location")
+	flag.StringVar(&cp.resourcePrefix, "resource-prefix", "intel.com",
+		"resource name prefix used for K8s extended re"+
+			"source")
+	flag.BoolVar(&cp.useCdi, "use-cdi", false,
+		"Use Container Device Interface to expose devices in containers")
+	flag.StringVar(&cp.hostConfigDir, "host-config-dir", "/host/etc/pcidp/",
```
adrianchiris, line 66: we dont need this one for now

adrianchiris: maybe we would want to have : cdiSpecDir, but for now mounting to standard CDI paths is enough (up 2 u if you want to add)

cmd/sriovdp/manager.go
```diff
@@ -16,59 +16,109 @@ package main
...
 	pluginWatchMode bool
 	rFactory        types.ResourceFactory
-	configList      []*types.ResourceConfig
+	ConfigList      []*types.ResourceConfig
 	resourceServers []types.ResourceServer
 	deviceProviders map[types.DeviceType]types.DeviceProvider
 }
 
-func newResourceManager(cp *cliParams) *resourceManager {
+// FlagInit parse command line flags
```
adrianchiris, line 57: do we really need to move this one ?

adrianchiris: Generally i think its good to move resource manager out of main to its own package under pkg, WDYT ?

we are more than half way there. ...

cmd/sriovdp/manager.go
```diff
@@ -16,59 +16,109 @@ package main
...
 }
 
-type resourceManager struct {
+// ResourceManager manages resources for SR-IOV Network Device Plugin binaries
+type ResourceManager struct {
 	cliParams
 	pluginWatchMode bool
 	rFactory        types.ResourceFactory
-	configList      []*types.ResourceConfig
+	ConfigList      []*types.ResourceConfig
```
adrianchiris, line 52: why does it need to be public ?

cmd/sriovdp/manager.go
```diff
@@ -100,13 +150,85 @@ func (rm *resourceManager) readConfig() error {
 	return nil
 }
 
-func (rm *resourceManager) initServers() error {
+// StoreConfig saves SR-IOV Device Plugin config into the file
+func (rm *ResourceManager) StoreConfig() error {
```
adrianchiris, line 154: why do we need to store config on host ?

cmd/sriovdp/manager.go
```diff
@@ -100,13 +150,85 @@ func (rm *resourceManager) readConfig() error {
...
+	_, err = io.Copy(destination, source)
+	if err != nil {
+		glog.Errorf("storeConfig(): error during copy config file to host: %v", err)
+		return err
+	}
+	return nil
+}
+
+// CreateCdiSpec creates CDI definition for container runtime
+func (rm *ResourceManager) CreateCdiSpec() error {
```
adrianchiris, line 179: This is not used, please remove

cmd/sriovdp/manager.go
```diff
@@ -131,11 +253,21 @@ func (rm *resourceManager) initServers() error {
...
 		}
+
 		rPool, err := rm.rFactory.GetResourcePool(rc, filteredDevices)
 		if err != nil {
 			glog.Errorf("initServers(): error creating ResourcePool with config %+v: %q", rc, err)
 			return err
 		}
+
+		if rm.useCdi {
+			err = cdi.CreateCDISpec(rm.resourcePrefix, filteredDevices, rPool)
```
adrianchiris, line 264: i think this should be part of  resource server.

each server (if its in cdi mode) will (re)create cdi specs for its resource pool when `triggerUpdate()` is called.

the ResourceManager should not concern itself with cdi specifics

thoughts ?

pkg/cdi/cdi.go
```diff
@@ -0,0 +1,105 @@
...
+	return nil
+}
+
+// CreateContainerAnnotations creates container annotations based on CDI spec for a container runtime
+func CreateContainerAnnotations(devicesIDs []string, cdiCache *cdi.Cache) (map[string]string, error) {
+	annotations := make(map[string]string, 0)
+	for _, id := range devicesIDs {
+		var cdiDevice *cdi.Device
+
+		for _, device := range cdiCache.ListDevices() {
```
adrianchiris, line 90: why cant you use `GetDevice()` directly ? what information are you missing ?

pkg/cdi/cdi.go
```diff
@@ -0,0 +1,105 @@
...
+			if strings.Contains(device, id) {
+				cdiDevice = cdiCache.GetDevice(device)
+			}
+		}
+		if cdiDevice != nil {
+			// TODO: (e0ne): add support of several devices
+			annotations["cdi.k8s.io/"+strings.ReplaceAll(cdiDevice.GetSpec().Kind, "/", "_")] = cdiDevice.GetSpec().Kind + "=" + cdiDevice.Name
```
adrianchiris, line 97: can you provide more info on the expected format of this annotation expected by container runtime ?

adrianchiris, line 97: we need to support multiple devices here as well.

cmd/sriovdp/manager.go
e0ne: It's used at https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/489/files#diff-f1a4e99ad0d1a580dda8274ab56d83450f41088bae0f4d0df0808d03c851d20dR216 and  https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/489/files#diff-f1a4e99ad0d1a580dda8274ab56d83450f41088bae0f4d0df0808d03c851d20dR264

cmd/sriovdp/manager.go
e0ne: > we dont need this one for now

We need the config to understand which devices should be included into the resource pool

cmd/sriovdp/manager.go
e0ne: agree, it should be deleted

deployments/cdi/README.md
```diff
@@ -0,0 +1,16 @@
...
+to be configured for the Device Plugin pod:
+
+```yaml
+    - mountPath: /etc/cdi/
+      name: default-cdi
+    - mountPath: /var/run/cdi
+      name: dynamic-cdi
+    - mountPath: /host/etc/pcidp/
```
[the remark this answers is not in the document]
e0ne: I'll remove it

cmd/sriovdp/manager.go
```diff
@@ -117,19 +119,76 @@ func (rm *resourceManager) initServers() error {
 
 		devices := dp.GetDevices(rc)
 		filteredDevices, err := dp.GetFilteredDevices(devices, rc)
-		if err != nil {
-			glog.Errorf("initServers(): error getting filtered devices for config %+v: %q", rc, err)
-		}
+
```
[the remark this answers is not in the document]
e0ne: What message do you mean?

deployments/cdi/sriovdp-daemonset.yaml
```diff
@@ -0,0 +1,120 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  annotations:
+    kubernetes.io/description: |
+      This daemon set launches the SR-IOV network device plugin on each node.
+    release.openshift.io/version: ""
+  name: sriov-device-plugin
+  namespace: network-operator
```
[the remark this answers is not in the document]
e0ne: let's leave it in this directory with own README to be more explicit that this is a separate deployment option

pkg/cdi/cdi.go
e0ne: GetDevice expects device ID in different format rather than device plugin provides

cmd/sriovdp/manager.go
e0ne: Only ResourceManager has information about filtered devices

adrianchiris:
@e0ne can you please address comments as well, i see you are applying changes,
this will make review process easier.

pkg/cdi/cdi.go
adrianchiris: The format is well defined and predictable, seeing you are the one who is writing this spec to begin with. so i think you can "reconstruct" it before calling getDevice.

if thats not the case, please explain and give example of the string format that GetDevice expects.

cmd/sriovdp/manager.go
adrianchiris: resource pool contains all the devices for the pool (which are already filtered by selectors)...

cmd/sriovdp/main.go
```diff
@@ -43,7 +46,7 @@ func main() {
 
 	glog.Infof("resource manager reading configs")
 	if err := rm.readConfig(); err != nil {
-		glog.Errorf("error getting resources from file %v", err)
+		glog.Error("error getting resources from file %v", err)
```
zeeke, line 49: nit: I think `glog.Errorf(...)` was good, as it contains a `%v`

pkg/resources/server.go
```diff
@@ -180,6 +201,16 @@ func (rs *resourceServer) ListAndWatch(empty *pluginapi.Empty, stream pluginapi.
 	}
 }
 
+func (rs *resourceServer) updateCDISpec() error {
+	rs.resourcePool.GetDevices()
```
zeeke, line 205: Do you know if this call has any effect? It sounds like a "get" where the return value is discarded. If it has any side  effect, please add a comment

zeeke:
CDI specification workflow is not yet 100% clear to me, but this PR shouldn't break any non-CDI scenarios.

There are no unit or end2end tests that show the changes work as intended, but I guess testing this isn't that trivial. 
if that's the case, LGTM

cmd/sriovdp/main.go
e0ne: fixed

pkg/resources/server.go
e0ne: `GetDevices` call is deleted now

deployments/cdi/sriovdp-daemonset.yaml
```diff
@@ -0,0 +1,115 @@
...
+        beta.kubernetes.io/os: linux
+        node-role.kubernetes.io/worker: ""
+      priorityClassName: system-node-critical
+      restartPolicy: Always
+      schedulerName: default-scheduler
+      securityContext: {}
+      serviceAccount: sriov-device-plugin
+      serviceAccountName: sriov-device-plugin
+      terminationGracePeriodSeconds: 30
+      tolerations:
```
adrianchiris, line 84: can you define this yaml like what we currently have under deployments  (with the added mounts and flag ofc) ?

images/entrypoint.sh
```diff
@@ -66,6 +71,11 @@ fi
 if [ "$CONFIG_FILE" != "" ]; then
     CLI_PARAMS="$CLI_PARAMS --config-file $CONFIG_FILE"
 fi
+
+if $USE_CDI
```
adrianchiris, line 75: please use same if format as above.

pkg/auxnetdevice/auxNetResourcePool.go
```diff
@@ -60,3 +64,8 @@ func (ap *auxNetResourcePool) GetDeviceSpecs(deviceIDs []string) []*pluginapi.De
 	}
 	return devSpecs
 }
+
+// GetPoolKind returns device kind for CDI spec
```
adrianchiris, line 68: nit: i dont think its relevant here that CDI uses this information.

lets name this method: GetPoolType which will essentially return the type of devices used by the  pool WDYT ?

pkg/types/types.go
```diff
@@ -193,6 +193,7 @@ type ResourcePool interface {
 	GetMounts(deviceIDs []string) []*pluginapi.Mount
 	StoreDeviceInfoFile(resourceNamePrefix string) error
 	CleanDeviceInfoFile(resourceNamePrefix string) error
+	GetPoolKind() string
```
adrianchiris, line 196: GetPoolType ? or Better yet, GetPoolDeviceType ?

pkg/resources/pool_stub.go
```diff
@@ -25,6 +25,10 @@ import (
 	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
 )
 
+const (
+	poolType = "net"
```
adrianchiris, line 29: net-pci ?

pkg/auxnetdevice/auxNetResourcePool.go
```diff
@@ -60,3 +64,8 @@ func (ap *auxNetResourcePool) GetDeviceSpecs(deviceIDs []string) []*pluginapi.De
...
+func (ap *auxNetResourcePool) GetPoolKind() string {
+	return auxPoolType
```
adrianchiris, line 70: should be IMO:

net-<aux Type>  or net-aux-<type>

e.g net-sf or net-aux-sf

i prefer the former it will also be consistent with net-pci  for the other pool.

pkg/resources/server.go
```diff
@@ -180,6 +202,15 @@ func (rs *resourceServer) ListAndWatch(empty *pluginapi.Empty, stream pluginapi.
 	}
 }
 
+func (rs *resourceServer) updateCDISpec() error {
+	err := cdiPkg.CreateCDISpec(rs.resourceNamePrefix, rs.resourcePool)
+	if err != nil {
+		glog.Errorf("initServers(): error creating CDI spec: %v", err)
```
adrianchiris, line 208: initServers ?

pkg/cdi/cdi.go
```diff
@@ -0,0 +1,102 @@
...
+			DeviceNodes: make([]*cdiSpecs.DeviceNode, 0),
+			Env:         make([]string, 0),
+		}
+
+		envs, err := rPool.GetEnvs(prefix, []string{dev.GetID()})
+		if err != nil {
+			glog.Errorf("CreateCDISpec(): Unable to create environment variables: %s", err)
+			return fmt.Errorf("unable to create environment variables: %s", err)
+		}
+		for name, value := range envs {
```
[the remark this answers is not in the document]
adrianchiris: also how will env variables will look in container ?

we may need to:
1. create CDI spec per allocate call
2. update CDI spec if exists in list and watch

adrianchiris (commented):
thx for addressing my comments @e0ne  !

added a few more :)
also this PR is missing unit tests to test changes in Allocate and ListAndWatch in server_test.go when in cdi mode.

cmd/sriovdp/main.go
adrianchiris: no systemd is needed

cmd/sriovdp/main.go
adrianchiris: was not fixed. its still glog.Error.

pkg/auxnetdevice/auxNetResourcePool.go
e0ne: done

pkg/accelerator/accelResourcePool.go
```diff
@@ -23,6 +23,10 @@ import (
 	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
 )
 
+const (
+	accelPoolType = "accel"
```
e0ne, line 27: @adrianchiris what pool name do you prefer for this kind?

pkg/auxnetdevice/auxNetResourcePool.go
e0ne: What do you mean? could you provide more details?

pkg/cdi/cdi.go
```diff
@@ -0,0 +1,102 @@
...
+	cdiSpecs "github.com/container-orchestrated-devices/container-device-interface/specs-go"
+	"github.com/golang/glog"
+
+	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
+)
+
+// CreateCDISpec creates CDI spec file with specified devices
+func CreateCDISpec(resourcePrefix string, rPool types.ResourcePool) error {
+	prefix := resourcePrefix
+	if prefixOverride := rPool.GetResourcePrefix(); prefixOverride != "" {
```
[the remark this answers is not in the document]
e0ne: it makes sence, fixed

pkg/cdi/cdi.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/container-orchestrated-devices/container-device-interface/pkg/cdi"
+	cdiSpecs "github.com/container-orchestrated-devices/container-device-interface/specs-go"
+	"github.com/golang/glog"
+
+	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
+)
+
+// CreateCDISpec creates CDI spec file with specified devices
+func CreateCDISpec(resourcePrefix string, rPool types.ResourcePool) error {
```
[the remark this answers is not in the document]
e0ne: done

pkg/cdi/cdi.go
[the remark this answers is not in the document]
e0ne: @adrianchiris I didn't understand your comment. Could you please describe what do you mean?

pkg/resources/pool_stub.go
e0ne: done

pkg/resources/server.go
```diff
@@ -155,6 +169,14 @@ func (rs *resourceServer) ListAndWatch(empty *pluginapi.Empty, stream pluginapi.
 		return err
 	}
 
+	if rs.cdiCache != nil {
+		err := rs.updateCDISpec()
```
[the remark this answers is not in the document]
e0ne: good catch, fixed

pkg/types/types.go
e0ne: I prefer to keep it as is because it's 'kind' in CDI spec

cmd/sriovdp/manager.go
e0ne: I removed a usage of Cache

cmd/sriovdp/manager.go
e0ne: done

cmd/sriovdp/manager.go
e0ne: done

pkg/resources/server.go
e0ne: done

pkg/resources/server.go
```diff
@@ -118,18 +122,46 @@ func (rs *resourceServer) NotifyRegistrationStatus(ctx context.Context,
...
-		containerResp.Envs = envs
 
-		containerResp.Mounts = rs.resourcePool.GetMounts(container.DevicesIDs)
+		if rs.useCdi {
+			containerResp.Annotations = make(map[string]string, 0)
+
+			for _, id := range container.DevicesIDs {
+				var cdiDevice *cdi.Device
+
+				for _, device := range rs.cdiCache.ListDevices() {
```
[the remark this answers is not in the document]
e0ne: this is refactored

cmd/sriovdp/manager.go
e0ne: removed

deployments/cdi/sriovdp-daemonset.yaml
```diff
@@ -0,0 +1,120 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  annotations:
+    kubernetes.io/description: |
+      This daemon set launches the SR-IOV network device plugin on each node.
+    release.openshift.io/version: ""
```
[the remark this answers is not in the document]
e0ne: fixed

pkg/cdi/cdi.go
e0ne: refactored and fixed

pkg/cdi/cdi.go
e0ne: refactored to use existing function

pkg/cdi/cdi.go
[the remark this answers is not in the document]
adrianchiris: These are general questions regarding this package and behaviour in general

1, you are creating a CDI spec in this package right ? how does a CDI spec file look like ?
2. when you allocate device to container through CDI what will be the env vars defined in container


1 and 2 when there are multiple devices per resource pool and per container allocation.

pkg/accelerator/accelResourcePool.go
adrianchiris: accel is OK imo

pkg/types/types.go
adrianchiris: so GetPoolCDIKind ? 

Kind lacks context here IMO

pkg/cdi/cdi.go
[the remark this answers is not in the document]
e0ne: 1. CDI spec file example could be found here: https://github.com/container-orchestrated-devices/container-device-interface#full-blown-cdi-specification
2. With CDI I we pass exactly the same env vars as we do as a device plugin mode now.
3. annotation key and values are generated bu cdi library and I don't want to provice an example in code comments because it could be outdated once we bump cdi version. For a several devices, we'll have following annotation: `example.com/kind=example.com/kind=0000:00.00.1,example.com/kind=0000:00.00.2`

pkg/types/types.go
e0ne: Let's simplify it to `GetPoolCDIKind`

e0ne: > Let's simplify it to `GetPoolCDIKind`

I mean `GetCDIKind`

deployments/cdi/sriovdp-daemonset.yaml
```diff
@@ -0,0 +1,234 @@
...
+        operator: Exists
+        effect: NoSchedule
+      serviceAccountName: sriov-device-plugin
+      containers:
+      - name: kube-sriovdp
+        image: ghcr.io/k8snetworkplumbingwg/sriov-network-device-plugin:latest-amd64
+        imagePullPolicy: IfNotPresent
+        args:
+        - --log-dir=sriovdp
+        - --log-level=10
```
almaslennikov, line 42: Should we define `--use-cdi` argument for cdi-specific daemonset?

pkg/accelerator/accelResourcePool.go
adrianchiris: on a second thought.

we have: `net-sf`, `net-pci`  so lets have this one as `net-accel`  :)

deployments/cdi/sriovdp-daemonset.yaml
```diff
@@ -0,0 +1,234 @@
...
+      labels:
+        name: sriov-device-plugin
+        tier: node
+        app: sriovdp
+    spec:
+      hostNetwork: true
+      nodeSelector:
+        kubernetes.io/arch: amd64
+      tolerations:
+      - key: node-role.kubernetes.io/master
```
adrianchiris, line 32: while at it, can you add toleration for `node-role.kubernetes.io/control-plane`

deployments/cdi/sriovdp-daemonset.yaml
```diff
@@ -0,0 +1,234 @@
...
+          hostPath:
+            path: /etc/cdi
+            type: DirectoryOrCreate
+        - name: dynamic-cdi
+          hostPath:
+            path: /var/run/cdi
+            type: DirectoryOrCreate
+
+
+---
```
adrianchiris, line 93: rest of the daemonsets dont have the needed cdi mounts...
nor the --use-cdi flag

pkg/types/types.go
```diff
@@ -193,6 +193,7 @@ type ResourcePool interface {
 	GetMounts(deviceIDs []string) []*pluginapi.Mount
 	StoreDeviceInfoFile(resourceNamePrefix string) error
 	CleanDeviceInfoFile(resourceNamePrefix string) error
+	GetCDIKind() string
```
adrianchiris, line 196: this should be GetCDIName() string

since this bit is essentially the name segment of Kind. see: [1]

[1] https://github.com/cncf-tags/container-device-interface/blob/main/SPEC.md#kind

pkg/cdi/cdi.go
```diff
@@ -0,0 +1,108 @@
...
+			containerEdit.DeviceNodes = append(containerEdit.DeviceNodes, &deviceNode)
+		}
+		device := cdiSpecs.Device{
+			Name:           dev.GetID(),
+			ContainerEdits: containerEdit,
+		}
+		cdiSpec.Devices = append(cdiSpec.Devices, device)
+	}
+
+	err := cdi.GetRegistry().SpecDB().WriteSpec(&cdiSpec, resourcePrefix)
```
adrianchiris, line 79: what will happen when we have two resource pools with the same prefix ? will it overwrite ? (i think yes)

pkg/resources/server_test.go
```diff
@@ -413,5 +414,47 @@ var _ = Describe("Server", func() {
 				close(done)
 			}, 30.0)
 		})
+		Context("when CDI is enabled", func() {
```
adrianchiris, line 417: can you also add a test case for Allocate when CDI is enabled ?

pkg/cdi/cdi.go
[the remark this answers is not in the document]
adrianchiris: @e0ne to keep compatibility with how currently device plugin exposes env variables, one option is to use current device plugin mechanism for env vars and just move device spec to CDI.

alternative that i can think of, is to generate a CDI device kind per allocation but thats hacky.

deployments/cdi/sriovdp-daemonset.yaml
e0ne: sure, we should

deployments/cdi/sriovdp-daemonset.yaml
e0ne: done

deployments/cdi/sriovdp-daemonset.yaml
e0ne: why? all these daemonsets should use CDI mode

pkg/types/types.go
e0ne: done

pkg/cdi/cdi.go
e0ne: it will but we shouldn't create two resource pools with the same name

pkg/resources/server_test.go
e0ne: done

deployments/cdi/README.md
```diff
@@ -0,0 +1,13 @@
+# SR-IOV Network Device Plugin in CDI mode deployment
+
+SR-IOV Network Device Plugin supports [Container Device Interface (CDI)](https://github.com/container-orchestrated-devices/container-device-interface).
+
+To enable CDI mode, SR-IOV Network Device Plugin should be started with `--use-cdi` CLI argument.
+This mode hase different deployment requirements: `sriovdp-daemonset.yaml`
```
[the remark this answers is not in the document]
e0ne: thanks, fixed

pkg/accelerator/accelResourcePool.go
e0ne: done

pkg/cdi/cdi.go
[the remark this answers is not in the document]
e0ne: environment variables are fixed in the latest version

deployments/cdi/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+SR-IOV Network Device Plugin supports [Container Device Interface (CDI)](https://github.com/container-orchestrated-devices/container-device-interface).
+
+To enable CDI mode, SR-IOV Network Device Plugin should be started with `--use-cdi` CLI argument.
+This mode has different deployment requirements: `sriovdp-daemonset.yaml`
+
+```yaml
+    - mountPath: /etc/cdi/
+      name: default-cdi
+    - mountPath: /var/run/cdi
```
adrianchiris, line 11: we need just /var/run/cdi imo we dont deal with statically configured files right ?

adrianchiris: we should align daemonsets as well

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.20
 
 require (
 	github.com/Mellanox/rdmamap v1.1.0
+	github.com/container-orchestrated-devices/container-device-interface v0.5.4
```
adrianchiris, line 7: reminder to check if we prefer to:

1. bump this to v0.6.0 and require containerd 1.7.5 or cri-o 1.28 or keep as is.

pkg/resources/server.go
```diff
@@ -324,6 +365,9 @@ func (rs *resourceServer) cleanUp() error {
 	if err := rs.resourcePool.CleanDeviceInfoFile(rs.resourceNamePrefix); err != nil {
 		errors = append(errors, err.Error())
 	}
+	if err := rs.cdi.CleanupSpecs(); err != nil {
```
adrianchiris, line 368: i think we should:

1. cleanup spec files which are related only to this resource  server her
2. in the resource manager we should cleanup any remaining cdi spec files before starting servers just in case on un-gracefull exit of previous device plugin instance.

thoughts ?

pkg/cdi/cdi.go
```diff
@@ -0,0 +1,124 @@
...
+type CDI interface {
+	CreateCDISpecForPool(resourcePrefix string, rPool types.ResourcePool) error
+	CreateContainerAnnotations(devicesIDs []string, resourcePrefix, resourceKind string) (map[string]string, error)
+	CleanupSpecs() error
+}
+
+// Impl implements CDI interface
+type Impl struct {
+}
+
```
adrianchiris, line 43: nit : define a New() method

func New() CDI {
    return &impl{}
}

and use it in resource server

also consider making Impl struct private

pkg/cdi/cdi.go
```diff
@@ -0,0 +1,124 @@
...
+// CreateCDISpecForPool creates CDI spec file with specified devices
+func (c *Impl) CreateCDISpecForPool(resourcePrefix string, rPool types.ResourcePool) error {
+	err := c.CleanupSpecs()
```
adrianchiris, line 46: this will delete spec files of other resource servers. see other comments  related to where/how we should clean up.

pkg/cdi/cdi.go
```diff
@@ -0,0 +1,124 @@
...
+	annotations[annoKey] = annoValue
+
+	return annotations, nil
+}
+
+// CleanupSpecs removes previously-created CDI specs
+func (c *Impl) CleanupSpecs() error {
+	for _, dir := range cdi.GetRegistry().GetSpecDirectories() {
+		specs, err := filepath.Glob(filepath.Join(dir, cdiSpecPrefix+"*"))
+		if err != nil {
```
adrianchiris, line 113: will this one fail if dir doesnt exist ?

if yes, then for such error we need to skip.

pkg/cdi/cdi_test.go
```diff
@@ -0,0 +1,56 @@
...
+		})
+		It("should not fail on non-existing device", func() {
+			deviceId := "0000:00:00.2"
+			cdi := cdiPkg.Impl{}
+			annotations, err := cdi.CreateContainerAnnotations([]string{deviceId}, "example.com", "net")
+			Expect(err).NotTo(HaveOccurred())
+			annoKey := "cdi.k8s.io/example.com_net"
+			annoVal := "example.com/net=0000:00:00.2"
+			Expect(annotations[annoKey]).To(Equal(annoVal))
+		})
```
adrianchiris, line 54: need to add test for cleanup method as well

adrianchiris (commented):
final comments on my side, once addressed im LGTM.

pkg/cdi/cdi.go
almaslennikov: > Glob returns the names of all files matching pattern or nil if there is no matching file. The syntax of patterns is the same as in Match. The pattern may describe hierarchical names such as /usr/*/bin/ed (assuming the Separator is '/').
> Glob ignores file system errors such as I/O errors reading directories. The only possible returned error is ErrBadPattern, when pattern is malformed.

https://pkg.go.dev/path/filepath#Glob

go.mod
e0ne: According to https://github.com/containerd/containerd/blob/v1.7.5/go.mod#L10, we need to keep v0.5.4 now

pkg/resources/server.go
e0ne: IMO, it's better to move this logic to resource manager only: it will cleanup all spec files before resorce servers start, so no raise condidion in this case and all outdated specs will be removed.  With such behaviour we don't need to filter for orphaned spec files if config changed during device plugin restart

pkg/cdi/cdi.go
e0ne: please loot at https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/489#discussion_r1322815114

pkg/cdi/cdi.go
e0ne: done

pkg/cdi/cdi_test.go
e0ne: done

pkg/cdi/cdi.go
```diff
@@ -0,0 +1,129 @@
...
+			containerEdit.DeviceNodes = append(containerEdit.DeviceNodes, &deviceNode)
+		}
+		device := cdiSpecs.Device{
+			Name:           dev.GetID(),
+			ContainerEdits: containerEdit,
+		}
+		cdiSpec.Devices = append(cdiSpec.Devices, device)
+	}
+
+	err = cdi.GetRegistry().SpecDB().WriteSpec(&cdiSpec, cdiSpecPrefix+resourcePrefix)
```
SchSeba, line 83: nit: please use fmt for this `cdiSpecPrefix+resourcePrefix`

pkg/cdi/cdi.go
```diff
@@ -0,0 +1,129 @@
...
+// New returns an instance of CDI interface implementation
+func New() CDI {
+	return &impl{}
+}
+
+// CreateCDISpecForPool creates CDI spec file with specified devices
+func (c *impl) CreateCDISpecForPool(resourcePrefix string, rPool types.ResourcePool) error {
+	err := c.CleanupSpecs()
+	if err != nil {
+		glog.Error("can not cleanup old spec files", err)
```
SchSeba, line 53: please add `CreateCDISpecForPool():` so we can be consistent in the logs or remove it from the log below

pkg/cdi/cdi_test.go
```diff
@@ -0,0 +1,56 @@
...
+			deviceId := "0000:00:00.1"
+			cdi := cdiPkg.New()
+			annotations, err := cdi.CreateContainerAnnotations([]string{deviceId}, "example.com", "net")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(len(annotations)).To(Equal(1))
+			annoKey := "cdi.k8s.io/example.com_net"
+			annoVal := "example.com/net=0000:00:00.1"
+			Expect(annotations[annoKey]).To(Equal(annoVal))
+		})
+		It("should not fail on non-existing device", func() {
```
SchSeba, line 46: are you sure the test is right?

pkg/resources/server.go
```diff
@@ -119,18 +124,28 @@ func (rs *resourceServer) NotifyRegistrationStatus(ctx context.Context,
...
 			return nil, err
 		}
-		containerResp.Envs = envs
 
-		containerResp.Mounts = rs.resourcePool.GetMounts(container.DevicesIDs)
+		if rs.useCdi {
+			containerResp.Annotations, err = rs.cdi.CreateContainerAnnotations(
+				container.DevicesIDs, rs.resourceNamePrefix, rs.resourcePool.GetCDIName())
+			if err != nil {
+				return nil, fmt.Errorf("cant create container annotation: %s", err)
```
SchSeba, line 141: nit: cant -> can't

pkg/resources/server.go
```diff
@@ -147,8 +162,12 @@ func (rs *resourceServer) ListAndWatch(empty *pluginapi.Empty, stream pluginapi.
 		devs = append(devs, dev)
 	}
 	resp.Devices = devs
+	err := rs.updateCDISpec()
+	if err != nil {
+		glog.Error("cannot update CDI specs", err)
```
SchSeba, line 167: should be errorf with %v

pkg/resources/server.go
```diff
@@ -147,8 +162,12 @@ func (rs *resourceServer) ListAndWatch(empty *pluginapi.Empty, stream pluginapi.
 		devs = append(devs, dev)
 	}
 	resp.Devices = devs
+	err := rs.updateCDISpec()
+	if err != nil {
```
SchSeba, line 166: nit: you can have this in the same line

SchSeba (commented):
nice work!
I left some comments

cmd/sriovdp/main.go
```diff
@@ -27,12 +27,15 @@ const (
...
 )
 
-// Parse Command line flags
+// flagInit parse command line flags
 func flagInit(cp *cliParams) {
 	flag.StringVar(&cp.configFile, "config-file", defaultConfig,
 		"JSON device pool config file location")
 	flag.StringVar(&cp.resourcePrefix, "resource-prefix", "intel.com",
-		"resource name prefix used for K8s extended resource")
+		"resource name prefix used for K8s extended re"+
```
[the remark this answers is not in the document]
e0ne: fixed

cmd/sriovdp/main.go
```diff
@@ -43,7 +46,7 @@ func main() {
 
 	glog.Infof("resource manager reading configs")
 	if err := rm.readConfig(); err != nil {
-		glog.Errorf("error getting resources from file %v", err)
+		glog.Error("error getting resources from file", err)
```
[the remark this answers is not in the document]
e0ne: done

cmd/sriovdp/manager.go
```diff
@@ -101,12 +109,16 @@ func (rm *resourceManager) readConfig() error {
 }
 
 func (rm *resourceManager) initServers() error {
+	err := rm.cleanupCDISpecs()
+	if err != nil {
+		glog.Infof("Unable to delete CDI specs: %s", err)
```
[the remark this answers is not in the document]
e0ne: done

cmd/sriovdp/manager.go
```diff
@@ -249,3 +261,12 @@ func (rm *resourceManager) discoverHostDevices() error {
 	}
 	return nil
 }
+
+func (rm *resourceManager) cleanupCDISpecs() error {
+	if rm.cliParams.useCdi {
+		if err := rm.cdi.CleanupSpecs(); err != nil {
+			return fmt.Errorf("unable to delete CDI specs: %s", err)
```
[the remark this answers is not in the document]
e0ne: fixed

pkg/cdi/cdi.go
e0ne: done

pkg/cdi/cdi_test.go
e0ne: You're righ. I'll delete this test

pkg/resources/server.go
e0ne: done

pkg/resources/server.go
e0ne: done

pkg/resources/server.go
e0ne: thanks, done!

pkg/cdi/cdi.go
```diff
@@ -0,0 +1,129 @@
...
+// New returns an instance of CDI interface implementation
+func New() CDI {
+	return &impl{}
+}
+
+// CreateCDISpecForPool creates CDI spec file with specified devices
+func (c *impl) CreateCDISpecForPool(resourcePrefix string, rPool types.ResourcePool) error {
+	err := c.CleanupSpecs()
+	if err != nil {
+		glog.Error("CreateCDISpecForPool(): can not cleanup old spec files", err)
```
zeeke, line 53: Shouldn't this be `glog.Errorf("....%v", err)`?

pkg/cdi/cdi.go
```diff
@@ -0,0 +1,129 @@
...
+
+	return nil
+}
+
+// CreateContainerAnnotations creates container annotations based on CDI spec for a container runtime
+func (c *impl) CreateContainerAnnotations(devicesIDs []string, resourcePrefix, resourceKind string) (map[string]string, error) {
+	annotations := make(map[string]string, 0)
+	annoKey, err := cdi.AnnotationKey(resourcePrefix, resourceKind)
+	if err != nil {
+		glog.Error("CreateContainerAnnotations(): can't create container annotation", err)
```
zeeke, line 97: >Shouldn't this be `glog.Errorf("....%v", err)`?
Same here and on line 106

zeeke (approved):
Left a few minor comments about error management. 
Beside that, LGTM

pkg/cdi/cdi.go
e0ne: fixed

pkg/cdi/cdi.go
e0ne: fixed

e0ne:
> Left a few minor comments about error management. Beside that, LGTM

thanks, @zeeke

SchSeba (commented):
just some last small nits

can you please also add a section about CDI in the main document that point to the CDI readme?

cmd/sriovdp/manager.go
```diff
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"os"
 
+	cdiPkg "github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/cdi"
```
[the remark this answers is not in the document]
e0ne: done

cmd/sriovdp/manager.go
```diff
@@ -101,12 +109,16 @@ func (rm *resourceManager) readConfig() error {
 }
 
 func (rm *resourceManager) initServers() error {
+	err := rm.cleanupCDISpecs()
+	if err != nil {
+		glog.Errorf("Unable to delete CDI specs: %s", err)
```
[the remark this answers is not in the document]
e0ne: done

pkg/accelerator/accelResourcePool.go
```diff
@@ -23,6 +23,10 @@ import (
 	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
 )
 
+const (
+	accelPoolType = "net-accel"
```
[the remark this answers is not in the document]
SchSeba: for now we decide to leave it with the net-

deployments/cdi/sriovdp-daemonset.yaml
```diff
@@ -0,0 +1,240 @@
...
+        resources:
+          requests:
+            cpu: "250m"
+            memory: "40Mi"
+          limits:
+            cpu: 1
+            memory: "200Mi"
+        volumeMounts:
+        - name: devicesock
+          mountPath: /var/lib/kubelet/
```
SchSeba, line 58: please update the mounts to be the same as https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/500

SchSeba (commented):
last comment and we can merge this PR! 

Great work!

deployments/cdi/sriovdp-daemonset.yaml
e0ne: done

SchSeba (approved):
Great work!

e0ne:
Thanks for the feedback, @SchSeba !

Issue #492, opened 2023-07-03
Wrong DeviceInfo file content for vhost-vdpa devices


lmilleri:
### What happened?
DeviceInfo file content (in directory /var/run/k8s.cni.cncf.io/devinfo/dp/) for vhost-vdpa devices is not generated properly.

### What did you expect to happen?
Wrong DeviceInfo (the **path** is not correct):
```
{
  "type": "vdpa",
  "version": "1.0.0",
  "vdpa": {
    "parent-device": "vdpa:0000:65:00.3",
    "driver": "vhost",
    "path": "pci/0000:65:00.3",
    "pci-address": "0000:65:00.3"
  }
}
```
Expected DeviceInfo:
```
{
  "type": "vdpa",
  "version": "1.0.0",
  "vdpa": {
    "parent-device": "vdpa:0000:65:00.3",
    "driver": "vhost",
    "path": "/dev/vhost-vdpa-0",
    "pci-address": "0000:65:00.3"
  }
}
```

### What are the minimal steps needed to reproduce the bug?
- Setup an OCP cluster with at least one BM worker. Also an NVIDIA Connect-6 DX card is present in the system
- Partition the NIC into VFs with SRIOV and create vhost-vdpa devices on top of VFs (switchdev mode and OVS HW offload enabled)
- Deploy the sriov-network-device-plugin
- apply DP config map like in this example:
```
{
  "resourceList": [
    {
      "resourceName": "mlxnics",
      "selectors": {
        "vendors": [
          "15b3"
        ],
        "devices": [
          "101e"
        ],
        "rootDevices": [
          "0000:65:00.0"
        ],
        "IsRdma": false,
        "NeedVhostNet": false,
        "vdpaType": "vhost"
      },
      "SelectorObj": null
    }
  ]
}
```

### Component Versions
Please fill in the below table with the version numbers of components used.

Component | Version|
------------------------------|--------------------|
|SR-IOV Network Device Plugin |latest|
|SR-IOV CNI Plugin            |Not applicable, using ovn-kubernetes|
|Multus                       |OCP 4.13|
|Kubernetes                   |OCP 4.13| 
|OS                           |Linux virtlab711.virt.lab.eng.bos.redhat.com 5.14.0-284.18.1.el9_2.x86_64 #1 SMP PREEMPT_DYNAMIC Wed May 31 10:39:18 EDT 2023 x86_64 x86_64 x86_64 GNU/Linux|

### Config Files
[...]

PR #493, opened 2023-07-03
vhost-vdpa: fix DeviceInfo file content

lmilleri:
This PR fixes an issue in the DeviceInfo file generation for vhost-vdpa devices.

For details please check the following issue:
https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/issues/492


adrianchiris (approved):

lmilleri:
@SchSeba @bn222 Can you please have a look and merge if looks good to you?

SchSeba (approved):

lmilleri:
@bn222 AFAIK there is no error condition to be handled at the call site. Let's say, if a path is returned, it'll be used to fill some structs, otherwise an empty path will be stored instead. The glog.Infof is just for convenience.

bn222:
It seems to me that whenever this function returns "", it was due to it being called on something that doesn't have this path, i.e. it should not have been called on it the first place?

lmilleri:
Added error handling for GetPath function

e0ne (approved):
LGTM

pkg/infoprovider/vdpaInfoProvider.go
```diff
@@ -60,7 +60,11 @@ func (vip *vdpaInfoProvider) GetDeviceSpecs() []*pluginapi.DeviceSpec {
 
 	// DeviceSpecs only required for vhost vdpa type as the
 	if vip.vdpaType == types.VdpaVhostType {
-		vdpaPath := vip.dev.GetPath()
+		vdpaPath, err := vip.dev.GetPath()
+		if err != nil {
+			glog.Errorf("Unexpected error when fetching the vdpa device path: %s", err)
+			return devSpecs
```
adrianchiris, line 66: shouldn't we return nil like above ? (When device not healthy)

lmilleri: I think it would be safer, will do the change thanks!

PR #496, opened 2023-08-02
Recognize modern taint in demo

jcpunk:
Modern kubernetes does not set the `master` designation any longer.  This patch adds in support for `control-plane` while preserving the old behavior.


SchSeba (approved):
LGTM

thanks for the PR

adrianchiris (approved):

PR #500, opened 2023-09-07
Update device plugins volumes mounts

e0ne:
There is no need to mount the whole '/var/lib/kubelet/' directory into a pod. It's enough to mount only '/var/lib/kubelet/device-plugins' and '/var/lib/kubelet/plugins_registry' directories.


adrianchiris (approved):

SchSeba (approved):
Nice!

small nit: can you please update here also?
https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/82d8aae3487e20f81901702a9e8e6aa66f8191fa/images/README.md#L32

PR #502, opened 2023-10-02
update the capacity to zero on shutdown/reset

SchSeba:
When the device plugin is restarted, kubelet marks the resource as unhealthy, but still reports the resource as existing for a grace period (5 mins). If a pod is scheduled before the device plugin comes up, the pod create fails without a retryloop with an error message Pod was rejected: Allocate failed due to no healthy devices present; cannot allocate unhealthy devices <DEVICE_NAME>, which is unexpected.

This commit allow the device plugin to send an empty list of devices before the reset or shutdown


pkg/resources/server.go
```diff
@@ -309,6 +322,14 @@ func (rs *resourceServer) restart() error {
 	// Send terminate signal to ListAndWatch()
 	rs.termSignal <- true
 
+	// wait for the terminated signal or 5 second
```
zeeke, line 325: In `resoruceServer.Stop()` we do the `termSignal-terminatedSignal` handshake before stopping the grpcServer. n restart, we do it after `rs.grpcServer.Stop()`. Is it intentional?

SchSeba: you are right! sorry

SchSeba:
@zeeke please give it another look :)

adrianchiris:
hmm im not sure that thats what kubelet expects (being suddenly reported there are no devices when device plugin shuts 
 down or restarts)

will it remove entries from checkpoint file ? 
what if there are pods already using resources ?

SchSeba:
>hmm im not sure that thats what kubelet expects (being suddenly reported there are no devices when device plugin shuts
down or restarts)
will it remove entries from checkpoint file ?
what if there are pods already using resources ?

From my tests it doesn't remove the checkpoint file and running pods continue to run (I will try to leave it down for a longer period to be sure there is no reconcile or something that will kill the running pods)

without this change, we have an issue when we take down the pod and the pod is allocated to that node the pod will not be able to start.

SchSeba:
related to https://github.com/openshift/sriov-network-operator/issues/812

adrianchiris:
@SchSeba been digging a bit into kubelet code 

kubelet will report updated status to kubelet every 10 seconds[1] (every NodeStatusUpdateFrequency[2])

once device plugin plugin exits (its endpoint no longer valid), all devices are deemed unhealty[3]

[1] https://github.com/kubernetes/kubernetes/blob/d61cbac69aae97db1839bd2e0e86d68f26b353a7/pkg/kubelet/kubelet.go#L1637
[2] https://github.com/kubernetes/kubernetes/blob/d61cbac69aae97db1839bd2e0e86d68f26b353a7/staging/src/k8s.io/kubelet/config/v1beta1/types.go#L267
[3] https://github.com/kubernetes/kubernetes/blob/d61cbac69aae97db1839bd2e0e86d68f26b353a7/pkg/kubelet/nodestatus/setters.go#L342

so after at most 10 seconds node will report zero allocatable resources if plugin has exited.
if any pod is scheduled to the node before that, it will enter admission error.
setting devices explicitly as un-healthy will not help as unhealty devices is only reported to kube api during that sync loop.

what i suggest, is in sriov-network-operator after we remove device plugin to spin up a goroutine to keep deleting pods in admission error if they consume resources from device plugin until new device plugin is up  or alternatively (even better option imo) once device plugin is up again clean up any pods which consume dp resources in admission error once.

SchSeba:
Interesting I will try to implement a POC in the operator but before that a question do you see any issue changing the number to 0 when we reboot? can that effect something?

adrianchiris:
1. i do not know what the original authors of the API (grpc) intended.
2. technically we dont achieve more reliability if we do it IMO.
3. when device plugin exits, it will mark the resources reported as invalid (internally) IIRC. which is the same as if plugin exited.
in both cases it only affects scheduler when node status is updated in apiserver.

PR #512, opened 2023-11-23
Fix push-manifest github action job for release

adrianchiris:
Add missing env variable


adrianchiris:
@zeeke PTAL

zeeke (approved):
LGTM

thanks for fixing this

PR #515, opened 2023-11-29
Fix image push worklow for release

adrianchiris:
- disable latest tag generation in docker meta action
- remove un-needed IMAGE_NAME var from manifest creation, docker meta tags is already with the full name.


adrianchiris:
https://github.com/adrianchiris/sriov-network-device-plugin/actions/runs/7032394480/job/19136039734

adrianchiris:
we got same problem in sriov cni ill submit PR there as well.

zeeke (approved):
LGTM

PR #526, opened 2024-01-23
Add fallback to get PF Name for AuxNetDevice

adrianchiris:
In case we fail to get Uplink Representor, fallback to get PF PCI address and use sysfs to get PF netdevice.

This is needed in case a device supports the creation of Auxiliary devices but is not in switchdev.


adrianchiris:
@Eoghan1232 PTAL if you have some time.

Eoghan1232 (approved):
this looks fine to me, nice to have some backups in place :)

adrianchiris:
CI issue not related to PR.

I have tested this one in my cluster. works well.

PR #527, opened 2024-01-29
Update go version and github actions to 1.21

almaslennikov:
Upcoming changes will require go 1.21 so it's better to update the toolchain separately


adrianchiris:
@Eoghan1232  PTAL, straight forward.

Eoghan1232 (approved):

PR #529, opened 2024-02-06
Downgrade ddp-builder container version

almaslennikov:
Upgrading to golang:1.21-alpine3.19 caused the ddp build to fail: https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/actions/runs/7738698140
Let's downgrade it until the root cause is found

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -10,7 +10,7 @@ WORKDIR /usr/src/sriov-network-device-plugin
 RUN make clean && \
     make build
 
-FROM golang:1.21-alpine3.19 as ddp-builder
+FROM golang:1.20-alpine3.16 as ddp-builder
 
 ADD images/ddptool-1.0.1.12.tar.gz /tmp/ddptool/
 


images/Dockerfile
```diff
@@ -10,7 +10,7 @@ WORKDIR /usr/src/sriov-network-device-plugin
 RUN make clean && \
     make build
 
-FROM golang:1.21-alpine3.19 as ddp-builder
+FROM golang:1.20-alpine3.16 as ddp-builder
```
SchSeba, line 13: I think a better idea will be to have a different container for the ddp tool we want the go code to be on 1.21

almaslennikov: I propose to address this in a separate PR, so that we unblock the image build first

almaslennikov: The go code is built with golang 1.21, ddp uses a different layer

adrianchiris: @SchSeba can you please open an issue on this ?

adrianchiris (approved):
LGTM

adrianchiris:
@Eoghan1232 PTAL, currently build&push image GA workflow is failing, this should fix it.

Eoghan1232 (approved):

images/Dockerfile
SchSeba: oh it was already done sorry I miss it...

PR #530, opened 2024-02-08
Expose RDMA device name in the devicespec if available

almaslennikov:
The DeviceInfoSpec allows to specify rdma device for a pci device: https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/blob/master/pkg/apis/k8s.cni.cncf.io/v1/types.go#L63

Multus propagates device spec to the network-status, so it's easy to track which RDMA device is allocated to a POD:
```
k8s.v1.cni.cncf.io/network-status:
	[{
		"name": "k8s-pod-network",
		"ips": [
			"192.168.107.206"
		],
		"default": true,
		"dns": {}
	},{
		"name": "default/sriov-net1",
		"interface": "net1",
		"ips": [
			"10.56.217.2"
		],
		"mac": "7a:bb:87:d9:b4:94",
		"dns": {},
		"device-info": {
			"type": "pci",
			"version": "1.1.0",
			"pci": {
				"pci-address": "0000:03:00.3",
				"rdma-device": "mlx5_4"
			}
		},
		"gateway": [
			"\u003cnil\u003e"
		]
	}]
```
This change can serve two purposes:
1. Simplify debugging. Users can just query pod's network status and understand which RDMA device belongs to which network
2. Allow applications in multi-network pods to understand, which RDMA device they should use.

### DIFF
--- a/pkg/netdevice/netResourcePool_test.go
+++ b/pkg/netdevice/netResourcePool_test.go
@@ -24,6 +24,8 @@ import (
 	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/netdevice"
 	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
 	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types/mocks"
+	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/utils"
+	utilsmocks "github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/utils/mocks"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


pkg/netdevice/netResourcePool_test.go
```diff
@@ -113,10 +113,12 @@ var _ = Describe("NetResourcePool", func() {
 
 			fake1 := &mocks.PciNetDevice{}
 			fake1.On("GetPciAddr").Return("0000:01:00.1").
-				On("GetVdpaDevice").Return(nil)
+				On("GetVdpaDevice").Return(nil).
```
adrianchiris, line 116: can you add a new test for rdma device.

SchSeba, line 116: can you also please add a test to validate we have the RDMA info in the deviceSpec?

pkg/netdevice/netResourcePool_test.go
```diff
@@ -113,26 +115,38 @@ var _ = Describe("NetResourcePool", func() {
...
 			fake2 := &mocks.PciNetDevice{}
 			fake2.On("GetPciAddr").Return("0000:01:00.2").
-				On("GetVdpaDevice").Return(nil)
+				On("GetVdpaDevice").Return(nil).
+				On("IsRdma").Return(false)
 			pcis := map[string]types.HostDevice{"fake1": fake1, "fake2": fake2}
 
+			fakeRdmaProvider := utilsmocks.RdmaProvider{}
+			fakeRdmaProvider.On("GetRdmaDevicesForPcidev", "0000:01:00.1").Return([]string{"rdmadevice1"})
+			utils.SetRdmaProviderInst(&fakeRdmaProvider)
```
adrianchiris, line 128: if we set it, i think we need to reset to orig value

adrianchiris (approved):

adrianchiris:
@SchSeba PTAL

SchSeba (approved):
LGTM

PR #543, opened 2024-03-12
Update deployment yamls

adrianchiris:
define a single daemonset which utilizes container manifest.


adrianchiris:
@Eoghan1232 PTAL , i just  changed yaml to use container mainfest instead of per-arch tags

Eoghan1232 (approved):

Eoghan1232:
+1

PR #551, opened 2024-04-11
remove grpc deprication functions

SchSeba:


pkg/resources/server.go
```diff
@@ -271,17 +271,31 @@ func (rs *resourceServer) Start() error {
 			glog.Errorf("serving incoming requests failed: %s", err.Error())
 		}
 	}()
-	// Wait for server to start by launching a blocking connection
-	ctx, _ := context.WithTimeout(context.TODO(), serverStartTimeout)
-	conn, err := grpc.DialContext(
```
SchSeba, line 276: new version of the grpc doesn't make a connection when creating the client so to maintain the same functionally here we need to create the client and run a connect function that is a blocking one

jaypipes (approved):
I hit this with my vendor removal patch. ++

SchSeba:
Thanks!

merging this one

PR #580, opened 2024-07-25
Allow to override go tags in build

bn222:


bn222:
@adrianchiris your previous PR didn't address this yet.

@SchSeba also ptal here and lgtm. We need to use this downstream.

SchSeba:
cc @adrianchiris 
please take a look

adrianchiris (approved):

PR #600, opened 2024-09-24
restart the device plugin if NotifyRegistrationStatus returns failure

liuyuan10:
Today it only logs and stop the rpc server. The plugin hangs after that.

### DIFF
--- a/pkg/resources/server.go
+++ b/pkg/resources/server.go
@@ -115,8 +115,8 @@ func (rs *resourceServer) NotifyRegistrationStatus(ctx context.Context,
 	if regstat.PluginRegistered {
 		glog.Infof("Plugin: %s gets registered successfully at Kubelet\n", rs.endPoint)
 	} else {
-		glog.Infof("Plugin: %s failed to be registered at Kubelet: %v; restarting.\n", rs.endPoint, regstat.Error)
 		rs.grpcServer.Stop()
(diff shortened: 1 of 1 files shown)


liuyuan10:
@adrianchiris I wonder if you can take a quick look of this PR? it should be a small fix

SchSeba (approved):
LGTM

SchSeba:
I re-open to trigger the CI system again

@adrianchiris when you have time please take a look on this one

pkg/resources/server.go
```diff
@@ -115,8 +115,8 @@ func (rs *resourceServer) NotifyRegistrationStatus(ctx context.Context,
 	if regstat.PluginRegistered {
 		glog.Infof("Plugin: %s gets registered successfully at Kubelet\n", rs.endPoint)
 	} else {
-		glog.Infof("Plugin: %s failed to be registered at Kubelet: %v; restarting.\n", rs.endPoint, regstat.Error)
 		rs.grpcServer.Stop()
+		glog.Fatalf("Plugin: %s failed to be registered at Kubelet: %v; restarting.\n", rs.endPoint, regstat.Error)
```
adrianchiris, line 119: im not sure im OK with logging a msg then panic,

maybe we need some way to signal errors and exit gracefully ?
what others think of this ?

SchSeba: sure we can switch to log.err and exit 1 or something like that

liuyuan10: isn't log.Fatal the same as log and exit 1?

PR #611, opened 2024-12-09
Fix the device spec file creation in rdma exclusive mode

SchSeba:
when using exclusive mode in RDMA if we restart the device plugin it will find the right rdma device if the device is allocated.

this change makes the file creation when we get a new allocation and not in the device plugin initialization


SchSeba:
Hi @e0ne @ykulazhenkov @rollandf @zeeke can you please take a look on this fix please?

rollandf (approved):
LGTM, thanks

zeeke (approved):
LGTM

PR #613, opened 2024-12-10
BUG: Remove device spec before creating a new one

SchSeba:


SchSeba:
Hi @zeeke @rollandf,

This is the fix that I introduce on the previews PR.
I didn't check that the nadFunc recreates the file instead it just return error :(

The clean function will not fail if the file doesn't exist so we should be good

zeeke (approved):
LGTM

There are some unit test to fix. Thanks for fixing this

e0ne (approved):

PR #640, opened 2025-04-09
Add health check for Net Devices

wizhaoredhat:
The implementation checks:
    1. If the physical function PF of the SR-IOV devices is carrier down. This
    should be marked unhealthy. Normally, SR-IOV would still function when the
    PF is carrier down. But in the case of DPUs/IPUs/SmartNics with an embedded
    CPU, the PF being down **can** signal that the embedded CPU is in reset or
    shutdown with carrier down.
    2. If any of the devices are gone. This could be due to someone changing
    the number of virtual functions. Or in the case of DPUs/IPUs/SmartNics with
    an embedded CPU, the driver needed to reset. This will cause the virtual
    functions to be removed. All devices that are gone should be marked
    unhealthy. Normally this won't be the case since the SR-IOV Network Operator
    will be managing the SR-IOV devices. However for DPUs/IPUs/SmartNics with
    an embedded CPU, would be externally managed with a seperate operator.
    
    Both these can be switched on and off using checkHealthOnPf and
    checkHealthOnDeviceExist within the resource config.


wizhaoredhat:
@zeeke / @SchSeba  PTAL

pkg/netdevice/netResourcePool_test.go
```diff
@@ -237,4 +237,107 @@ var _ = Describe("NetResourcePool", func() {
...
+			fake2 := &mocks.PciNetDevice{}
+			fake2.On("GetPfNetName").Return("fakepf1")
+			fake2.On("GetPciAddr").Return("0000:01:00.2")
+			fake2.On("SetHealth", Anything).Run(func(args Arguments) {
+				fake2health = args.Bool(0)
+			}).Return()
+
+			pcis := map[string]types.HostDevice{"fake1": fake1, "fake2": fake2}
+
+			rp := netdevice.NewNetResourcePool(nadutils, rc, pcis)
```
zeeke, line 271: can you move this section in a `BeforeEach()`? code outside ginkgo construct is called very early during tests.
This would also allow getting rif of the `mock.Unset()` calls

pkg/netdevice/netResourcePool_test.go
```diff
@@ -237,4 +237,107 @@ var _ = Describe("NetResourcePool", func() {
...
+			}
+
+			SetCurrentDeviceExistance := func(exist bool) {
+				fake1.On("DeviceExists").Unset()
+				fake2.On("DeviceExists").Unset()
+				fake1.On("DeviceExists").Return(exist, nil).Once()
+				fake2.On("DeviceExists").Return(exist, nil).Once()
+			}
+
+			RunTest := func(health, link, exist bool) {
```
zeeke, line 296: we can leverage the ginkgo DescribeTable feature 
https://onsi.github.io/ginkgo/#table-specs

or we can repeat the logic a little in each of the `It()` test case, removing the `if/else` statement from the assertion. Though this involves repeating the code, it makes the test cases linear and readable. 

Up to you which solution to adopt

zeeke (approved):
Left a few minor comments

pkg/netdevice/netResourcePool_test.go
wizhaoredhat: Ah very good point, I have implemented this, it is much cleaner now, thanks!

pkg/netdevice/netResourcePool_test.go
wizhaoredhat: I think it makes sense to repeat the logic a little in each It() case. This was what I had it before. I cleaned it up now.

wizhaoredhat:
@zeeke Please take a look at the changes again.

wizhaoredhat:
@adrianchiris PTAL

SchSeba:
Hi @wizhaoredhat,
please check the CI failed

SchSeba (commented):
overall looks good check two small comments.

a following question is where are we going to use this feature? it needs to be part of the sriov-network-operator or a different operator that deploys a device plugin?

pkg/netdevice/netResourcePool.go
```diff
@@ -66,6 +68,82 @@ func (rp *netResourcePool) GetDeviceSpecs(deviceIDs []string) []*pluginapi.Devic
...
+			deviceExistsLog = fmt.Sprintf("Device %s", netDev.GetPciAddr())
+			if deviceExists {
+				deviceExistsLog = fmt.Sprintf("%s is existing, ", deviceExistsLog)
+			} else {
+				deviceExistsLog = fmt.Sprintf("%s is missing, ", deviceExistsLog)
+			}
+		}
+
+		if pfIsUp && deviceExists && !currentHealth {
+			glog.Infof("%s%sdevice was unhealthy, marking device %s as healthy", pfIsUpLog, deviceExistsLog, id)
```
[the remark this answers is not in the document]
wizhaoredhat: Done

pkg/netdevice/netResourcePool.go
```diff
@@ -66,6 +68,82 @@ func (rp *netResourcePool) GetDeviceSpecs(deviceIDs []string) []*pluginapi.Devic
...
+			deviceExists = netDev.DeviceExists()
+			deviceExistsLog = fmt.Sprintf("Device %s", netDev.GetPciAddr())
+			if deviceExists {
+				deviceExistsLog = fmt.Sprintf("%s is existing, ", deviceExistsLog)
+			} else {
+				deviceExistsLog = fmt.Sprintf("%s is missing, ", deviceExistsLog)
+			}
+		}
+
+		if pfIsUp && deviceExists && !currentHealth {
```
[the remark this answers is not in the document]
wizhaoredhat: Please take a look now at the reduce if statement. Let me know if it is better.

wizhaoredhat:
> overall looks good check two small comments.
> 
> a following question is where are we going to use this feature? it needs to be part of the sriov-network-operator or a different operator that deploys a device plugin?

Some operators use the Device Plugin with the SR-IOV Network Operator. I think some NVIDIA operators do this. But for the DPU Operator, I plan to include this change and use reuse the SR-IOV device plugin. DPUs and Smart NICs would benefit with this change since DPUs or SmartNICs can be reset/reboot/shutdown asynchronously.

One thing I dislike is the fact that kubernetes doesn't put the pod into a failed state when the underlying resource allocation device becomes unhealthy. I think the expectation of k8s is to use liveliness probes. I plan to dig a little bit more on this.

I think these flags included now in the Config Map can be exposed to the Sriov Network Operator API after this change has been merged. I think the SriovNetworkNodePolicy? Just wondering if you think this would be useful for the SR-IOV Network Operator or not.

wizhaoredhat:
@adrianchiris  could you please take a look? This would be useful when the NVIDIA DPU is not managed by the DPF operator.

README.md
```diff
@@ -286,6 +286,8 @@ This plugin creates device plugin endpoints based on the configurations given in
...
 | "deviceType"      | N        | Device Type for a resource pool.                                                                                                       | string value of supported types. Default: "netDevice" | Currently supported values: "accelerator", "netDevice", "auxNetDevice" |
 | "excludeTopology" | N        | Exclude advertising of device's NUMA topology                                                                                          | bool Default: "false"                                 | "excludeTopology": true                                                |
+| "checkHealthOnPf" | N        | Check the health of a net device by inspecting the link state of the PF                                                                                       | bool Default: "false"                                 | "checkHealthOnPf": true                                                |
```
adrianchiris, line 289: should this be a list ? that way we dont overcumber the API 

e.g
`healthChecks: ["DeviceNetworkLink", DeviceExists]`

pkg/resources/server.go
```diff
@@ -73,7 +73,7 @@ func NewResourceServer(prefix, suffix string, pluginWatch, useCdi bool, rp types
 		termSignal:         make(chan bool, 1),
 		updateSignal:       make(chan bool),
 		stopWatcher:        make(chan bool),
-		checkIntervals:     20, // updates every 20 seconds
+		checkIntervals:     5, // updates every 5 seconds
```
wizhaoredhat, line 76: Light be too short for Telco environments.

PR #655, opened 2025-08-05
chore: adjust golangci.yml file

adrianchiris:
- disable check for switch statement instead of if/else
- disable check for method reciever naming


adrianchiris:
@SchSeba did some tweaks to the golangci file, we now allow if/elseif statements like before and not force the use of switch.

PR #656, opened 2025-08-05
fix: dont stop grpc server on send failure in ListAndWatch

adrianchiris:
in case of failure to send, return error. client (kubelet) can then retry.


silivanmunguar:
I was running into issues rebasing and syncing the fork. So here is a new pr on a new branch all together https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pull/658

PR #658, opened 2025-08-11
Fixes #648 Return an err on transient failures rather than stopping the server to allow for retry

### LINKED ISSUE #648: SR-IOV Device Plugin Crash During ListAndWatch Causes Resource Reporting Failure
## What issue would you like to bring attention to?
A node is incorrectly reporting zero available Mellanox SR-IOV VFIO resources, which prevents it from being used for  worker node deployment. The root cause appears to be that the SR-IOV device plugin's gRPC server shuts down when it encounters an error and does not attempt to start again

## What is the impact of this issue?
A node becomes unavailable for worker deployment, which can lead to resource underutilization and scheduling failures. This also introduces operational overhead in diagnosing and recovering from the issue manually.

## Do you have a proposed response or remediation for the issue?
Consider modifying the[ gRPC server behavior](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/15ceeb3245f35a2d4a66a802d27b4ee8f81d24f5/pkg/resources/server.go#L174) in the device plugin to restart on error using glog.Fatal() as is done [here ](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/15ceeb3245f35a2d4a66a802d27b4ee8f81d24f5/pkg/resources/server.go#L294)or a similar mechanism. This would allow the plugin to recover automatically instead of just stopping entirely.

silivanmunguar:
This PR addresses an issue where the SR-IOV device plugin's gRPC server would terminate unexpectedly due to a transient error during the ListAndWatch operation. This premature shutdown led to false reporting of zero available Mellanox SR-IOV VFIO resources, causing affected nodes to be marked as unschedulable.

### To improve resilience:
The gRPC server no longer stops the server when a transient error occurs when it attempts to advertise resources.
Instead, the error is logged, and an err is returned allowing the server to continue running and giving the device plugin an opportunity to recover through the kubelet's  retry.
Additionally, the Watch method already includes this code https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/5cc9452448734b5ca01c7441f3de9a2ba48f4724/pkg/resources/server.go#L343, which attempts recovery before ultimately terminating the server if retries fail. This ensures the plugin has a chance to recover from transient issues and continue reporting resources accurately, reducing the risk of node exclusion due to temporary failures, even in scenarios where the kubelet is down and a race condition is suspected.

### DIFF
--- a/pkg/resources/server.go
+++ b/pkg/resources/server.go
@@ -178,7 +178,6 @@ func (rs *resourceServer) ListAndWatch(empty *pluginapi.Empty, stream pluginapi.
 	glog.Infof("%s: send devices %v\n", methodID, resp)
 	if err := stream.Send(resp); err != nil {
 		glog.Errorf("%s: error: cannot update device states: %v\n", methodID, err)
-		rs.grpcServer.Stop()
 		return err
 	}
 


zeeke (approved):
LGTM

Thanks for haveing addressed our comments

PR #676, opened 2025-10-28
add s390x support to build and image workflows.

ashokpariya0:


ashokpariya0:
> Hi @ashokpariya0 like in the other projects lets have one job to publish all the arch to the same container image tag please

>
Sure, I am on it.

PR #689, opened 2025-12-26
Set networkPolicy to drop all

jcpunk:
Since this neither produces no consumes network traffic, it should be safe to drop everything right?

## Summary by CodeRabbit

* **New Features**
  * Added network isolation policy for SR-IOV device plugin pods, establishing network access controls and traffic restrictions within the cluster infrastructure. The policy governs incoming and outgoing communications for these infrastructure components.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/deployments/sriovdp-daemonset.yaml
+++ b/deployments/sriovdp-daemonset.yaml
@@ -76,3 +76,21 @@ spec:
             items:
             - key: config.json
               path: config.json
+
+---
+apiVersion: networking.k8s.io/v1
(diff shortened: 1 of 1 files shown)


jcpunk:
I was thinking primarily that, while ineffective, this would at least provide "documentation" of the expected network traffic.  But if you'd rather not carry this, I'll not object.

SchSeba:
closing this one. as the container is plan to only run as hostNetwork: true

PR #709, opened 2026-04-14
chore: removing entrypoint.sh and adding its content to sriovdp

heyvister1:


cmd/sriovdp/main.go
```diff
@@ -16,19 +16,39 @@ package main
...
+			"\t--config-file=\n"+
+			"\t--use-cdi\n",
+		os.Args[0], defaultLogLevel)
+}
+
+// flagInit registers CLI flags (same surface as the former cmd/entrypoint wrapper).
+func flagInit(cp *cliParams, logDir *string, logLevel *int) {
+	flag.Usage = usage
+	flag.StringVar(logDir, "log-dir", "", "Log directory under /var/log/")
+	flag.IntVar(logLevel, "log-level", defaultLogLevel, "Log verbosity level")
```
SchSeba, line 51: you are missing `CLI_PARAMS="-v $LOG_LEVEL"` as gemini commented

cmd/sriovdp/main.go
```diff
@@ -37,10 +57,37 @@ func flagInit(cp *cliParams) {
...
+
+	return nil
+}
+
 func main() {
 	cp := &cliParams{}
-	flagInit(cp)
+	var logDir string
+	logLevel := defaultLogLevel
+	flagInit(cp, &logDir, &logLevel)
```
SchSeba, line 85: why we need to pass the logDir and logLevel to the function? I think it's better to just return that from the function or have it as a global variable in the package

PR #715, opened 2026-06-02
Persistent logging enhancement for SRIOV device plugin

gavrielg1:
Design documentation for applying persistent logging for SR-IOV device plugin based [CNF-24291](https://redhat.atlassian.net/browse/CNF-24291)


SchSeba (commented):
looks good to me.

question after the restart the new pod will continue to write to the same open files as before? will it be a new file?
if it's the same file I think we need to add to the main a starting header like `------` + pod name or something like that

gavrielg1:
> question after the restart the new pod will continue to write to the same open files as before? will it be a new file?
> if it's the same file I think we need to add to the main a starting header like `------` + pod name or something like that

It will continue write to the same open file, I will add it to the final implementation

PR #717, opened 2026-06-09
Add dedicated tmpfs volume for /tmp and enhanced security

jcpunk:
My site security audit complained that I didn't have a dedicated `/tmp` volume.

I've also added a second commit that reduces the privilege set.

### DIFF
--- a/deployments/cdi/sriovdp-daemonset.yaml
+++ b/deployments/cdi/sriovdp-daemonset.yaml
@@ -55,7 +55,21 @@ spec:
         - --log-level=10
         - --use-cdi
         securityContext:
-          privileged: true
+          privileged: false
+          capabilities:
+            add:
+              - BPF
+              - IPC_LOCK
+              - NET_ADMIN
+              - NET_RAW
+              - SYS_ADMIN
+              - SYS_PTRACE
+              - SYS_RAWIO
+            drop:
+              - ALL
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          runAsNonRoot: false
         resources:
           requests:
             cpu: "250m"
--- a/deployments/cdi/sriovdp-daemonset.yaml
+++ b/deployments/cdi/sriovdp-daemonset.yaml
@@ -64,6 +64,8 @@ spec:
             cpu: 1
             memory: "200Mi"
         volumeMounts:
+        - name: tmp
+          mountPath: /tmp
         - name: devicesock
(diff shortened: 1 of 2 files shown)


deployments/cdi/sriovdp-daemonset.yaml
```diff
@@ -77,6 +79,9 @@ spec:
         - name: dynamic-cdi
           mountPath: /var/run/cdi
       volumes:
+      - name: tmp
+        emptyDir:
+          sizeLimit: 32Mi
```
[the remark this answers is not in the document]
jcpunk: explicit backing not required.

deployments/sriovdp-daemonset.yaml
```diff
@@ -57,6 +59,9 @@ spec:
         - name: device-info
           mountPath: /var/run/k8s.cni.cncf.io/devinfo/dp
       volumes:
+        - name: tmp
+          emptyDir:
+            sizeLimit: 32Mi
```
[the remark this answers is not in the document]
jcpunk: explicit backing not required.

SchSeba (approved):
nice improvement.

can you please take a look on https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/bindata/manifests/plugins/sriov-device-plugin.yaml and lets have a similiar PR on that repo if possible?

jcpunk:
I don't actually use the operator, so I'm worried about test verification... should I start using the operator instead?

deployments/cdi/sriovdp-daemonset.yaml
```diff
@@ -55,7 +55,21 @@ spec:
...
+              - NET_ADMIN
+              - NET_RAW
+              - SYS_ADMIN
+              - SYS_PTRACE
+              - SYS_RAWIO
+            drop:
+              - ALL
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          runAsNonRoot: false
```
adrianchiris, line 72: does it work when --useCdi flag is set to true ? (writes files to host mount)

jcpunk: I don't have a system to test `--userCdi`, but any physical directories mounted in from the host should inherit permissions as normal.

adrianchiris: ack.

adrianchiris (approved):

SchSeba (approved):

PR #726, opened 2026-07-08
Apply least-privilege best practices to DaemonSet

yahlifried:
Use explicit capabilities instead of privileged mode, add a seccomp profile, and scope volume mounts to only the paths the plugin actually needs. Disable service account token automount since the plugin does not interact with the Kubernetes API.

Both the standard and CDI DaemonSet variants are updated.

## Summary by CodeRabbit

* **Security**
  * Tightened security settings for the related DaemonSet pods.
  * Disabled automatic ServiceAccount token mounting, reduced container privileges, and applied a default seccomp profile.
  * Restricted Linux capabilities to network administration.
  * Updated storage mounts so logs are written to a dedicated path and the host `/sys` directory is mounted read-only.

### DIFF
--- a/deployments/cdi/sriovdp-daemonset.yaml
+++ b/deployments/cdi/sriovdp-daemonset.yaml
@@ -27,6 +27,7 @@ spec:
     spec:
       hostNetwork: true
       serviceAccountName: sriov-device-plugin
+      automountServiceAccountToken: false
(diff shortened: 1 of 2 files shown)


SchSeba:
Hi @yahlifried please rebase this one

PR #730, opened 2026-07-20
Add Slingshot CXI Support

### LINKED ISSUE #734: Add Slingshot CXI Support
Support Slingshot Cassini (CXI) VFs. Specifically, mount the VF device `/dev/cxiN` similarly to how RDMA devices are mounted.

The next Slingshot Host Software release will include SR-IOV support. Ideally, it would integrate nicely with Kubernetes using the existing sriov-network-device-plugin. The network interface is getting setup fine, but without the device mounted, RDMA traffic fails.

codambro:
## Description
Similar to RDMA devices, Slingshot CXI NIC devices need to be mounted in the pods.
However, the `isRdma` flag only looks in /dev/infiniband. To avoid confusion between devices, decided to add a separate `isCxi` config flag. If true, the CXI device associated with the VF net interface will be mounted.

Also added example to usage docs.

Fixes #734 

## Testing
Modified the sample test yaml:
```
cn-0007:~ # cat pod-tc1.yaml
apiVersion: v1
kind: Pod
metadata:
  name: testpod1
  annotations:
    k8s.v1.cni.cncf.io/networks: cxi-net
spec:
  containers:
  - name: appcntr1
    image: docker.io/flannel/flannel:v0.25.2
    imagePullPolicy: IfNotPresent
    command: [ "/bin/bash", "-c", "--" ]
    args: [ "while true; do sleep 300000; done;" ]
    resources:
      requests:
        hpe.com/cxi: '1'
      limits:
        hpe.com/cxi: '1'
```

And verified both the ethernet interface and CXI device is configured in the pod:
```
cn-0007:~ # kubectl exec testpod1 -- sh -c 'ls /dev/cxi* 2>/dev/null; echo "PCIDEVICE_HPE_COM_CXI=$PCIDEVICE_HPE_COM_CXI"; ip addr show net1'
/dev/cxi37
PCIDEVICE_HPE_COM_CXI=0000:5a:00.1
109: net1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9000 qdisc mq state UP group default qlen 1000
    link/ether 02:00:01:00:08:1b brd ff:ff:ff:ff:ff:ff
    altname enp90s0f1
    inet 10.150.200.30/16 brd 10.150.255.255 scope global net1
       valid_lft forever preferred_lft forever
    inet6 fe80::1ff:fe00:81b/64 scope link proto kernel_ll
       valid_lft forever preferred_lft forever
cn-0007:~ #
```

## New Features
- Added support for Cassini (CXI) devices.
- CXI devices can be selected using the `isCxi` configuration field.
- Detected CXI character devices are exposed to workloads at `/dev/cxiN`.
- Added a sample `cxi` resource configuration, including the deployment ConfigMap.

## Documentation
- Documented the `isCxi` selector and its device-mounting behavior.

## Tests
- Added coverage for CXI device discovery, selection, and missing-device scenarios.

### DIFF
--- a/pkg/infoprovider/cxiInfoProvider.go
+++ b/pkg/infoprovider/cxiInfoProvider.go
@@ -62,8 +62,13 @@ func (ip *cxiInfoProvider) GetDeviceSpecs() []*pluginapi.DeviceSpec {
 
 	// Confirm the corresponding /dev char device actually exists before mounting it.
 	devPath := filepath.Join(CxiDevDir, filepath.Base(cxiDev))
-	if _, err := os.Stat(devPath); err != nil {
+	info, err := os.Stat(devPath)
+	if err != nil {
 		glog.Errorf("GetDeviceSpecs(): cxi device file %s does not exist for device: %s, %s", devPath, ip.pciAddr, err.Error())
+		return devSpecs
+	}
+	if info.Mode()&os.ModeCharDevice == 0 {
+		glog.Errorf("GetDeviceSpecs(): cxi device file %s for device %s is not a character device", devPath, ip.pciAddr)
 		return devSpecs
 	}
 
--- a/pkg/infoprovider/cxiInfoProvider.go
+++ b/pkg/infoprovider/cxiInfoProvider.go
@@ -15,12 +15,19 @@
 package infoprovider
 
 import (
+	"os"
+	"path/filepath"
+
 	"github.com/golang/glog"
 	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
 
 	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/types"
 	"github.com/k8snetworkplumbingwg/sriov-network-device-plugin/pkg/utils"
 )
+
+// CxiDevDir is the base directory for CXI char devices.
+// It is a variable to allow overriding the path in unit tests.
+var CxiDevDir = "/dev"
 
 /*
 cxiInfoProvider provides the Cassini (CXI) char device information.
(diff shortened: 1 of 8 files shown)


codambro:
I disagree with CodeRabbit. This PR is more in-line with current coding practices of the repo, matching error messages and test stylings of other providers. It's nit-picking, and in doing so introduced a bug that I had to then track down and revert.

codambro:
@SchSeba @jcpunk 
Can I get a review on this please?

SchSeba:
Hi @codambro you have plans to use it via the sriovNetworkOperator? because in that case we will need to expose this also in the sriovNetworkNodePolicy of the sriov operator

pkg/infoprovider/cxiInfoProvider.go
```diff
@@ -0,0 +1,86 @@
+// Copyright 2018 Intel Corp. All Rights Reserved.
```
SchSeba, line 1: nit: can you update this one I think it should be 2026 and the k8snetworkplumbingwg org

pkg/infoprovider/cxiInfoProvider.go
```diff
@@ -0,0 +1,86 @@
...
+
+	cxiDev, err := utils.GetCxiDeviceFile(ip.pciAddr)
+	if err != nil {
+		glog.Errorf("GetDeviceSpecs(): error getting cxi device file for device: %s, %s", ip.pciAddr, err.Error())
+		return devSpecs
+	}
+
+	// Note: the existence of the /dev char device is intentionally not verified here.
+	// HostPath is resolved by kubelet on the host, whereas this code runs inside the
+	// device-plugin container which does not mount the host's /dev.
```
SchSeba, line 58: the pod is priviledge so it should be able to see the `/dev` it's the same as we do for the /dev/vfio devices

pkg/infoprovider/cxiInfoProvider.go
```diff
@@ -0,0 +1,86 @@
...
+		glog.Errorf("GetEnvVal(): error getting cxi device file for device: %s, %s", ip.pciAddr, err.Error())
+	} else {
+		envs["dev-mount"] = cxiDev
+	}
+
+	return envs
+}
+
+func (ip *cxiInfoProvider) GetMounts() []*pluginapi.Mount {
+	return nil
```
SchSeba, line 83: please follow the same as the other provides and lets have

```
mounts := make([]*pluginapi.Mount, 0)
	return mounts
```

SchSeba (commented):
in general looks good I left some small comments.

also there is a plan from your company to provide a CI with this card? is it possible to have a virtual device that will support this? 

in general it's hard to add support for new features when we do no have CI lanes to tests the hardware.

codambro:
Currently there is no plan to use the sriovNetworkOperator. The targeted initial use-case is for an admin to create VFs ahead of time that kubernetes can use.
We have added testing for CXI VFs with kubernetes and the sriov-network-device-plugin to our internal CI testing

PR #735, opened 2026-07-30
Validate resourcePrefix to prevent path traversal

yahlifried:
The resourcePrefix field from the device plugin ConfigMap was used unsanitized in socket path construction, CDI spec naming, and device info file paths. A malicious prefix containing path separators could escape the base directory via filepath.Join normalization.

Add ValidResourcePrefix regex validation in validConfigs() and defense-in-depth checks in GetResourceServer() and NewResourceServer().

## Summary by CodeRabbit

* **Bug Fixes**
  * Configured resource prefixes are now validated, and invalid values are rejected before resource services are created.
  * Invalid socket paths that would escape the intended socket directory now return an error instead of creating a service with a modified path.
  * Valid custom prefixes and existing default behavior remain supported.

* **Tests**
  * Added coverage for invalid resource prefixes and socket paths, confirming that no resource service is created when validation fails.

### DIFF
--- a/pkg/resources/server.go
+++ b/pkg/resources/server.go
@@ -58,9 +58,16 @@ const (
 // NewResourceServer returns an instance of ResourceServer
 func NewResourceServer(prefix, suffix string, pluginWatch, useCdi bool, rp types.ResourcePool) types.ResourceServer {
 	sockName := fmt.Sprintf("%s_%s.%s", prefix, rp.GetResourceName(), suffix)
-	sockPath := filepath.Join(types.SockDir, sockName)
+	baseDir := types.SockDir
 	if !pluginWatch {
-		sockPath = filepath.Join(types.DeprecatedSockDir, sockName)
+		baseDir = types.DeprecatedSockDir
+	}
+	sockPath := filepath.Join(baseDir, sockName)
+	if !strings.HasPrefix(filepath.Clean(sockPath), filepath.Clean(baseDir)+string(filepath.Separator)) {
+		glog.Errorf("computed socket path %q escapes base directory %q; using sanitized prefix", sockPath, baseDir)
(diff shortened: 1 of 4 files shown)


pkg/utils/utils.go
```diff
@@ -303,6 +303,12 @@ func ValidResourceName(name string) bool {
 	return validString.MatchString(name)
 }
 
+// ValidResourcePrefix returns true if the prefix contains only characters safe for filesystem paths and DNS subdomains
+func ValidResourcePrefix(prefix string) bool {
```
SchSeba: please check in the k8s api machinery I think they should have a function to validate resources already and we can just reuse them

yahlifried: Done

SchSeba:
closed and re-open to trigger the CI

pkg/resources/server.go
```diff
@@ -58,9 +58,16 @@ const (
...
-	sockPath := filepath.Join(types.SockDir, sockName)
+	baseDir := types.SockDir
 	if !pluginWatch {
-		sockPath = filepath.Join(types.DeprecatedSockDir, sockName)
+		baseDir = types.DeprecatedSockDir
+	}
+	sockPath := filepath.Join(baseDir, sockName)
+	if !strings.HasPrefix(filepath.Clean(sockPath), filepath.Clean(baseDir)+string(filepath.Separator)) {
+		glog.Errorf("computed socket path %q escapes base directory %q; using sanitized prefix", sockPath, baseDir)
+		prefix = "invalid-prefix"
```
SchSeba, line 68: lets not do this lets change the function header to return error and we use it in the factory to return in `GetResourceServer`

yahlifried: Done

